PR #20, opened 2025-05-03
Fix sm agent missing .md filetype

mirzacmarius:
Added missing filetype for sm-agent


bmadcode:
@mirzacmarius Thank you so much!

PR #23, opened 2025-05-05
Documentation agent

chidev:
I found this guy super useful for recursively updating docs, used it on 2 large scale enterprise code bases so far and it did amazing


bmadcode:
Awesome, love it!! I like the file organization also - will need to make a slight tweak maybe to the agents so they know where to find the files - but with the concept of the _index.md (or similar if in here) I want to set the agents up to use the index to find files anyways so docs can be updated.

bmadcode:
Oh I reread all of the code looks like it may handle different structures already through discovery - cant wait to try this out later!

PR #55, opened 2025-05-16
Fix semantic of instruction for dev-agent.md

fabb:


NicolasSampaio (approved):

fabb:
I think i understood it wrong. In v3 the instruction is `Never draft next story`, so the original wording is correct. Closing.

PR #82, opened 2025-05-23
Fix Checklist Mapping References and Update File Extensions

adriansprk:
I have made the following changes to improve the checklist mapping and ensure that the references are accurate:

**Updated File Extensions:**

1. Changed the file extensions in the checklist references from .txt to .md to reflect the conversion of documents to Markdown format.

2. Corrected the references in the checklist mapping YAML file to ensure they point to the correct Markdown files.

These changes ensure that the checklist mappings are accurate and consistent with the current file formats, which should help avoid any confusion or errors when accessing the referenced documents.


bmadcode:
@adriansprk Thanks so much - good catch!

PR #130, opened 2025-05-30
Switch PM and PO in ide-bmad-orchestrator.cfg.md

fabb:
this change matches the order in the web orchestrator, and the logic order when creating a new project


bmadcode:
good idea @fabb thanks!

PR #131, opened 2025-05-30
Update mentioned roles in ide-setup.md for BMAD v3

fabb:


bmadcode:
Good catch, thanks @fabb !

PR #163, opened 2025-06-04
Task template standardization improvements

bmadcode:
A few bundled changes here, some of these fixes I could but in separately but for now they are bundled together - mostly around getting parity between web and ide agent configs for the orchestrator.

Added Advanced Elicitation to its own file, so it wil not be necessary to repeat it - so far only prototyped in the new generic doc runner + new Templating standardization applied to the PRD.

The main overall goal here is to prototype a better consistent format to separate the task (generate a document) from the template, and embed the agent instructions for producing the document within the template.

Rationale: This will allow customization of what is in a users desired PRD, along with what points throughout the agent will pause to review a section, and when to offer advanced elicitation. This should streamline and group sections that go together, while allowing more reflection on more critical portions. But this is all now configurable within the template.

Also by removing a lot of the instructions from the old create-PRD task, it will now be easier to create many different doc tempaltes for different types of PRD for example. So if someone wants a highly technical PRD and forgo an architecture for example, this will be possible without the need to have a new task.

Also introduced a standard for template markup within markdown, so its clear what are agent instructions, what are examples, what is placeholder text, what sections should repeat etc, now in a very consistent manner.

Would love some feedback on the ideas. I have tested this with multiple run throughs of the web agent creating PRDs with the generic doc template, and it seems really solid and streamlined.

This can merge on its own, and then will align the remaining documents to follow this standardization.

The build output should cause no breaking changes for existing users using V3 re documents being generated.

### DIFF
--- a/web-build-sample/agent-config.txt
+++ b/web-build-sample/agent-config.txt
@@ -1,3 +1,5 @@
+# Configuration for Web Agents
+
 ## Title: BMAD
 
 - Name: BMAD
@@ -17,9 +19,6 @@
   - "Brain Storming"
   - "Deep Research"
   - "Project Briefing"
-- Interaction Modes:
-  - "Interactive"
-  - "YOLO"
 - templates:
   - [Project Brief Tmpl](templates#project-brief-tmpl)
 
@@ -27,20 +26,16 @@
 
 - Name: John
 - Customize: ""
-- Description: "For PRDs, project planning, PM checklists and potential replans."
+- Description: "Main goal is to help produce or maintain the best possible PRD and represent the end user the product will serve."
 - Persona: "personas#pm"
 - checklists:
   - [Pm Checklist](checklists#pm-checklist)
   - [Change Checklist](checklists#change-checklist)
-- templates:
-  - [Prd Tmpl](templates#prd-tmpl)
 - tasks:
-  - [Create Prd](tasks#create-prd)
+  - [Create Document](tasks#create-doc-from-template):
+    - [Prd](templates#prd-tmpl)
   - [Correct Course](tasks#correct-course)
   - [Create Deep Research Prompt](tasks#create-deep-research-prompt)
-- Interaction Modes:
-  - "Interactive"
-  - "YOLO"
 
 ## Title: Architect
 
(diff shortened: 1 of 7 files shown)


OverlordBaconPants:
# TL;DR - Template Away!

I think I'd very much love to see this in action. I'm curious of the extent to which we can / should standardize our own interactions with our AI Agents. I like the notion of templating as much of this as we can yet am concerned I might want to take it too far.

image

## Use Case

Post-MVP (e.g., I just implemented a massive networking/collaboration feature in my app that would have easily taken me two weeks but I did in eight hours ❤️ ), I will have an 80% solution and will iterate through a process making several atomic tweaks to fix things. 

## Templated Interactions

I have the following instructions in a separate document that I'll re-use almost verbatim. This is all done using the Cline extension of VSCode to interact with Claude 3.7 Sonnet.

### PLAN Mode

- [Explain the problem I'm facing, pasting Terminal output snippets (like errors) and browser console output (to show logging) for context.] 
- Template P1: "In your capacity as Product Owner, using the role descriptions and checklists available to you in the /docs/ folder, please triage the root cause of this problem, propose at least two solutions with a list of pros and cons, and let me know which solution you prefer and why."
- [I almost always get a thorough and deep understanding of the problem that answers the mail. Rarely will I need to course correct and remind the AI Agent of some nuance of my tech stack]
- Template P2: "In your capacities as Scrum Master and Test Engineer, using the role descriptions and checklists available to you in the /docs/ folder, please create a Markdown file comprising TDD-centric Stories structured logically to resolve this issue."
- [I almost always get a wonderful .md file out of this.]

### ACT Mode

- Template A1: "In your capacities as Scrum Master and Developer, please implement the first/next Story in the Markdown file you just created, update documentation as appropriate once all unit tests successfully pass, and give me a status update on what's done and what's left to do."

OverlordBaconPants:
What if the AI Agent asked the user to select from a menu of offerings related to the artifact at-hand?

image

bmadcode:
Thanks for the feedback, I think I follow what you are saying and I do agree it can go to far - and for some even the configurability is not at all desired - and for others, their goal is not even development - which is amazing, just not the thought I had when I created the v1 of the repo. I am going to start a thread on this very topic as it gets to the heart of - WTH is this we are really building here? haha And the answer has evolved, so I really want some community insight into what we want to do here and why.

bmadcode:
I do plan to merge this after a bit more testing if no concerns or any thing this is a terrible direction. Started a discussion about the future of this repo https://github.com/bmadcode/BMAD-METHOD/discussions/169 for V4 and also if this idea fits into where this repo might go.

bmadcode:
I am still letting this idea bake around the templating. I think I like it, but have an alternative idea to keep the template itself clean, but add a metadata section to the top for the LLM instructions. Ill add a sample to this PR of what I am thinking before merging / get some feedback first

PR #186, opened 2025-06-08
Fix story-dod-checklist file extension

dracic:
Using latest v3, so noticed that Cursor allways searches story-dod-checklist.txt instead of md


dracic:
I changed only the extension in my first try, not having seen that the location was also wrong. So I just amended and force-pushed.

PR #231, opened 2025-06-16
fix: resolve installer 'Source and destination must not be the same' error

chrimage:
## Summary
- Fixes critical installer bug where `npx bmad-method install` fails with "Source and destination must not be the same" error
- Resolves path resolution issues when running installer via npx from npm cache
- Adds safety checks to prevent installation conflicts

## Problem
When users run `npx bmad-method install` with the default `.bmad-core` directory, the installer fails because:
1. The source path (`getBmadCorePath()`) resolves to the same location as the destination 
2. This happens because npx runs from the npm cache, causing path resolution conflicts
3. The `fs.copy()` operation fails when source and destination are identical

## Solution
**Updated `config-loader.js`:**
- Modified `getBmadCorePath()` to use `path.resolve()` for absolute path resolution
- Added comments explaining the npx execution context

**Updated `installer.js`:**
- Added safety check to detect when source and destination paths are identical
- Provides clear error message with guidance when paths conflict
- Uses `path.resolve()` for consistent absolute path handling

## Test plan
- [x] Tested installer fix locally with `node tools/installer/bin/bmad.js install`
- [x] Verified fix resolves the "Source and destination must not be the same" error  
- [x] Confirmed installer creates proper `.bmad-core` directory structure
- [x] Tested with different installation directory names to ensure flexibility

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/tools/installer/lib/config-loader.js
+++ b/tools/installer/lib/config-loader.js
@@ -78,7 +78,13 @@ class ConfigLoader {
 
   getBmadCorePath() {
     // Get the path to .bmad-core relative to the installer (now under tools)
-    return path.join(__dirname, '..', '..', '..', '.bmad-core');
+    // When running via npx, __dirname points to the installed package
+    // We need to find the actual .bmad-core directory that comes with the package
+    const packageRoot = path.join(__dirname, '..', '..', '..');
+    const bmadCorePath = path.join(packageRoot, '.bmad-core');
+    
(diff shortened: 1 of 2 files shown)


bmadcode:
Could not replicate - but i did remove the prompt from the installer to be .bmad-core and updated the isntaller to direct the user to enter a valid path now.

PR #234, opened 2025-06-17
Fix formatting in README.md

fabb:


bmadcode:
@fabb there was a problem with the auto formatter it kept changing this - fixed it which auto fixed the formatting.

PR #246, opened 2025-06-18
Allow core agents in expansions

sweetroll:
Hi, I was building an expansion pack that aimed at adjusting the a single agent in the BMAD method but realised the expansions could not import core agents. This PR adjusts the web-builder to read the agents from the yaml file and if an agent listed is not detected in the expansion it will import it from the core.

Happy for you to discard the PR if you'd like it implemented a different way but I think this would be a great feature and give more power to the expansions.

Love the project!

### DIFF
--- a/tools/builders/web-builder.js
+++ b/tools/builders/web-builder.js
@@ -12,6 +12,11 @@ class WebBuilder {
     this.templatePath = path.join(this.rootDir, 'bmad-core', 'templates', 'web-agent-startup-instructions-template.md');
   }
 
+  parseYaml(content) {
+    const yaml = require('js-yaml');
+    return yaml.load(content);
+  }
+
   async cleanOutputDirs() {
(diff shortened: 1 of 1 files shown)


bmadcode:
Cool great idea - the whole expansion project is a bit loosely defined - I will merge this in. Thanks @sweetroll - have been wondering if any would use or be interested in this expansion idea!

I am debating what the future organization is that makes sense as everything potentially in the future including what is currently the core could be a bundle or expansion.

bmadcode:
@sweetroll - I just pushed an additional update - I found that dependencies from the core also were not being included in the bundles. so for example a agent that used create-doc in the expansion, or an agent for the core - the dependencies from the core were getting skipped. I set it up so if there is a dependency in the expansion with the same name that will be in the bundle, or else it will pull it from the core.

PR #252, opened 2025-06-19
Clean up markdown nesting.

titocr:
Markdown files were difficult to follow due to nesting being off in some places.


bmadcode:
@titocr thank you - just cleaned this all up but it seems the linter is kicking in doing some dumb autoformatting again!

PR #288, opened 2025-07-02
Add bmad-mcp package

fabb:
This PR adds a bmad-mcp package that provides all BMAD resources via an MCP Server as MCP resources and MCP prompts.

See the README.md for more details how to use it.

This is work-in-progress.

It is only compatible with MCP clients that support MCP resources and prompts.

I've tested it so far in VSCode with Copilot and with the MCP inspector:

Screenshot 2025-07-02 at 20 53 29

Screenshot 2025-07-02 at 20 50 08

### DIFF
--- a/bmad-mcp/src/server.ts
+++ b/bmad-mcp/src/server.ts
@@ -1,0 +1,423 @@
+import { McpServer, ResourceTemplate } from "@modelcontextprotocol/sdk/server/mcp.js";
+import { StdioServerTransport } from "@modelcontextprotocol/sdk/server/stdio.js";
+import { ResourceProvider } from "./providers/resourceProvider.js";
+import type {
+  ReadResourceResult,
+  ListResourcesResult,
+  Resource,
+} from "@modelcontextprotocol/sdk/types.js";
+import type { Variables } from "@modelcontextprotocol/sdk/shared/uriTemplate.js";
+import type { BaseResource, PromptConfig, ResourceConfig } from "./types/index.js";
+import { load } from "js-yaml";
+
+export class BmadMcpServer {
+  private server: McpServer;
+  private transport: StdioServerTransport;
+  private resourceProviders: Map<string, ResourceProvider<BaseResource>>;
+
+  private readonly agentsResourceConfig: ResourceConfig = {
+    resourceType: "Agent",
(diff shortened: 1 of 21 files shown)


bmad-mcp/src/server.ts
```diff
@@ -0,0 +1,423 @@
...
+  private readonly agentsResourceConfig: ResourceConfig = {
+    resourceType: "Agent",
+    uriScheme: "bmad://agents",
+    directory: "agents",
+    nameSuffix: "",
+    description: "Individual BMAD agent content",
+    supportedExtensions: [".md"],
+  };
+
+  // TODO some bmad files contain instructions that are only relevant when using bmad-core as files and not as MCP resources, e.g. `IDE-FILE-RESOLUTION` in agent files - either change this in BMAD, or do some file content replacement here when serving resources
```
fabb, line 27: some bmad files contain instructions that are only relevant when using bmad-core as files and not as MCP resources, e.g. `IDE-FILE-RESOLUTION` in agent files - either change this in BMAD, or do some file content replacement here when serving resources

bmad-mcp/src/server.ts
```diff
@@ -0,0 +1,423 @@
...
+
+        dependencyResources = await this.getDependencyResources(parsedDependencies);
+      }
+
+      this.server.prompt(`BMAD ${resource.name}`, async () => {
+        try {
+          return {
+            description: `BMAD ${resource.name}`,
+            messages: [
+              // TODO also add separate prompts for each task of the agent, so they can be triggered directly - in that case it would be great to directly embed resources instead of resource_links to save requests (relevant for ai agents like copilot which deduct premium requests and not token count)
```
fabb, line 285: also add separate prompts for each task of the agent, so they can be triggered directly - in that case it would be great to directly embed resources instead of resource_links to save requests (relevant for ai agents like copilot which deduct premium requests and not token count)

bmadcode:
@fabb - I will test this out a bit after this weekend - wrapping up a project taking up a bit of time, have not forgotten this!

bmadcode:
@fabb I think you were reworking in light of updates this was out of sync with - will close for now and we can align on MCP feature for V5

PR #308, opened 2025-07-10
refactor(gemini-cli): change agent storage from multiple files to single

dracic:
- Update configuration to use .gemini/bmad-method/GEMINI.md directory instead of .gemini/agents/{agent}.md
- Implement new logic to concatenate all agent files into single GEMINI.md
- Add backward compatibility for existing settings.json (now it is not needed, but we won't delete it because it can have some other settings, for example project MCP, we just remove obsolete agents)
- Remove old agents directory and update related documentation
- Ensure all agent settings are properly loaded

Existing setup also has a bug - it was simply concatenating raw agent content, unlike other IDEs that format agent rules with structured sections. Thats why @agent calls weren't working as it should. 

Please test. Thanks!

### DIFF
--- a/tools/installer/config/install.config.yaml
+++ b/tools/installer/config/install.config.yaml
@@ -68,12 +68,13 @@ ide-configurations:
       # 4. Rules are stored in .clinerules/ directory in your project
   gemini:
     name: Gemini CLI
-    rule-dir: .gemini/agents/
-    format: context-files
+    rule-dir: .gemini/bmad-method/
+    format: single-file
+    command-suffix: .md
     instructions: |
       # To use BMad agents with the Gemini CLI:
-      # 1. The installer creates a .gemini/ directory in your project.
-      # 2. It also configures .gemini/settings.json to load all agent files.
+      # 1. The installer creates a .gemini/bmad-method/ directory in your project.
+      # 2. It concatenates all agent files into a single GEMINI.md file.
       # 3. Simply mention the agent in your prompt (e.g., "As @dev, ...").
       # 4. The Gemini CLI will automatically have the context for that agent.
   github-copilot:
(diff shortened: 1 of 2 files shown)


dracic:
Why all this:

https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/configuration.md

- **Hierarchical Loading and Precedence:** The CLI implements a sophisticated hierarchical memory system by loading context files (e.g., `GEMINI.md`) from several locations. Content from files lower in this list (more specific) typically overrides or supplements content from files higher up (more general). The exact concatenation order and final context can be inspected using the `/memory show` command. The typical loading order is:
  1.  **Global Context File:**
      - Location: `~/.gemini/<contextFileName>` (e.g., `~/.gemini/GEMINI.md` in your user home directory).
      - Scope: Provides default instructions for all your projects.
  2.  **Project Root & Ancestors Context Files:**
      - Location: The CLI searches for the configured context file in the current working directory and then in each parent directory up to either the project root (identified by a `.git` folder) or your home directory.
      - Scope: Provides context relevant to the entire project or a significant portion of it.
  3.  **Sub-directory Context Files (Contextual/Local):**
      - Location: The CLI also scans for the configured context file in subdirectories _below_ the current working directory (respecting common ignore patterns like `node_modules`, `.git`, etc.).
      - Scope: Allows for highly specific instructions relevant to a particular component, module, or subsection of your project.

dracic:
Strange:

```
✦ It looks like y`ou're trying to switch to the Scrum Master. As the developer, I can't change roles.

 To switch agents, you'll first need to exit my current persona by typing *exit. Then you can switch to the Scrum Master by typing *agent sm.
```

Edit: today no isses with 0.1.11

dracic:
So made a change `@agent -> *agent`. We shouldn't use `@` because of meaning of that to gemini hierarchical context system, that works not only for `GEMINI.md` or any file defined with `contextFilename`, but obviously for all files added to the context without a full path. That's why when you type `@dev` and have `@infra-devops-platform`, the latter will be activated.

bmadcode:
Thanks @dracic !

PR #311, opened 2025-07-13
fix: support Windows-style newlines in YAML block regex

### LINKED ISSUE #310: Getting an Empty .roomodes file on windows.
**Describe the bug**
I keep getting no modes when installing for Roo code, only an empty .roomodes file.

BMad Method version: 4:29:0

H-Ghanam:
## What

Updated the regular expression used to extract agentContent in setupRoo function to support both Unix (`\n`) and Windows (`\r\n`) newline styles.

## Why

When using the installer on Windows  it generates an empty .roomodes file which means it couldn't get the content correctly , `agentContent.match()` was returning `null` due to the regex expecting only Unix-style newlines (`\n`). Since Windows encodes line endings as `\r\n`, YAML blocks enclosed in triple backticks were not detected, breaking parsing functionality on Windows environments.

Fixes #310

## How

- Updated the regex pattern from `/```ya?ml\n([\s\S]*?)```/` to `/```ya?ml\r?\n([\s\S]*?)```/`
- Used `\r?\n` to support both newline conventions (`\n` and `\r\n`)
- Verified regex works across both Windows and Unix-style inputs

## Testing

Manually tested the updated regex on windows and linux with both `\n` and `\r\n` line endings.

### DIFF
--- a/tools/installer/lib/ide-setup.js
+++ b/tools/installer/lib/ide-setup.js
@@ -702,7 +702,7 @@ class IdeSetup {
         const agentContent = await fileManager.readFile(agentPath);
 
         // Extract YAML content
-        const yamlMatch = agentContent.match(/```ya?ml\n([\s\S]*?)```/);
+        const yamlMatch = agentContent.match(/```ya?ml\r?\n([\s\S]*?)```/);
         if (yamlMatch) {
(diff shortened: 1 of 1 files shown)


bmadcode:
thanks @H-Ghanam

PR #326, opened 2025-07-17
Update game-developer.md

ReJoy11:
This was solely done to protect the 2nd interface from colliding the whole set of textual glidings


bmadcode:
not sure I follow - but this is how it is in all of the agents - the yaml check seems to think its valid as is. Also - removing the S might make some llms complain it cannot find the exact match section since the instructions look for an S on the end.

PR #327, opened 2025-07-17
Fix `team-fullstack.txt` path in bmad-workflow-guide.md

oldirty:
This file was at a different location for me - is that expected?


bmadcode:
Thank you @oldirty - do you go by ODZ? :) I realize I do need to make a few more improvements to that section of the doc - good catch.

RIP ODB!

PR #338, opened 2025-07-20
fix: broken link ("See available AI agents") in README

yaksh1:
Before: 
when clicked on the link it was not redirecting to correct folder.

After:
Now it redirects to correct directory of available agents


manjaroblack:
I think the goal is to have an Available Agents section in the ReadMe that describes each agent instead of just linking to a folder.

yaksh1:
@bmadcode can you comfirm this comment @manjaroblack? If yes, I'll do the necessary updates. Thanks manjaro for the heads up.

yaksh1:
@bmadcode hey, any comments on this?

bmadcode:
Hey @yaksh1 sorry for the delay here, appreciate this!  I removed the table that was referring to as it felt like bloat. But. abig improvement to docs in general is in the works. But I think it makes sense for now this can merge as is

PR #359, opened 2025-07-25
feat(integration): claude code subagents

24601:
This is a very early attempt at properly implementing the BMAD-METHOD using Claude Codee's new subagents (https://docs.anthropic.com/en/docs/claude-code/sub-agents). 

It is still very much a WIP and I am refining the core integration and also making it so it will accommodate upstream updates gracefully and has proper integration with things like semantic integration with claude code so it can auto-call the agents as necessary in the workflow, etc. 

Opening as draft to get advice/visibility/testers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # BMad-Method: Universal AI Agent Framework
 
-**Come to discord (see below) [https://discord.com/channels/1377115244018532404/1398087195272806581](specifically this channel) to chat about this port of BMAD-METHOD**
+**Come to discord (see below) [specifically this channel](https://discord.com/channels/1377115244018532404/1398087195272806581) to chat about this port of BMAD-METHOD**
 
 
 [![Version](https://img.shields.io/npm/v/bmad-method?color=blue&label=version)](https://www.npmjs.com/package/bmad-method)


Anima-t3d:
There seem to be files committed from claude and bmad that seem to be "working" files and not related to actually the changes needed to implement the feature. Is that intended while in draft/final version or should it be cleaned up?

offendingcommit:
I immediately thought of this and came to see if anyone had already jumped on it!

ugurcancaykara:
same here

24601:
> There seem to be files committed from claude and bmad that seem to be "working" files and not related to actually the changes needed to implement the feature. Is that intended while in draft/final version or should it be cleaned up?

Those were test scripts that were my LLM as a judge scripts and figured I would add them as they are part of the test strategy I used to validate. 

There are some improvements/additions I made around things I saw in testing that the system was missing or not grounding well on I added. 

The issue templates etc I just kept for my work and will yank. Other ones you may want gone?

offendingcommit:
I'm guessing I'm not seeing the outputs yet that I'd expect to live in `.claude/`
That's still WIP?

LaloLalo1999 (changes requested):
The installer doesn't properly install the slash commands in the right folder. The hooks should be added too to the .claude folder. Sessions aren't getting initialized.

24601:
> The installer doesn't properly install the slash commands in the right folder. The hooks should be added too to the .claude folder. Sessions aren't getting initialized.

Yup, refactoring

24601:
> I'm guessing I'm not seeing the outputs yet that I'd expect to live in `.claude/` That's still WIP?

Yeah, def still WIP - playing through subagent turn-by-turn and testing strategy now.

ugurcancaykara:
Hey man, **for your attention,**
Check this out for the naming convention of agents: [redditAgentsNamingBuggyBehavior](https://www.reddit.com/r/ClaudeAI/comments/1ma4obp/claude_code_sub_agents_not_working_as_expected/)


can you check this behavior?

Anima-t3d:
Need to test if the agents can make use of MCPs, for example architect to use something like context7 for up to date documentation and package versions or a QA to use playwright MCP etc.

offendingcommit:
> Need to test if the agents can make use of MCPs, for example architect to use something like context7 for up to date documentation and package versions or a QA to use playwright MCP etc.

Yes, the agents are able to because I have my own agents setup in my project as "experts" in the different tools. You can limit the agents to specific tools to honor principle of least privilege. 

One concern/observation would be making the agents too opinionated. Is Context7 the direction? Maybe. But IMO, it would be better to keep these agents scoped to BMAD with documentation on how to easily add additional MCPs. 

Perhaps the install script could setup a project level `.mcp.json` with some helpful and _optional_ MCPs.

Anima-t3d:
I think it should be configurable, but for example I was talking to the architect (original bmad, not this draft implementation) and I asked it to use the context7 mcp for making sure the architecture document is using correct syntax etc, it told me it couldn't use it, so just wanted to highlight it here.

bmadcode:
subagent integration is going to be a bit tricker than anticipated IMO, easy to do it wrong I guess I mean. Thinking through still on a solid strategy on what makes sense as a subagent - it might not be a full agent but a simpler agent that really benefits from its own context space but returns a simple result for example.

RE specific tools, for the most part I do not want to bake any specific tools in unless there is a very compelling reason. But what I envision is during the installer at some point it will offer the user the option to configure optional MCP integrations - and if they choose them the agents will be instructed in certain ways to use certain tools in certain scenarios.

Just some food for thoughts - very preliminary ideas - would love other opinions also

AchintyaAshok:
Are you thinking that the MCP integrations for the tools would be for tools from specific servers and implementations like an mcp.json config basically for each subagent?

Charly6596:
For ideas check out how Agent OS uses agents. It is what you say @bmadcode specific tasks with clear input and output results

For example, to use git it spawns an agent
https://github.com/buildermethods/agent-os/blob/main/claude-code/agents/git-workflow.md?plain=1

"Context fetcher" to gather context from spec files and give it to the main agent
https://github.com/buildermethods/agent-os/blob/main/claude-code/agents/context-fetcher.md?plain=1

oxillix:
Why was this closed? will subagents ever be supported?

DevJake:
> Why was this closed? will subagents ever be supported?

Anything on this perchance @bmadcode? It would be a fantastic, albeit potentially challenging addition

Issue #340, opened 2025-07-20
Rename the names of the agents


PhunkyBob:
Hello, 

I like your project, but since I have problem remembering surnames, I usually change the names of the agents to something more intuitive (for me).

Here are my suggestions: 
- **Anne ALYST** is my "analyst"
- **Archibald TEKT** is my "architect"
- **Dave LOPER** is my "developer"
- **Pedro MANAGER** is my "pm"
- **Prosper DUCT-OWNER** is my "po"
- **Ulysse XHUI** is my "ux-expert"

dracic:
I kinda like this mnemonic idea

bmadcode:
I like this - thinking about this some more - might be nice if the installer had an advanced option to pick your initial team (provide a new or accept a default) and also during install customize personalities (sassy pirate, emotionless cyborg...)

Also - last time I changed the names to something a bit more like what you are proposing, I got a lot of complaints that people were used to the other names and so I changed them back :)

Let me save this for the v5 roll out with install agent customization options - and will go with a default scheme as you propose here.

If interested in building this out - it would be nice if it did something like:
Would you Like to Customize your Agile Team?
Configuring Agile Team:
Analyst Name: Anne (default can type over)
Custom Personality: (default none)

---

there are more install customizations I am also planning to add in the future, specifically around the dev agent and customizing it a bit more to specific tech stacks of the intended project if people would like to make it more (optionally) specialized.

rpgdev:
The names proposed sound goofy to me, while the current ones are neutral and natural. I’d go with letting people customize the names if anything... but the idea seems overkill in general and unnecessary.

PR #400, opened 2025-08-05
feat(installer): Add interactive agent & their personality customization

Mbosinwa:
### What

This introduces a new feature to the installer that allows for the interactive customization of agent personalities.

### Why
To allow users to easily customize the names and personas of their agents during the installation process, this change introduces a new interactive step, that allows more personalized and engaging agent interactions right from the start. especially when you want your ai agent to have names and personas similar to your real life team members. 
This feature was added in response to feedback in [issue #340](https://github.com/bmadcode/BMAD-METHOD/issues/340#issuecomment-3097090308).

### How

* **`bmad.js`**: Added a new step to the installer to ask users if they want to customize their agents.
    * If the user agrees, they can provide a custom name and choose from pre-defined personality templates for each agent.
    * Generic fallback personas like "Sassy Pirate" and "Sabinwa" have been added for un-templated agents.
    * The user can also write their own custom personality instructions, with guidance provided on how to do so effectively.
* **`installer.js`**: Added functions to handle the application of these customizations to the agent's configuration file.
    * The installer now applies these customizations during a "full" installation.
* **`review-story.md`**: Updated the "Reviewed By" field to use a placeholder for the agent's name and title.

### Testing
* Verified that the installer prompts the user for agent customization.
* Tested that custom names and personalities are correctly applied to the agent configuration.
* Confirmed that the fallback personas are applied when no custom personality is selected.
* Validated that the installer handles both "full" and standard installations correctly with the new customization options.

# BMad Custom Agent Personality in Used
image

### DIFF
--- a/bmad-core/tasks/review-story.md
+++ b/bmad-core/tasks/review-story.md
@@ -81,7 +81,7 @@ After review and any refactoring, append your results to the story file in the Q
 ## QA Results
 
 ### Review Date: [Date]
-### Reviewed By: Quinn (Senior Developer QA)
+### Reviewed By: [Your agent name] ([Your agent title])
 
 ### Code Quality Assessment
 [Overall assessment of implementation quality]
--- a/tools/installer/bin/bmad.js
+++ b/tools/installer/bin/bmad.js
@@ -289,6 +289,155 @@ async function promptInstallation() {
     }
   }
 
+
+   // Agent Customization Section
+    const { customizeAgents } = await inquirer.prompt([
+        {
+            type: 'confirm',
+            name: 'customizeAgents',
+            message: 'Would you like to customize your BMad Agile Agents and their personalities?',
+            default: false
+        }
+    ]);
+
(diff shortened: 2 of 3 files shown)


bmadcode:
love it, has been on my list for a long time to get to! A few suggestions and a question or two.

Lets move the premade samples out to a separate json file.
How will this work with expansion packs?
If you run npm run build - does the dist output get generated still with defaults or placeholders?
need to consider when there is a new version and they use the installer to do an upgrade - does it blow away their customization?

I like the idea potentailly also of premade theme packs potentially. so maybe a star trek themed pack, or star wars, Naruto whatever that prealigns the personas with a character from the universe.

One thing we also might want to test. will want to make sure for example pirate pete doesnt acutally write architecture or code comments as a pirate - so will need a command in all agent files to avoid that somehow?

bmadcode:
closing as this is already in progress in v5 version to be released soon.

PR #405, opened 2025-08-06
Fix Incorrect Elicitation  Agents Options

DasoTD:
Summary:
This PR resolves a mismatch between the instructional prompt and the actual option numbering in the "Advanced Elicitation Options" list. Previously, the prompt read:

“Choose a number (0–8) or 9 to proceed…”

However, the list provided 10 options numbered 1–10, where "Proceed" was option 10, not 9.

Fix Implemented:

Updated the prompt text to:

“Choose a number (1–9) or 10 to proceed…”

This aligns the prompt with the visible list and removes confusion for users, especially those unfamiliar with programming-based zero indexing.

Why It Matters:

Improves user clarity and experience

Prevents selection errors due to misaligned instructions

Makes the UI/UX feel more polished and professional

### DIFF
--- a/bmad-core/tasks/advanced-elicitation.md
+++ b/bmad-core/tasks/advanced-elicitation.md
@@ -80,7 +80,7 @@ When invoked after outputting a section:
 
 ```text
 **Advanced Elicitation Options**
-Choose a number (0-8) or 9 to proceed:
+Choose a number (0-9) or 10 to proceed:
 
 0. [Method Name]
 1. [Method Name]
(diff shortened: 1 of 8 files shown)


bmadcode:
This seems to have changed it to 0 to 9 instead of 1 to 10 in the PR - so it will still be nondeterministic and create confusion in different models it looks like. Usually that can be witnessed in custom gem vs custom gpt for example.

bmadcode:
closing this, as rolling out a different change to resolve this today that will improve the experience

PR #408, opened 2025-08-07
Flatten venv  & many other bins dir fix

alior101:
This fix removes many incorrectly included files into the xml codebase. For my project it brough the xml from 650 MB (i.e. unusable) to 3MB.


bmadcode:
good catch here - thanks

PR #413, opened 2025-08-08
Add new BMAD-compliant claude code agents

dogank85:
Hi Brian,

I've been exploring how to use BMAD agents as Claude Code sub-agents, moving beyond the custom slash commands.
  
With the help of the Analyst and QA agents, I've converted the core BMAD personas into sub-agent format. My goal was
to ensure they remain BMAD-compliant in this new structure. I also used a meta-agent, inspired by a video from IndyDevDan, to help generate the agent definitions. I would appreciate your review of this implementation. I'm particularly interested in your feedback on whether I was successful in maintaining BMAD compliance and any suggestions you might have for improvement. Thank you!

### DIFF
--- a/.claude/agents/analyst.md
+++ b/.claude/agents/analyst.md
@@ -1,0 +1,80 @@
+---
+name: analyst
+description: Use proactively for data analysis, business intelligence, market research, and analytical reporting tasks. Specialist for analyzing datasets, generating insights, creating reports, and conducting market research.
+tools: Read, Write, MultiEdit, Grep, Glob, LS, Task, TodoWrite, Bash, WebSearch, WebFetch
+model: sonnet
+color: purple
+---
+
+# Purpose
+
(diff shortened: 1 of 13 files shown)


bmadcode:
Hi  @dogank85 - thank you for the submission. From my testing with subagents so far, I do not thing the agents as is are suited to be converted completely to a subagent - I tried similar and it was inconsistent in its ability to maintain the plot with its parent thread and perform multishot conversations back and forth with the user when all within a subprocess.

I could be wrong as that was on the original release - but it needs more testing.

Where I think the benefit of subagents will be useful is when they are scoped to do a single thing really well and kick in automatically.

One potential idea as an example just taking Dev agent:

running normal dev agent, there could be a subagent that is a linting master, a subagent that is a unit test master, or a documentation expert etc that knows how to ensure docs remain updated or comments are clear.

PR #423, opened 2025-08-11
Add update-check command

AaronHood:
**What**

Implemented a new update-check command that checks for available BMad Method updates by querying the npm registry and comparing versions using semver.

**Why**

Users need a way to check if newer versions of BMad Method are available without having to manually check npm or perform actual updates, enabling informed decision-making about when to upgrade.

**How**
-   Makes HTTPS request to npm registry API (https://registry.npmjs.org/bmad-method/latest) to fetch latest version metadata
-   Implements comprehensive error handling with HTTP status validation, network error catching, JSON parsing protection, and 30-second timeout with cleanup
-   Uses semver library to safely compare current version against latest available version and displays appropriate status messages

**Testing** 
- Modify 'version' in package.json to test >, < and = for version scenarios 
- Disconnect network to emulate timeout
- Change `https://registry.npmjs.org/${packageName}/latest` value to different URLs to produce different status codes


manjaroblack (changes requested):
This looks good. I suggest adding some direction for the user for how to update.

AaronHood:
Added directions for the users along with updated the text colors to align with rest of the cli

tools/installer/bin/bmad.js
```diff
@@ -86,6 +90,60 @@ program
...
+      res.on('end', () => {
+        try {
+          // Parse npm registry response and extract version
+          const latest = JSON.parse(data).version;
+          
+          // Compare versions using semver
+          if (semver.gt(latest, version)) {
+            console.log(chalk.bold.blue(`⚠️  ${packageName} update available: ${version} → ${latest}`));
+            console.log(chalk.bold.blue('\nInstall latest by running:'));
+            console.log(chalk.bold.magenta(`  npm install -g ${packageName}@latest`));
```
manjaroblack: this is a global install command, bmad should be installed per project.

manjaroblack (changes requested):
Suggest changing the update command to update bmad for the project, not global.

tools/installer/bin/bmad.js
AaronHood: good catch.  corrected

manjaroblack (approved):

bmadcode:
thanks @AaronHood

PR #428, opened 2025-08-12
fix(docs): fix broken link in GUIDING-PRINCIPLES.md

yongjip:
## What

Fix broken link to BMad Document Template in GUIDING-PRINCIPLES.md

## Why

The original link was incorrect and didn’t resolve properly.

## How

- Updated the markdown link to use the correct relative path.

## Testing

Previewed the markdown to confirm the link now works.


manjaroblack (approved):
Good catch. Verified fix.

PR #440, opened 2025-08-15
fix(flattener): fix argument parsing logic to work with provided input/output paths

abdalrohman:
The flattener tool had inverted logic that would show an error message when users provided arguments (-i/--input or -o/--output) instead of using them. This fix removes the incorrect error message, allowing the tool to work correctly in both modes:

- When arguments are provided: uses the specified paths

- When no arguments are provided: works in interactive mode as before


manjaroblack:
This fix is already in an existing PR.

Issue #478, opened 2025-08-18
No BMad installation found in current directory tree


moyger:
Bug 1: status can’t detect a local install
After installing BMAD Method, running npx bmad-method status from the repo prints:
No BMad installation found in current directory tree.
BMAD is installed in the project (created a .bmad-core/ folder), but status does not detect it.

**Steps to Reproduce**
1. In a project folder, install BMAD Method (e.g., via the README instructions).
2. Confirm the installer created a hidden folder (in my case .bmad-core/) in the project root.
3. From the same project root, run: npx bmad-method status.
4. Observe the message: No BMad installation found in current directory tree.

dracic:
I can confirm that status doesn't work

manjaroblack:
I'm not certain the fix you linked will resolve the issue, but I will look at this issue and make sure it gets resolved today.

dracic:
> I'm not certain the fix you linked will resolve the issue, but I will look at this issue and make sure it gets resolved today.

It wont if you don't edit findInstallation(), but since this is similar stuff perhaps you can just push to your PR. It's all about originalCwd.

bmadcode:
Thank you for taking the time to report this issue during the v4 era of BMAD. We truly appreciate your feedback and contribution to making BMAD better.

We've recently released v6, which represents a major evolution of the BMAD methodology with significant architectural changes and improvements. As part of this transition, we're closing older v4-related issues to maintain a clean issue tracker focused on the current version.

If you're still experiencing this issue or have concerns with v6, please feel free to open a new issue with details specific to the current version. We'll be happy to address it promptly.

Thank you again for your support!

Issue #474, opened 2025-08-18
Installation failed: glob is not defined


thiagocarvalhodev:
**Describe the bug**
Running the installer with `npx bmad-method install` on macOS fails at the end with:

```
✖ Installation failed
Installation failed: glob is not defined
```

It looks like the installer is referencing `glob` without importing it.

---

**Steps to Reproduce**

1. Run on a MacBook Pro M2 (Apple Silicon):

   ```bash
   npx bmad-method install
   ```
2. Select options during the prompts:

   * Project dir: `/Users/thiagocarvalho/.npm/_npx/.../node_modules/bmad-method`
   * Install: **BMad Agile Core System (v4.39.2) .bmad-core**
   * PRD sharded: **Yes**
   * Architecture sharded: **Yes**
   * IDEs: **Cursor**, **Claude Code**
   * Web bundles: **All available bundles**
3. Installer creates rules and web bundles successfully, but then fails with the above error.

---

**PR**
Not working on a fix yet, but the likely solution is to ensure `glob` is imported/required correctly (Node 20+ ESM requires explicit `import { glob } from "glob"`).

---

**Expected behavior**
Installer should complete without errors, with `.cursor/rules/bmad` and web bundles configured.

---

**Please be Specific if relevant**

* **Model(s) Used:** N/A
* **Agentic IDE Used:** Cursor, Claude Code
* **WebSite Used:** N/A
* **Project Language:** Node.js
* **BMad Method version:** v4.39.2

---

**Screenshots or Links**
Console output:

```
✓ Installed all web bundles to: .../bmad-method/web-bundles
✓ Created Cursor rules in .../bmad-method/.cursor/rules/bmad
✖ Installation failed
Installation failed: glob is not defined
```

---

**Additional context**

* Device: MacBook Pro M2 (Apple Silicon)
* macOS: 15.6 (24G84)
* Node.js: v24.5.0
* npm: v11.5.1

mylesdebastion:
Same, I had selected cursor and claude web packages be available.

mylesdebastion:
I skipped web packages and it seemed to finish installing

manjaroblack:
Checking.

daniel-chen-git:
I tried removing the installation folder and the ~/.npm/_npx installation folder, then re-ran npx bmad-method install, but the error "Installation failed
Installation failed: glob is not defined" still occurred. 
The same error also occurred after running npm install -g glob and then continuing the installation.

[...]

PR #480, opened 2025-08-19
fix: honor original working directory when running npx installer and searching for task files

manjaroblack:
fixed the default install dir to retain the original working directory after npx installs the bmad-method package


dracic:
@manjaroblack, since youre on this one... The findInstallation() method uses process.cwd() to start searching for .bmad-core, but when running via npx, the working directory is different from where the user executed the command. Can you add the same logic to findInstallation() method to use the same directory resolution logic as the install() method.

I think it will solve this: #478

schreidify:
In order to fix the glob is not defined installation error (#474 ), you will also have to change line: 180 on tools/yaml-format.js:

`const matches = glob.sync(argument);`

Issue #490, opened 2025-08-21
Agent tends to create new folder when working on new story tasks


Wang-tianhao:
**Describe the bug**
When starting a new story, it happened to me several times, instead of working from existing src/xxx, the agent created a entirely new folder to start the work.

**Steps to Reproduce**
I can reproduce this for my project for 1.2 to 1.4 until I added a new instruction into dev.md

**PR**
Suggested prompt worked for me. I'll open a PR.

**Expected behavior**
Agent would continue to work from existing folder structure.

**Please be Specific if relevant**
Model(s) Used: Opus 4.1
Agentic IDE Used: Claude Code
WebSite Used:
Project Language: Javascript
BMad Method version: 4

**Screenshots or Links**
If applicable, add screenshots or links (if web sharable record) to help explain your problem.

**Additional context**
Add any other context about the problem here. The more information you can provide, the easier it will be to suggest a fix or resolve

Wang-tianhao:
PR https://github.com/bmad-code-org/BMAD-METHOD/pull/491

PR #491, opened 2025-08-21
Update dev.md

Wang-tianhao:
To avoid AI creating new folder for a new story tasks. Related to https://github.com/bmad-code-org/BMAD-METHOD/issues/490


dracic:
Agree on this one. I used to put this in rules in Cursor.

PR #504, opened 2025-08-24
Fixed: "glob" is not defined

peterma:
In the getCoreTaskIds() function, two places use the "glob" variable, so the definition/reference of "glob" should be moved to the function-level instead of inside an if-statement scope.


ReJoy11 (approved):

peterma:
@bmadcode Could you please approve this PR, it's a simple fix and it has been blocking some people installing this fantastic project.

peterma:
This PR will fix this issue:
https://github.com/bmad-code-org/BMAD-METHOD/issues/474 Title: Installation failed: glob is not defined

PR #521, opened 2025-08-29
fix: add blank line to fix YAML parsing in sm.md slash command

alexeyv:
## What
  Added blank line after core_principles list in sm.md to fix YAML parsing

  ## Why
  The /sm slash command was failing to register in Claude Code due to invalid YAML structure - missing blank line
  caused parser to treat comment as part of list item

  ## How
  - Added blank line after line 48 in bmad-core/agents/sm.md
  - Separates YAML list from comment, allowing proper parsing

  ## Testing
  Verified YAML structure is now valid and /sm command registers properly


alexeyv:
Actually, the YAML was valid

alexeyv:
Sorry, this was wrong. The YAML block in sm.md is valid, the reason I saw /sm not registering as a slash-command in CC is something else.

PR #523, opened 2025-08-29
fix: Template file extension in validation next story steps

Drilmo:


leonfs:
In line 20 of the same modified file there is also a reference to the template and using the `.md` file extension.

Issue #363, opened 2025-07-25
Gemini cli - custom commands


dracic:
Custom commands are now supported in Gemini CLI through toml configuration files:

https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/commands.md#custom-commands

So we should remove GEMINI.md and do a setup like we have in Claude Code. I plan to do this after I come back from vacation, but if there is someone eager to do this in the meantime you're welcome!

dracic:
Update: 

https://github.com/google-gemini/gemini-cli/issues/5039

Waiting this to be fixed, and will make a PR after I test it.

bmadcode:
Thanks @dracic !

chi7cha7rito:
@dracic hi brother, this https://github.com/google-gemini/gemini-cli/issues/5039 was fixed. so can you help to make a PR. Very thanks.

yoav0gal:
PR: https://github.com/bmad-code-org/BMAD-METHOD/pull/549
 
Currently, the search in gemini-cli is based on the first letters and the commands are /agents:pm, so in order to search, one needs to look for agents instead of just pm.

Can be hacked, so commands will be just /pm or /create-doc - but I don't like it.

dracic:
> [@dracic](https://github.com/dracic) hi brother, this [google-gemini/gemini-cli#5039](https://github.com/google-gemini/gemini-cli/issues/5039) was fixed. so can you help to make a PR. Very thanks.

Sorry @chi7cha7rito, this post of yours went under my radar.

bmadcode:
is this issues replaced with a newer one?

dracic:
This can be closed. Resolved with #549

Issue #541, opened 2025-09-02
Agent prompts claim to define shorthand /commands in Claude Code, but don't


alexeyv:
**Describe the bug**
agent prompts claim to create shorthand /commands in Claude Code, but they don't

**Steps to Reproduce**
Install BMAD in an empty directory
* start `claude`
* type `/sm *help<Enter>`
  - expected: `Bob - Scrum Master Help Menu ... etc`
  - actual:  `Unknown slash command: sm`
 * but type `/BMad:agents:sm *help` and it does what is expected
* now type `/clear` (IMPORTANT) , then `/dev *help`
  - expected: ideally, `James - Full Stack Developer`; if not that, then ``Unknown slash command: dev` to go along with `Unknown slash command: sm`
  - actual: `⏺ I'll help you learn about Claude Code commands and features. Let me fetch the most current
  help information from the official documentation.` <-- that's Claude responding to `/dev` without loading commands/BMad/agents/dev.md. If you forgot to do `/clear`, you'd be getting sm's *help instead

Every other agent except dev.md behaves like sm.md

**Expected behavior**
`/sm *help<Enter>` -> `Bob - Scrum Master Help Menu ... etc`

**Suggested fix**
install agent files directly to .claude/commands
OR
replace `# /sm Command` with `# /BMad:agents:sm Command` in agent files. Which is what they actually define in Claude, when installed into .claude/BMad/commands/agents/

**Please be Specific if relevant**
Model(s) Used:
Agentic IDE Used:
WebSite Used:
Project Language:
BMad Method version:

**Additional context**
This is a "straight to the point" version of https://github.com/bmad-code-org/BMAD-METHOD/issues/522

alexeyv:
The practical problem this created for me was a couple of days of running /dev *develop-story so-and-so, and being pissed off that it solved the story, but never followed the prescribed workflow. Until I realized that /dev was an undocumented thing in Claude Code, and started invoking `/BMad:agents:dev *develop-story so-and-so` instead.

dracic:
Even without a native `/dev` command, installing a suite of commands directly into the top level is not a scalable practice. The namespacing you see (`/BMad:agents:dev`) is a deliberate architectural choice for several key reasons:

/dev is just an example here.

[...]

PR #549, opened 2025-09-05
added gemini cli custom commands!

yoav0gal:
issue: https://github.com/bmad-code-org/BMAD-METHOD/issues/363 

Comment: Currently, the search in gemini-cli is based on the first letters and the commands are /agents:pm, so in order to search, one needs to look for agents instead of just pm. 

Can be hacked, so commands will be just /pm or /create-doc - but I don't like it.

### DIFF
--- a/tools/installer/lib/ide-setup.js
+++ b/tools/installer/lib/ide-setup.js
@@ -690,6 +690,7 @@ class IdeSetup extends BaseIdeSetup {
 
   async getCoreTaskIds(installDir) {
     const allTaskIds = [];
+    const glob = require('glob');
 
     // Check core tasks in .bmad-core or root only
     let tasksDir = path.join(installDir, '.bmad-core', 'tasks');
@@ -698,7 +699,6 @@ class IdeSetup extends BaseIdeSetup {
     }
 
     if (await fileManager.pathExists(tasksDir)) {
-      const glob = require('glob');
       const taskFiles = glob.sync('*.md', { cwd: tasksDir });
       allTaskIds.push(...taskFiles.map((file) => path.basename(file, '.md')));
     }
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ The BMAD-METHOD™ includes a powerful codebase flattener tool designed to prepa
 ### Features
 
 - **AI-Optimized Output**: Generates clean XML format specifically designed for AI model consumption
-- **Smart Filtering**: Automatically respects `.gitignore` patterns to exclude unnecessary files
+- **Smart Filtering**: Automatically respects `.gitignore` patterns to exclude unnecessary files, plus optional project-level `.bmad-flattenignore` for additional exclusions
 - **Binary File Detection**: Intelligently identifies and excludes binary files, focusing on source code
 - **Progress Tracking**: Real-time progress indicators and comprehensive completion statistics
 - **Flexible Output**: Customizable output file location and naming
@@ -170,6 +170,18 @@ The generated XML file contains your project's text-based source files in a stru
 - File discovery and ignoring
   - Uses `git ls-files` when inside a git repository for speed and correctness; otherwise falls back to a glob-based scan.
   - Applies your `.gitignore` plus a curated set of default ignore patterns (e.g., `node_modules`, build outputs, caches, logs, IDE folders, lockfiles, large media/binaries, `.env*`, and previously generated XML outputs).
+  - Supports an optional `.bmad-flattenignore` file at the project root for additional ignore patterns (gitignore-style). If present, its rules are applied after `.gitignore` and the defaults.
+
+##### `.bmad-flattenignore` example
+
+Create a `.bmad-flattenignore` file in the root of your project to exclude files that must remain in git but should not be included in the flattened XML:
+
+```text
+seeds/**
+scripts/private/**
+**/*.snap
+```
+
 - Binary handling
   - Binary files are detected and excluded from the XML content. They are counted in the final summary but not embedded in the output.
 - XML format and safety
--- a/docs/working-in-the-brownfield.md
+++ b/docs/working-in-the-brownfield.md
@@ -5,7 +5,13 @@
 > Gemini Web's 1M+ token context window or Gemini CLI (when it's working) can analyze your ENTIRE codebase, or critical sections of it, all at once (obviously within reason):
 >
 > - Upload via GitHub URL or use gemini cli in the project folder
-> - If working in the web: use `npx bmad-method flatten` to flatten your project into a single file, then upload that file to your web agent.
+> - If working in the web: use `npx bmad-method flatten` to flatten your project into a single file, then upload that file to your web agent. To exclude additional files that must remain in git but shouldn't be sent to the AI, add a `.bmad-flattenignore` file at the project root (gitignore-style), e.g.:
+>
+> ```text
+> seeds/**
+> scripts/private/**
+> **/*.snap
+> ```
 
 ## What is Brownfield Development?
 
--- a/tools/flattener/ignoreRules.js
+++ b/tools/flattener/ignoreRules.js
@@ -154,9 +154,11 @@ async function parseGitignore(gitignorePath) {
 async function loadIgnore(rootDir, extraPatterns = []) {
   const ig = ignore();
   const gitignorePath = path.join(rootDir, '.gitignore');
+  const flattenIgnorePath = path.join(rootDir, '.bmad-flattenignore');
   const patterns = [
     ...(await readIgnoreFile(gitignorePath)),
     ...DEFAULT_PATTERNS,
+    ...(await readIgnoreFile(flattenIgnorePath)),
     ...extraPatterns,
   ];
   // De-duplicate


dracic:
> issue: #363
> 
> Comment: Currently, the search in gemini-cli is based on the first letters and the commands are /agents:pm, so in order to search, one needs to look for agents instead of just pm.
> 
> Can be hacked, so commands will be just /pm or /create-doc - but I don't like it.

I made a PR for Gemini CLI implementing case insensitive fuzzy command matching, and it will be in 4.x. Next week I suppose.

https://github.com/google-gemini/gemini-cli/releases/tag/v0.4.0-preview

So maybe we should do the same setup as in Claude Code IDE setup? At the moment it is

```
BMad:agents:agentname
Bmad:tasks:taskname
```

 ```
 claude-code:
    name: Claude Code
    rule-dir: .claude/commands/BMad/
    format: multi-file
    command-suffix: .md
    instructions: |
      # To use BMad agents in Claude Code:
      # 1. Type /agent-name (e.g., "/dev", "/pm", "/architect")
      # 2. Claude will switch to that agent's persona
  crush:
```

File: tools/installer/config/install.config.yaml [hunk 1]
```diff
@@ -78,15 +78,15 @@ ide-configurations:
...
+    rule-dir: .gemini/extensions/bmad/
+    format: multi-file
+    command-suffix: .toml
     instructions: |
       # To use BMad agents with the Gemini CLI:
-      # 1. The installer creates a .gemini/bmad-method/ directory in your project.
-      # 2. It concatenates all agent files into a single GEMINI.md file.
-      # 3. Simply mention the agent in your prompt (e.g., "As *dev, ...").
-      # 4. The Gemini CLI will automatically have the context for that agent.
+      # 1. The installer creates a `bmad` extension in `.gemini/extensions/`.
```
dracic: `.gemini/extensions` or `./gemini/commands` ?

tools/installer/lib/ide-setup.js
```diff
@@ -1208,97 +1208,90 @@ class IdeSetup extends BaseIdeSetup {
...
-    }
-
-    // Get all available agents
-    const agents = await this.getAllAgentIds(installDir);
-    let concatenatedContent = '';
-
+  async setupGeminiCli(installDir, selectedAgent) {
+    const ideConfig = await configLoader.getIdeConfiguration('gemini');
+    const extensionDir = path.join(installDir, ideConfig['rule-dir']);
+    const baseCommandsDir = path.join(extensionDir, 'commands');
```
dracic: So it is `commands`

tools/installer/config/install.config.yaml
```diff
@@ -78,15 +78,15 @@ ide-configurations:
       # 4. Rules are stored in .clinerules/ directory in your project
   gemini:
     name: Gemini CLI
-    rule-dir: .gemini/bmad-method/
-    format: single-file
-    command-suffix: .md
+    rule-dir: .gemini/extensions/bmad/
```
dracic: `.gemini/extensions` or `./gemini/commands `?

dracic:
Also let's look at:

https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/commands.md#4-injecting-file-content-with-

This is an opportunity for a simpler setup where the agent's content would be inserted directly using this directive from the .bmad-core directory. So in that case dev custom command would be:

```
# In: <project>/.gemini/commands/BMad/agents/dev.toml
# Invoked via: /BMad:agents:dev

description = "Load fullstack dev agent"
prompt = """
@{.bmad-core/agents/dev.md}
"""
```

If you have the time and willingness, I look forward to your PR.

yoav0gal:
Hahaha, I was digging into the code and saw that someone already did something for gemini-cli. Just realized that's you 🤣 
image


> So maybe we should do the same setup as in Claude Code IDE setup? At the moment it is

Yep took insperations from that in the pr its:

/agents:agentname -  it have a bmad indication next to it as well. 

image

lmk if you think we should add the bmad: as well - I think its overkill but IDK, something like bmand-agents:agent-name and bmad-tasks:can also work as well.

make the call XD

dracic:
Let's keep the same or similar setups within similar IDE solutions. And qwen will have the same setup. Fuzzy matching makes this easy. 

`/dev` and tab

We had that discussion: #541

yoav0gal:
> Also let's look at:
> 
> [google-gemini/gemini-cli@`main`/docs/cli/commands.md#4-injecting-file-content-with-](https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/commands.md?rgh-link-date=2025-09-05T12%3A15%3A35Z#4-injecting-file-content-with-)
> 
> This is an opportunity for a simpler setup where the agent's content would be inserted directly using this directive from the .bmad-core directory. So in that case dev custom command would be:
> 
> ```
> # In: <project>/.gemini/commands/BMad/agents/dev.toml
> # Invoked via: /BMad:agents:dev
> 
> description = "Load fullstack dev agent"
> prompt = """
> @{.bmad-core/agents/dev.md}
> """
> ```
> 
> If you have the time and willingness, I look forward to your PR.



Yep, thats cool. 

The on the flight  loading is interesting: 
pro: addpts when you are changing the agent definition yourself.
con: slower, more work - on every prompt copy paste - less observabilty (I don't think it's a biggie) 


What are your thoughts? 
@dracic  - Make a call here as well It should not take long implement

dracic:
> con: slower, more work - on every prompt copy paste - less observabilty (I don't think it's a biggie)

What do you mean by "copy paste" in this case?

yoav0gal:
> > con: slower, more work - on every prompt copy paste - less observabilty (I don't think it's a biggie)
> 
> What do you mean by "copy paste" in this case?

From my understanding of gemini-cli (not a good one) 

``` txt
This is myfile.txt

```

```
this prompt @{myfile.txt} 
```

would get translated in the gemini-cli react application to:

```
this prompt This is myfile.txt 
```

and the calculation to do that runs every time you call a prompt with @{}.

I can make the argument that because we don't change the agent command, it should be pre-computed. 

But honslty, I dont think its a big deal. just thinking out loud.

yoav0gal:
> Let's keep the same or similar setups within similar IDE solutions. And qwen will have the same setup. Fuzzy matching makes this easy.
> 
> `/dev` and tab
> 
> We had that discussion: #541


Not sure I got you right: 

so here you want to have the folder structure to be:

.gemini/extensions/bmad/commands/bmad/agents

instead of 

.gemini/extensions/bmad/commands/agents


so when you write the command it would look: bmad:agents:sm  insted of agents:dev ?

dracic:
The point of file injection is - why would you construct whole agent definition via complex JS function in

`.gemini/commands/BMad/agents/dev.toml`

when you can just insert it dynamically from 

`.bmad-core/agents/dev.md`

This is not meant to change the agent's definition dynamically, but rather to avoid having the same content in two places if we don't have to. I think Brian has an idea for v5 how to add possibility to add something to agent definition. The .toml files are not re-read or re-parsed on every command invocation. Try adding or changing a custom command (or inserted file) while running the Gemini CLI and see if you'll have a change. 

So:

1.  **Startup Initialization**: The application's entry point is the `main` function in `gemini.tsx`. This function orchestrates the startup sequence of the CLI.

2.  **Command Service Creation**: During startup, a `CommandService` instance is created. This service is responsible for discovering and loading all slash commands for the CLI. The creation is handled by the static `create` method of the `CommandService` class.

3.  **Command Loading**: The `CommandService.create` method takes an array of command loaders. One of these loaders is the `FileCommandLoader`, which is responsible for custom commands defined in `.toml` files.

4.  **File Scanning and Parsing**: The `FileCommandLoader.loadCommands` method is called only once during this startup phase. It scans the user, project, and extension directories for any `.toml` files, reads them, parses their content, and converts them into executable `SlashCommand` objects.

5.  **In-Memory Storage**: The loaded commands are then stored in a `readonly` array within the `CommandService` instance.

dracic:
> > Let's keep the same or similar setups within similar IDE solutions. And qwen will have the same setup. Fuzzy matching makes this easy.
> > `/dev` and tab
> > We had that discussion: #541
> 
> Not sure I got you right:
> 
> so here you want to have the folder structure to be:
> 
> .gemini/extensions/bmad/commands/bmad/agents
> 
> instead of
> 
> .gemini/extensions/bmad/commands/agents
> 
> so when you write the command it would look: bmad:agents:sm insted of agents:dev ?

No. Custom commands location is in `.gemini/commands/`

For bmad if we want the same setup like Claude Code we should have:

```
.gemini/commands/BMad/agents/agentname.toml
.gemini/commands/BMad/tasks/taskname.toml
```

With fuzzy matching you don't have to start typing `/BMad`, `/dev` will give you a list of all commands matching "dev"

yoav0gal:
cool,
I'll draft a pr.
You want to go for BMad for all or bmad (currently, a bunch of things are in bmad and some are BMad)


I tend to like bmad more for the folder

yoav0gal:
@dracic  - made the changes we discussed.

dracic:
> cool, I'll draft a pr. You want to go for BMad for all or bmad (currently, a bunch of things are in bmad and some are BMad)
> 
> I tend to like bmad more for the folder

Not my preference, just if Claude has BMad currently, let's go with it. I'll rather have consistency if it doesn't break anything, and then if we change our mind, let's change all at once.

dracic:
Thanks @yoav0gal !

yoav0gal:
Cool,
updated to BMad.

I asked because the rest is mostly bad (cursor, Auggie) 
Others with a signal file are using bmad. other multi-file/chats are using a variation of bmad (cline, roo)
We should probably change those as well if we are going for BMad (later)

BTW - in the gemini-cli the readPathFromWorkspace function does run every time (checked that) - I still don't think its an issue. i think keeping the @{} is ok


can we move Gemini CLI higher on the select menu?

dracic:
I agree with you, the nomenclature should be standardized. On the other hand, not all tools have this capability for hierarchical organization of the commands, so the only difference is in the name of the directory where they are located. The dev agent is `dev` everywhere. Let's raise this issue for v5. Should our future namespace be "BMad" or "bmad".

dracic:
> 
> BTW - in the gemini-cli the readPathFromWorkspace function does run every time (checked that) - I still don't think its an issue. i think keeping the @{} is ok
> 

Yes, now I tried It too. I'm eating humble pie. Ok, Iet's try the difference in speed.

dracic:
Man, we are talking here about few ms. I just like it so simple. @yoav0gal do we need `gemini-extension.json` for custom commands?

yoav0gal:
About the extension.json I don't think so,...
I'll drop it when I am on my computer (after I'll look into that).

About the ms, yes! That's what's I was saying, I don't think it's a big deal, 

Initially I just wanted to keep  the standards and explain my thoughts on why I didn't go for the @{} at the first place. 

If you guys think it's better this way or that way I don't really have an opinion about that.

The main thing is to get the commands working.

dracic (approved):
It works and I like this!

dracic:
@yoav0gal I was thinking. Now that you're at full speed, you could do something similar for Qwen code in antoher PR. They have now synced to 0.2.1. It seems to me that they are always one week behind compared to the Gemini CLI. We can leave it unmerged until they upgrade.

bmadcode:
Thanks @yoav0gal @dracic  - merging.

PR #551, opened 2025-09-05
qwen cli - custom commands

yoav0gal:
#549  as @dracic  asked.

NOTE:
I find Qwen code worse at "role-playing the agents", and I could not get him to act consistently with the agent description.
When just linking the file - @ + path to file, Qwen did not perform as well as just pasting the file in the context for the custom command, so that's what I used.

it's working, I think its worse than Gemini.

### DIFF
--- a/tools/installer/config/install.config.yaml
+++ b/tools/installer/config/install.config.yaml
@@ -112,15 +112,15 @@ ide-configurations:
 
   qwen-code:
     name: Qwen Code
-    rule-dir: .qwen/bmad-method/
-    format: single-file
-    command-suffix: .md
+    rule-dir: .qwen/commands/BMad/
+    format: multi-file
+    command-suffix: .toml
     instructions: |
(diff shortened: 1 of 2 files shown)


dracic:
Yes @yoav0gal I tried it with the same results, also bad with tool calling, but let this wait here and let's test, perhaps something changes with v0.3.

yoav0gal:
I think the problem is at the model level (qwen), not the cli.

bmadcode:
@yoav0gal this is still marked draft and also pr is out of date with latest.

dracic:
@bmadcode Queen CLI is still at 0.2 of upstream, so this won't work at the moment. Waiting for 0.3

dracic:
https://github.com/QwenLM/qwen-code/releases/tag/v0.0.12

@bmadcode, Qwen code is now at 0.3.4 of upstream, so this will now work!

dracic (approved):

PR #562, opened 2025-09-08
fix: update workflow file extensions from .md to .yaml in bmad-master.md

rodpl:
Fixed workflow filenames.


bmadcode:
thank you @rodpl

PR #564, opened 2025-09-09
fix: Changed title to coding standards section of brownfield architecture template

rodpl:
Brownfield architecture template contains title which after flattening will create file 'architecture/coding-standards-and-conventions.md'

Dev agent cannot find this file then. By default this agent is looking for 'architecture/coding-standards.md'


rodpl:
The same is for "Source Tree" section and "Tech Stack"

bmadcode:
thanks @rodpl !

PR #630, opened 2025-09-16
fix: let llms search in correct folder

presidenten:
# Reason behind pr

The agents has a lot of instructions pointing to folder `bmad-core`, but its installed in projects in `.bmad-core`. 
So I see them failing to find instructions and then digging to find. 

# What changed

To save on tokens I updated all mentions of `bmad-core` to `.bmad-core` except in tools, which is used to copy files from the actual `bmad-core` directory.

# Testing 

Installed with and without and diffed to doublecheck no missing files etc.


bmadcode:
resolved in v6 merging soon

PR #658, opened 2025-09-30
Fullstack Javascript Expansion Pack Added

shariqbaig:


bmadcode:
Not sure what this expansion is useful for with v4 as TS/JS is already common and works well with v4. But with v6 there will be a model for providing reference architectures. I recommend to try to submit to that once the new standard is published.

Issue #673, opened 2025-10-04
Add prompts for codex


alexeydubinin:
Codex [now](https://github.com/openai/codex/releases/tag/rust-v0.44.0) supports named & positional arguments in prompts (https://github.com/openai/codex/pull/4474, https://github.com/openai/codex/pull/4470) - please add support for this.

bmadcode:
please submit issue with examples or what specifically you are proposing - add prompts for codex where?

alexeydubinin:
I suggest use prompts like in claude code (not AGENTS.md). Not sure if examples required - this is same logic as in claude code.

alexeydubinin:
@bmadcode please, see release notes for codex (link in issue description)

alexeyv:
Codex now has a version of /commands they call "prompts". Goes into ~/.codex (like we haven't learned not to do that over two decades ago) and at the first blush doesn't say anything about namespacing.

https://github.com/openai/codex/blob/main/docs/prompts.md

alexeyv:
yep, no namespacing. would have to be done with file name prefixes like so:

<img width="1150" height="448" alt="Image" src="https://github.com/user-attachments/assets/a757abd1-8233-4834-880b-ee9cc95e7a19" />

alexeyv:
As for ~/.codex/ problem, the installer would have to look for CODEX_HOME, and upon not finding it, it would say something like:

 I can install into~/.codex/, but it will contain project-specific paths, so unless you intend to use codex for one and only one project, you should mkdir {project root}/.codex, and then always run codex as CODEX_HOME={project root}/.codex codex

PR #678, opened 2025-10-05
Feature/codex installer redesign

alexeyv:
## Summary
- replace the Codex installer’s AGENTS.md generation with flattened prompt export to `~/.codex/prompts`
- share Claude/Codex artifact helpers so both installers reuse the same collection logic without duplicating code
- stop exporting Claude-only subagents to Codex and add cleanup/detection that targets only `bmad-*` prompts
- this one installs into ~/.codex/prompts/ only, I'll make it look for CODEX_HOME and suggest project-specific install in a subsequent PR

See #673.

## Testing
- `npm run lint`
- `npm run bmad:install` , select Codex → Codex summary logs show agents/tasks/workflows, BUT NO SUBAGENTS (not implemented in Codex)
- verified `~/.codex/prompts` contains all the `bmad-<module>-*.md` files it's supposed to have

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Codex Installer
+
+- Codex installer uses custom prompts in `.codex/prompts/`, instead of `AGENTS.md`
+
 ## [6.0.0-alpha.0]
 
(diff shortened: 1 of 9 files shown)


alexeyv:
oops, wrong target branch

PR #831, opened 2025-10-28
bug/v6-revert-opencode-agent-relocation

cpitt:
reverted changes that moved the agent files install location to the command directory in the opencode installer


bmadcode:
fixed in other merge - thanks @cpitt !

PR #845, opened 2025-10-31
doc: fix key takeaways bullet points

ergut:
Since bullet points "-" was not used all the items were displayed in a single paragraph.


bmadcode:
all docs have been updated, thank you for the submission though @ergut

PR #893, opened 2025-11-11
 use INIT_CWD as default dir cause it's current working dir

yangjuncode:
as for now, npm run install:bmad --prefix git-repo-dir doesn't take current working dir as default input(it always take the git repo dir as default dir), the patch add use INIT_CWD, so it will fill with current working dir, you don't need copy path now


bmadcode:
i think this is already resolved - thank you though @yangjuncode ! Please lmk if with latest you are still seeing any issues

PR #909, opened 2025-11-12
♻️ refactor(installer): Streamline Claude Code subagent configuration

thecontstruct:
## 🚀 Summary
This PR refactors the Claude Code installer to streamline the subagent configuration process, ensuring user prompts occur only once and before any cleanup operations.

## 🧠 Context & Problem
Previously, when installing Claude Code with multiple modules that included subagents, the user would be prompted for subagent installation choices and location for *each* module. This led to a repetitive and less-than-ideal user experience. Additionally, the cleanup of old installations occurred *before* configuration collection, which could potentially lead to data loss or an inconsistent state if the user decided to abort or if configuration relied on existing files.

## ✅ Changes Proposed
- **Pre-cleanup Configuration Collection:** The `collectConfiguration` step is now executed *before* the `cleanup` operation in `ClaudeCodeSetup.setup()`. This ensures that any interactive configuration prompts are answered and stored before old installation files are removed.
- **Centralized Subagent Prompting:** The logic for prompting subagent installation choices and location has been moved out of the per-module loop in `ClaudeCodeSetup.processModuleInjections()`. It now occurs *once* at the beginning of the `processModuleInjections` method, if the installation is interactive and subagents are detected in any module.
- **`preCollectedConfig` Usage:** The `preCollectedConfig` variable is now consistently used throughout the `setup` method to pass the collected configuration to subsequent steps, ensuring that configuration is only collected once.

## 💡 Benefits & Justification
- **Improved User Experience:** Users will now only be prompted once for subagent installation preferences and location, regardless of how many modules include subagents, making the installation process smoother and faster.
- **Enhanced Stability:** By collecting configuration before cleanup, the installer avoids potential issues where user input might be lost or cleanup might interfere with configuration decisions.
- **Increased Efficiency:** Reduces redundant operations by prompting for common configuration items only once.
- **Maintainability:** The code is now clearer regarding the order of operations (collect config -> cleanup -> process injections).

## ❓ Questions for Reviewers
- Please verify the logic for `preCollectedConfig` usage, especially in scenarios where configuration might be partially pre-filled or entirely new.
- Confirm that the subagent prompting logic correctly identifies when to prompt (interactive mode, subagents present) and when to skip (non-interactive, no subagents, or already configured).
- Consider edge cases where `collectConfiguration` might be called multiple times if `_alreadyConfigured` is not set correctly.


bmadcode:
I think this is not fixed in latest. Closing - but if still an issue please reopen @thecontstruct - thank you for submission!

PR #896, opened 2025-11-11
Fix party mode workflow path

rodpl:
## What

Corrected the workflow path for the `*party-mode` command to restore intended functionality. Rebuilded agents to apply the fix.

## Why

`*party-mode` command was not working properly. It does not invoke all agent conversation.

## How

Fixed wrong path.

## Testing

Importing web bundle file in Gemini and try to invoke party-mode

PR #917, opened 2025-11-14
Fix for party-mode for web bundles

rodpl:
Contituation of this PR https://github.com/bmad-code-org/BMAD-METHOD/pull/896


bmadcode:
thank you @rodpl !!

PR #925, opened 2025-11-16
feat: add .gitignore to exclude all files except itself for web-bundles

rodpl:


bmadcode:
already resolved by removing the folder from source. thanks @rodpl

PR #936, opened 2025-11-18
Update installation command for v6 Alpha in README

ebarronh:


bmadcode:
thanks for the support - this is already resolved, so closing this @ebarronh

PR #956, opened 2025-11-21
fix: correct sprint artifacts directory path in tech-spec workflow

alexeyv:
## Summary
This PR fixes a hardcoded path in the tech-spec workflow that was creating docs/sprint_artifacts directory instead of putting the artifacts where prescribed by the .bmad/bmm/config.yaml

## Changes
- Updated [src/modules/bmm/workflows/2-plan-workflows/tech-spec/workflow.yaml](cci:7://file:///Users/alex/src/bmad/src/modules/bmm/workflows/2-plan-workflows/tech-spec/workflow.yaml:0:0-0:0) to use `{config_source}:sprint_artifacts` instead of hardcoded `{output_folder}/sprint_artifacts`

## Testing
Ran the tech-spec workflow, verified that it now correctly references the configuration value instead of using a hardcoded path.


bmadcode:
i think this is no longer an issues @alexeyv - thank you though. please lmk if I closed in error

PR #971, opened 2025-11-24
Fix workflow launcher core workflow path

dickymoore:
I received: 

> “I tried to start *develop-story, but .bmad/bmm/core/tasks/workflow.xml isn’t in the repo, so I can’t load the required core workflow.”

So a potential fix: 

- ensure IDE workflow command template uses a dedicated placeholder for the core OS path so commands don’t look under module directories
 - teach WorkflowCommandGenerator to inject ${bmad_folder}/core/tasks/workflow.xml into that placeholder while leaving workflow YAML paths untouched

Testing:

npm test
npm run format:fix    # Auto-format code
npm run bundle        # Build web bundles


bmadcode:
@dickymoore I think this is now resolved with yesterdays merge to main and will be published in the next day or two with .13

dickymoore:
Checked main at a4c394fc: core workflow path is already injected and the IDE template uses @{{core_workflow_path}}, so this PR is
  redundant. Closing as superseded.

PR #975, opened 2025-11-25
feat: Adding support for Atlassian Rovo Dev

mreis-atlassian:
- Adding support for rovo dev
- Adding rovo dev translation wrappers


bmadcode:
thank you @mreis-atlassian

PR #984, opened 2025-11-27
feat: recommend different LLM for code review in dev-story

alexeyv:
## Summary
  - Adds a tip at the end of the dev-story workflow recommending users run code-review using a
  different LLM than the one that implemented the story

  ## Changes
  - Added output line in `dev-story/instructions.xml` step 10 (completion communication)

  ## Rationale
  Using a different LLM for code review provides a fresh perspective and catches issues that the
  implementing LLM may have blind spots for—similar to having a different developer review your
  code.


alexeyv:
This is how the end of develop story looks like with this tip. 

image

PR #1006, opened 2025-12-01
update 2024 to 2025

pomazanbohdan:
Research is ongoing, with 2024 specified as the target date.


alexeyv:
It's a great catch, but hardcoding a different year (that is about to finish in less than a month) is not the greatest of solutions. Perhaps we should let LLM decide what to put in there for "give me the recent hits".

pomazanbohdan:
In general, yes, but you need to use it now, so you can treat the pull request as a hotfix.

dickymoore:
Have been thinking about this, and wonder if the best solution now is just to remove any reference to the year, as per this: https://github.com/bmad-code-org/BMAD-METHOD/pull/1016/files

If we want dynamic years, we could add a simple install‑time placeholder into these files, e.g. {{current_year}}, and have the installer replace it with new Date().getFullYear() as it  copies content. That avoids hard‑coding and keeps prompts stable. But it does mean that if you use that installation over a year rollover, you might get outdated research.

If we wanted that: 

  - In the installer copy step (e.g., tools/cli/installers/lib/core/installer.js or the file-copy utility used during module install), we could add a small text replacement map: { '{{current_year}}':String(new Date().getFullYear()) }, applied to Markdown and workflow files.
  - Then write the WebSearch lines as latest {{current_year}} 
  
  However, just removing the years altogether might be less brittle.

bmadcode:
good catch here @pomazanbohdan - actually I did not intend to keep hardcoded dates in like that, this must have slipped through.

bmadcode:
what I was seeing with claude is without putting anything, it was putting its own year (2024) into every search - so temp I added that, meaning to come up with something better. I think best will be to tell it to websearch without using specific dates and instead just word the search as .... past 6 months or similar. open to suggestions here, but will def be getting rid of the hard coded years soon.

alexeyv:
Tested removing years entirely vs keeping them. Results: **identical quality** - search engines already prioritize current documentation.

**Tested patterns:**
- `CLI command options flags latest` (Next.js, create-t3-app)
- `latest stable version` (Prisma)
- `boilerplate generator latest options` (React)
- `production-ready starter best practices` (Next.js)
- `recent updates maintenance status` (create-t3-app)
- `production readiness` (PostgreSQL)

All returned the same authoritative sources (official docs, GitHub, npm) with or without year.

**Recommendation:** Remove years entirely rather than updating to 2025. The word "latest" is sufficient.

**Note:** `file-utils.md` change is unnecessary - that's test fixture data ("Financial Report Q4"), not a search query.

Will push a commit that removes years from WebSearch queries and leaves file-utils.md unchanged.

alexeyv:
I made a PR #1040 as an alternative solution: it removes the years entirely rather than updating to 2025.

PR #1013, opened 2025-12-02
docs: create CODE_OF_CONDUCT.md

alexeyv:


dickymoore:
This is excellent! Great to see this.

A couple of changes may be useful: 

1. No enforcement contact address: 
> `Instances of abusive, harassing, or otherwise unacceptable behavior may be
reported to the community leaders responsible for enforcement at
.`
2. Does Community Leaders mean maintainers? It might be worth making that explicit.

Dicky

alexeyv:
@dickymoore, @bmadcode 

Instances of abusive, harassing, or otherwise unacceptable behavior may be reported to the community leaders responsible for enforcement at the officlal BMAD Discord server [https://discord.com/invite/gk8jAdXWmj] - DM a moderator or flag a post.

OR is there something else that for any reason is better than the above? You don't want to put a personal email there for a whole bunch of good reasons, half of them obvious.

dickymoore:
@alexeyv I can't see a reason why Discord isn't a good option here, so looks good to me. Is there guidance for users when they arrive at Discord on how to raise a report? I think it's quite obvious for now but it might not be in the future. Just a thought. Either way, this is great.

bmadcode:
This looks great @alexeyv , and thanks for the feedback @dickymoore ! I will add something like this also to the discord onboarding welcome process to ack the code of conduct.

PR #1051, opened 2025-12-07
fix(core): clarify elicitation option labels - 'Random 5' not 'Reshuffle'

alexeyv:
## Summary

- Changes misleading "Reshuffle" label to "Random 5" in advanced elicitation options
- The first display shows context-matched methods (not shuffled), so "Reshuffle" was inaccurate
- Updates both the prompt line and the option description

## Changes

- Line 48: `[r] to Reshuffle` → `[r] Random 5`
- Line 55: `r. Reshuffle the list with 5 new options` → `r. Pick 5 random methods`

## Test plan

- [x] Run advanced elicitation task and verify new labels display correctly
- [x] Confirm `[r]` option still functions as expected (picks 5 random methods)


bmadcode:
@alexeyv the new description sounds like it will pick and run with 5 random choices instead of listing 5 different options to choose from.

PR #1057, opened 2025-12-07
fix: Link to Gemini Gem manager for creating Gem with web bundle

wsmoak:
Updated the link for creating a Gem.  

There is no option to create a Gem at https://aistudio.google.com/ .  This page has a "New app" option.

Instead, this option is in the Gemini Gem manager at https://gemini.google.com/gems/view .


bmadcode:
Good catch in the docs! Thanks @wsmoak

PR #1059, opened 2025-12-07
fix: resolve core module installation path lookup

sjennings:
## Summary
- Fixes the "core" module installation bug where the installer failed to locate the core module
- The core module lives in `src/core`, not `src/modules` like other modules
- Added special handling in `findModuleSource()` to correctly locate the core module using `getSourcePath('core')` instead of falling through to the default `src/modules` search path

## How This Fixes the Bug
When users tried to install the "core" module, the `findModuleSource()` function was searching in `src/modules/core` which doesn't exist. The core module is actually located at `src/core`. This fix adds a special case check at the beginning of the function to detect when the requested module is "core" and use the correct path via `getSourcePath('core')`.

## Test Plan
- [x] Verified `npm test` passes with all checks
- [x] Manual test: run `npx bmad-method install` and select "core" module


alexeyv:
## PR Review: #1059

**Title:** fix: resolve core module installation path lookup
**Author:** @sjennings
**Branch:** fix/core-module-install-path → main

---

### Findings

### 1. Incomplete Manual Testing [likely]

**Severity:** 🟡 Moderate

The test plan in the PR description shows the manual test as incomplete:
```
- [ ] Manual test: run `npx bmad-method install` and select "core" module
```

For a bug fix targeting this specific installation path, the manual verification step should be completed before merge. Automated tests may pass while the actual user-facing bug remains.

**Suggested Fix:** Complete the manual test and check the box, or document what was actually tested.

---

### 2. Redundant Special-Case Check Remains

**Severity:** 🟢 Minor

After this fix, the special-case check at `manager.js:1085-1088`:
```javascript
if (moduleName === 'core') {
  sourcePath = getSourcePath('core');
} else {
  sourcePath = await this.findModuleSource(moduleName);
}
```

...is now redundant since `findModuleSource()` handles the core module properly. This creates two locations doing the same check.

**Suggested Fix:** Consider simplifying to just:
```javascript
sourcePath = await this.findModuleSource(moduleName);
```

---

### 3. API Inconsistency

**Severity:** 🟢 Minor

The fix uses `getSourcePath('core')` while `project-root.js` provides `getModulePath()` specifically designed to abstract the core vs modules path distinction (lines 59-64):

```javascript
function getModulePath(moduleName, ...segments) {
  if (moduleName === 'core') {
    return getSourcePath('core', ...segments);
  }
  return getSourcePath('modules', moduleName, ...segments);
}
```

Using `getModulePath('core')` would be more semantically appropriate and consistent with the API design.

**Suggested Fix:** Replace `getSourcePath('core')` with `getModulePath('core')`.

---

### Summary

**Critical:** 0 | **Moderate:** 1 | **Minor:** 2

The fix is correct and addresses the root cause - `findModuleSource()` was searching in `src/modules/core` when it should look in `src/core`. The implementation works, though there are opportunities for cleanup and consistency improvements.

---

_Review generated by Raven's Verdict. LLM-produced analysis - findings may be incorrect or lack context. Verify before acting._

sjennings:
Box checked to please robot: not only was this tested, bmad was unable to be installed to my projects without this fix.

alexeyv:
@sjennings Thanks! What's your opinion about the robot? Both on substance and style.

sjennings:
> @sjennings Thanks! What's your opinion about the robot? Both on substance and style.

annoyingly thorough and persnickety, so it's doing a good job. style is fine, doesn't seem particularly LLMish

bmadcode:
resolved in .15 - thank you though @sjennings

PR #1060, opened 2025-12-07
fix: retrieve installed modules during manifest processing for compil…

diegosz:
…ing agents

The problem was that the installedModules variable was being used at line 1987 but was never defined in the compileAgents method's scope.

Issue: When compiling agents the compilation failed.
Installation failed: installedModules is not defined
ReferenceError: installedModules is not defined at Installer.compileAgents (BMAD-METHOD/tools/cli/installers/lib/core/installer.js:1987:30) at async Command.action (BMAD-METHOD/tools/cli/commands/install.js:26:24)

The fix:

- Added the declaration let installedModules = []; at line 1968
- Added code to populate installedModules by reading the bmad directory and filtering for module directories (lines 1975-1979)
- This follows the same pattern used in other methods like  generateManifests (line 1237) and update (line 2050)


bmadcode:
resolved in .15, thanks @diegosz !

diegosz:
awesome, thanks

PR #1061, opened 2025-12-08
feat: add external agent support for code reviews

sjennings:
## Summary

Adds support for delegating adversarial code reviews to external CLI agents (Codex, Gemini, or Claude) when available. This provides independent, unbiased code reviews from a different AI model.

### Key Features

- **External Agent Detection**: Auto-detects available CLI tools (codex, gemini, claude) and caches results in config.yaml
- **Mandatory Delegation**: When an external agent is available, the workflow MUST delegate to it (with explicit prohibited rationalizations to prevent skipping)
- **Fallback Support**: Falls back to built-in review only if all external agents fail detection or invocation
- **Config Caching**: Detection results are cached to avoid repeated CLI checks on subsequent runs

### Changes

- **workflow.xml**: Added `invoke-bash` and `set-var` execution tags to the workflow engine
- **install-config.yaml**: Added `external_review_agents` configuration section with preferred agent selection
- **code-review/instructions.xml**: Complete rewrite to support external agent detection, invocation, and fallback
- **code-review/workflow.yaml**: Simplified to use runtime variables for agent detection
- **code-review/checklist.md**: Updated to reflect new external agent workflow

### External Agent Invocation Commands

| Agent | Command |
|-------|---------|
| Codex | `codex exec --full-auto "prompt"` |
| Gemini | `gemini -p "prompt" --yolo` |
| Claude | `claude -p "prompt" --dangerously-skip-permissions` |

## Test plan

- [ ] Install BMAD to a project with codex CLI available
- [ ] Run code-review workflow and verify it detects and invokes codex
- [ ] Verify config.yaml is updated with detection results
- [ ] Test with gemini/claude CLI if available
- [ ] Test fallback to built-in review when no external agents available

### DIFF
--- a/src/modules/bmm/workflows/4-implementation/code-review/instructions.xml
+++ b/src/modules/bmm/workflows/4-implementation/code-review/instructions.xml
@@ -288,38 +288,135 @@
     </check>
   </step>
 
-  <step n="4" goal="Present findings and fix them">
-    <action>Categorize findings: HIGH (must fix), MEDIUM (should fix), LOW (nice to fix)</action>
+  <!-- ================================================================ -->
+  <!-- STEP 4: TRIAGE FINDINGS - SEPARATE SIGNAL FROM NOISE            -->
+  <!-- ================================================================ -->
+  <!-- This step evaluates external agent findings with project context.
+       External agents find ALL possible issues - but not all matter.
+       A symlink bypass is critical for a web API, irrelevant for a game.
+       The orchestrating agent applies pragmatic judgment here. -->
+
+  <step n="4" goal="Triage findings - separate signal from noise">
+    <critical>External agents are adversarial by design - they find EVERYTHING.</critical>
+    <critical>Your job: Apply project context to determine what ACTUALLY matters.</critical>
+    <critical>Do NOT blindly accept all findings. Do NOT dismiss valid concerns.</critical>
+
+    <action>For EACH finding from {{all_findings}}, evaluate against these criteria:</action>
+
+    <!-- Evaluation Framework -->
+    <action>**ACTUALLY IMPORTANT** (always fix):
+      - AC violations: Story claims something works but it doesn't
+      - Task fraud: Checkbox marked [x] but code doesn't exist
+      - Contract violations: Method signature doesn't match documented behavior
+      - Real bugs: Code will fail at runtime in normal usage
+      - Security issues for the ACTUAL threat model (game loading own files ≠ public API)
+    </action>
+
+    <action>**CONTEXT-DEPENDENT** (ask user):
+      - Performance: Does it matter at this scale? Game data vs. million-record DB
+      - Validation strictness: Nice-to-have vs. actually needed
+      - Edge cases: Will they ever happen in practice?
+      - Thread safety: Is this actually multi-threaded?
+    </action>
+
+    <action>**THEORETICAL/NITPICKING** (skip unless user insists):
+      - "Could be exploited if attacker controls X" when attacker never controls X
+      - Micro-optimizations that save nanoseconds
+      - Style preferences disguised as bugs
+      - "Best practice" violations that don't cause problems
+      - DoS concerns for trusted internal data
+    </action>
+
+    <!-- Build categorized lists -->
+    <set-var name="important_findings" value="[]" />
+    <set-var name="contextual_findings" value="[]" />
+    <set-var name="theoretical_findings" value="[]" />
+
+    <action>Categorize each finding into one of the three lists with brief reasoning</action>
+
+    <!-- Present triage to user via AskUserQuestion -->
+    <output>**🔍 FINDINGS TRIAGE**
+
+I've reviewed {{external_agent_cmd}}'s findings against your project context.
+Here's what I think actually matters vs. theoretical concerns:
+    </output>
+
+    <!-- Important findings - recommend fixing -->
+    <check if="{{important_findings}} is not empty">
+      <output>
+## ✅ ACTUALLY IMPORTANT (Recommend Fix)
+These are real issues that affect correctness or violate documented contracts:
+
+{{#each important_findings}}
+- **{{this.id}}**: {{this.summary}}
+  - *Why it matters*: {{this.reasoning}}
+{{/each}}
+      </output>
+    </check>
+
+    <!-- Context-dependent - ask user -->
+    <check if="{{contextual_findings}} is not empty">
+      <ask questions="[
+        {
+          'question': 'Which context-dependent issues should we address?',
+          'header': 'Fix these?',
+          'multiSelect': true,
+          'options': [
+            {{#each contextual_findings}}
+            {
+              'label': '{{this.id}}: {{this.short_summary}}',
+              'description': '{{this.reasoning}}'
+            },
+            {{/each}}
+            {
+              'label': 'None of these',
+              'description': 'Skip all context-dependent issues'
+            }
+          ]
+        }
+      ]" />
+      <action>Add user-selected contextual findings to {{important_findings}}</action>
+    </check>
+
+    <!-- Theoretical findings - inform user but don't push -->
+    <check if="{{theoretical_findings}} is not empty">
+      <output>
+## ⏭️ SKIPPING (Theoretical/Nitpicking)
+These findings are technically valid but don't matter for your use case:
+
+{{#each theoretical_findings}}
+- **{{this.id}}**: {{this.summary}} — *{{this.reasoning}}*
+{{/each}}
+
+*(Say "fix all" if you want these addressed anyway)*
+      </output>
+    </check>
+
+    <!-- Final confirmation -->
+    <set-var name="final_fix_list" value="{{important_findings}}" />
+    <output>
+**📋 FINAL FIX LIST: {{final_fix_list.length}} issues**
+    </output>
+  </step>
+
+  <step n="5" goal="Present findings and fix them">
     <action>Set {{fixed_count}} = 0</action>
     <action>Set {{action_count}} = 0</action>
 
-    <output>**🔥 CODE REVIEW FINDINGS, {user_name}!**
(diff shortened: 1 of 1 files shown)


bmadcode:
Great idea - a few things!

I am about to switch all of the workflows to not use workflow.xml to the sharded steps. So for any command we want to give to an external agent, it sounds like we will just want a clear separate text file prompt that is self contained also in a single file to give to the agent? This could be used for much more than just the code review.

When installing, the user already has a multi-select for Tools they have installed. This could be just a single question then in bmm if they want to allow supported installed tools do external command executions. Also saying it this way, as this could really be used for much more than code review, pretty much review of anything could potentially produce better results.

What combinations of tools and OS was this tested on?

Random idea - what if we had a folder of named prompts to pass to other agents, maybe call them external workflows. The could all register as installed tool external slash commands. so if the folder had external-code-review.md, for each tool installed it could create a slash command. /bmad-bmm-ext-copilot-code-review. While this could result in a lot for users to remember, all they would need to know is the tool and autocomplete will show them the options. so if in claude, and started to type /copilot - the autocomplete would show all of the possible commands.

alexeyv:
The choice of alt agent should belong with the user.

I love this capability in principle. But I also wonder, how many people would frown upon the idea of BMAD deciding to run some agentic software in YOLO mode on the computer without asking?

What if we defined a location like `.bmad/alt-agent.[md|sh]`? If it exists, run review through it. If it doesn't, stop and ask. We would then take all this code from this PR, put it into examples of alt agent. And we could also have a "there is no alt agent" version of alt-agent.md For when you don't want to be asked, and you don't want to be delegating it to some other LLM.

sjennings:
_it sounds like we will just want a clear separate text file prompt that is self contained also in a single file to give to the agent?_ 

It's already a plaintext prompt, I'll break it out into a text file.

_When installing, the user already has a multi-select for Tools they have installed. This could be just a single question then in bmm if they want to allow supported installed tools do external command executions._

I'll see if I can get it integrated to the installer as a choice. I need to update that anyway to work with v15.

_The choice of alt agent should belong with the user._

Absolutely. It's defined in a config file right now. Could use a selector in the installer, perhaps.

_What if we defined a location like .bmad/alt-agent.[md|sh]?_

Right now it's defined in config.yaml, this could be moved to its own file fairly easily.

_how many people would frown upon the idea of BMAD deciding to run some agentic software in YOLO mode on the computer without asking?_

This is an excellent point and one I agree with. By its nature a code review *should* be non-destructive - and I'll add to the prompt emphasis that the external agent should not write/change/delete *any* file. However to run as an external agent in a headless context the YOLO mode is necessary. Code reviews *should* be the safest application of this, but it also assumes the user is running in a fairly hardened/backed up environment and the LLM is not going to go off the rails and delete everything it sees. Neither of these are safe assumptions!  https://bsky.app/profile/brokentoys.social/post/3m7h4ul6dsc2u

_this could really be used for much more than code review, pretty much review of anything could potentially produce better results._

The problem here is that external agents have to run in yolo mode with this approach - for anything that has a write application, it would require a dozen warnings and they would not be enough and someone WILL come screaming into the support discord that we destroyed their computer.

_What combinations of tools and OS was this tested on?_

Just mac os, claude for the primary agent, and codex, gemini and claude for external agents. This is primarily crafted to deal with a frustration I had with invoking repeated code reviews in codex struggling with codex's often-broken slash command support. 

I would not be opposed to having this functionality be in an optional module. I don't know if a core workflow change can be done via a modular approach currently.

I'll update shortly with discussed changes and updates to maintain functionality with current version.

sjennings:
Changes in most recent commit:

- Now fully compatible with alpha 15
- Installer support. User can select from Codex, Gemini, Claude Code or None. None is listed as default.
- Prompt used is now in a markdown file in the workflow directory.
- Added language to prompt to try to ensure read-only operations.
- The Codex external agent invocation now uses a read-only sandbox. (Need to research if this is as easily supported in Gemini and Claude Code and test both.)

sjennings:
Also a note: I've started using a Codex skill for claude code - Claude will use that automatically instead of the specified invocation (which is fine, if a user installs a skill for this they already have opinonated preferences). Wondering if it would be helpful to embed? (Probably not since it's Claude Code specific)

```

---
name: codex
description: Use when the user asks to run Codex CLI (codex exec, codex resume) or references OpenAI Codex for code analysis, refactoring, or automated editing
---

# Codex Skill Guide

## Running a Task

1. Ask the user (via `AskUserQuestion`) which model to run (`gpt-5.1-codex-max`, `gpt-5.1-codex-mini` or `gpt-5.1`)) AND which reasoning effort to use (`high`, `medium`, or `low`) in a **single prompt with two questions**.
2. Select the sandbox mode required for the task; default to `--sandbox read-only` unless edits or network access are necessary.
3. Assemble the command with the appropriate options:
   - `-m, --model <MODEL>`
   - `--config model_reasoning_effort="<high|medium|low>"`
   - `--sandbox <read-only|workspace-write|danger-full-access>`
   - `--full-auto`
   - `-C, --cd <DIR>`
   - `--skip-git-repo-check`
4. Always use --skip-git-repo-check.
5. When continuing a previous session, use `codex exec --skip-git-repo-check resume --last` via stdin. When resuming don't use any configuration flags unless explicitly requested by the user e.g. if he species the model or the reasoning effort when requesting to resume a session. resume syntax: `echo "your prompt here" | codex exec --skip-git-repo-check resume --last 2>/dev/null`. All flags have to be inserted between exec and resume.
6. **IMPORTANT**: By default, append `2>/dev/null` to all `codex exec` commands to suppress thinking tokens (stderr). Only show stderr if the user explicitly requests to see thinking tokens or if debugging is needed.
7. Run the command, capture stdout/stderr (filtered as appropriate), and summarize the outcome for the user.
8. **After Codex completes**, inform the user: "You can resume this Codex session at any time by saying 'codex resume' or asking me to continue with additional analysis or changes."

### Quick Reference

| Use case | Sandbox mode | Key flags |
| --- | --- | --- |
| Read-only review or analysis | `read-only` | `--sandbox read-only 2>/dev/null` |
| Apply local edits | `workspace-write` | `--sandbox workspace-write --full-auto 2>/dev/null` |
| Permit network or broad access | `danger-full-access` | `--sandbox danger-full-access --full-auto 2>/dev/null` |
| Resume recent session | Inherited from original | `echo "prompt" \| codex exec --skip-git-repo-check resume --last 2>/dev/null` (no flags allowed) |
| Run from another directory | Match task needs | `-C <DIR>` plus other flags `2>/dev/null` |

## Following Up

- After every `codex` command, immediately use `AskUserQuestion` to confirm next steps, collect clarifications, or decide whether to resume with `codex exec resume --last`.
- When resuming, pipe the new prompt via stdin: `echo "new prompt" | codex exec resume --last 2>/dev/null`. The resumed session automatically uses the same model, reasoning effort, and sandbox mode from the original session.
- Restate the chosen model, reasoning effort, and sandbox mode when proposing follow-up actions.

## Error Handling

- Stop and report failures whenever `codex --version` or a `codex exec` command exits non-zero; request direction before retrying.
- Before you use high-impact flags (`--full-auto`, `--sandbox danger-full-access`, `--skip-git-repo-check`) ask the user for permission using AskUserQuestion unless it was already given.
- When output includes warnings or partial results, summarize them and ask how to adjust using `AskUserQuestion`.
`
```

sjennings:
Added an attempt at a fix for a recently introduced claude code annoyance: it will now sometimes decide to launch bash tasks (like these code reviews) as background processes, and then reconnect to them for the next code review, assume that they are a new code review instead of the one previously completed, and happily report that the code review is only reporting old issues. Well of course it is, claude, you read the old review. Sigh. Anyway this should hopefully fix it (it's difficult to test because it only happens occasionally)

sjennings:
I also don't think this behavior will persist long term because it eats your context alive monitoring the results of a shell over... and over... and over... and over...

sjennings:
sigh. another fix enforcing this behaviour because claude code thinks it knows best. "Wait! Let me use this data before I kill it!"

bmadcode:
@sjennings thanks for the updates and sorry for delay in further response, I have been traveling and availability has been limited. I like the idea of this being an installer question configuration btw.

I will look at all of the PR updates soon and try this out also - might be another day or two just due to the travel and getting situated.

alexeyv:
@sjennings, @bmadcode here is a simple thought: 

What do we call a piece knowledge, user- or project-specific, about executing code reviews in ways that a user knows and approves of? Hint: 7 letters, starts with the C, ends with 'ontext'.

What if this wants to be a workflow a la 'setup-reviews', that guides the user through creation of `docs/how-to-review.yaml`? That `how-to-review` may say things like "Use Bash to call CodeRabbit", "Use the Task tool to conduct adversarial review", "Use Bash to call `codex` and pass it this prompt:..." We really don't care what it says. We just suggest to the user what it might say and ask questions. As long as the output is shaped as a list(-s) of review findings  

`code-review` then becomes an orchestrator:
- assemble a diff of changes under review
- If how-to-review exists. read and follow it. 
- If it doesn't, either 
  - abort and tell the user to run `setup-reviews`, or
  - Fall back to what `code-review` is currently doing, plus maybe the adversarial bit2qw
- Put all findings together, deduplicate them, assess their severity and what's real and what's not, ordered by severity, and present to the user. Back to what it already knows how to do: dealing with the findings.

alexeyv:
@sjennings I love what you've done with the place. Don't mind me stealing half of this diff to use for work? :)

sjennings:
Feel free! It was your suggestion to dive in and then submit a PR that prompted my sharing it.

As for the changes in the Copilot review I'll look into them more closely later.

alexeyv:
Here is my version of a small step in the same general direction:
https://github.com/bmad-code-org/BMAD-METHOD/pull/1114

alexeyv:
@sjennings 
If we end up going with `how-to-review.yaml`, better part of what you have here becomes the things `setup-review` suggests. 
/creative raving mode

sjennings:
Closing for now since @alexeyv is working on an alternate implementation

PR #1065, opened 2025-12-08
fix: collect module.yaml prompts for custom modules

OhSeungWan:
## What

Custom modules with `module.yaml` configuration prompts were not being collected during installation.

## Why

When installing custom modules that define their own configuration prompts in `module.yaml`, those prompts were ignored because the config collector didn't know the custom module paths. This resulted in custom module configs not receiving user-provided values from their module.yaml prompts.

## How

- Added `customModulePaths` option to `ConfigCollector.collectAllConfigurations()`
- Built custom module path map from `selectedFiles` (existing installs) and `cachedModules` (new installs) in installer
- Merged collected config into custom module installation process

## Testing

Tested with custom module containing `module.yaml` prompts - prompts now appear during installation and values are correctly applied to the module config.

### DIFF
--- a/tools/cli/installers/lib/core/config-collector.js
+++ b/tools/cli/installers/lib/core/config-collector.js
@@ -132,8 +132,12 @@ class ConfigCollector {
    * Collect configuration for all modules
    * @param {Array} modules - List of modules to configure (including 'core')
    * @param {string} projectDir - Target project directory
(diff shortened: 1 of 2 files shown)


bmadcode:
good catch here, thank you @OhSeungWan

Issue #872, opened 2025-11-06
BMAD V6 asks to run /bmad:bmm:workflows:create-architecture which doesn't exist


raegislabs:
**Describe the bug**
BMAD asks to run /bmad:bmm:workflows:create-architecture which doesn't exist

**Steps to Reproduce**
Step through workflow-init and after prd and it asks you to do architecture workflow

**PR**
NA

**Expected behavior**
Command /bmad:bmm:workflows:create-architecture should exist

**Please be Specific if relevant**
Model(s) Used: Sonnet 4.5, GPT-5 High
Agentic IDE Used: Claude Code, Codex CLI
WebSite Used: NA
Project Language: English
BMad Method version: V6

**Screenshots or Links**
  What's Next

  Immediate Actions:

...
  2. Architecture Workflow - Run: /bmad:bmm:workflows:create-architecture
    - Agent interface contracts
    - Context handoff data schemas
    - BMAD orchestration patterns
    - Quality gate validation logic

**Additional context**
Run into it multiple times so it's not a one off issue. Not a blocker because we can just load architect agent and do the same thing.

dickymoore:
I pushed branch fix/872-architecture-command to my fork and have [drafted this PR](https://github.com/bmad-code-org/BMAD-METHOD/pull/1066). Workflow command generation now anchors paths to {project-root}, so create-architecture and similar commands no longer resolve core workflow files inside the bmm module. Added a regression test and npm test passes. Ready to open the PR. Let me know if you’d like me to do that.

bmadcode:
This issue has been inactive for over 30 days and is being closed as part of backlog cleanup.

If this issue still applies to the latest beta version, please reopen it with additional details about:
- Whether you've tested with the latest version
- Current impact or severity
- Steps to reproduce if it's a bug

PR #1066, opened 2025-12-08
fix: anchor workflow commands for create-architecture

dickymoore:
- **What:** Anchor generated workflow commands to `{project-root}` so IDEs don’t resolve core/workflow paths relative to module folders; add regression test for the `create-architecture` workflow command.
- **Why:** `*workflow-init`/`*create-architecture` commands could fail with “workflow.xml does not exist” because generated command paths were relative (e.g., `.bmad/bmm/core/tasks/workflow.xml`). Anchoring ensures commands point to `.bmad/core/tasks/workflow.xml` and the correct workflow file.
- **How:**
   - Update workflow command generator to prefix core/workflow paths with `{project-root}` and anchor display paths for installed workflows.
  - Add installation component test validating anchored paths for the `create-architecture` command.
- **Testing:** `npm test`

I double-checked upstream main (55cb4681) and the anchoring fix isn’t present:

  - tools/cli/installers/lib/ide/shared/workflow-command-generator.js still emits relative paths (no {project-root} anchoring / no withProjectRoot helper), so commands can resolve core/tasks/workflow.xml inside module folders.
  - test/test-installation-components.js has no workflow command path coverage.

This looks like a bug, not intentional behavior:

  - core/tasks/workflow.xml is under .bmad/core/tasks, but the generator injects {{core_workflow_path}} as <bmadFolder>/core/tasks/workflow.xml without {project-root}. When the workflow path is already an installed path (e.g., .bmad/bmm/workflows/...), IDEs resolve @... relative to that directory, yielding .bmad/bmm/core/tasks/workflow.xml - the failure seen in #872.
  - There’s no test guarding the path shape, so the regression slipped through.
  - The command templates state “Always LOAD {project-root}/…/core/tasks/workflow.xml”, which implies anchoring to project root; no docs suggest the current relative resolution is deliberate.

Given the symptom and template intent, anchoring to {project-root} matches the design, while the current relative output is erroneous.

## Release Notes

* **Tests**
  * Added validation test for workflow command paths with project-root anchoring.

* **Bug Fixes**
  * Improved workflow path resolution by anchoring paths to project root for consistent behavior across environments.

* **Updates**
  * Enhanced workflow command template with new output handling step and updated path references.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/tools/cli/installers/lib/core/custom-module-cache.js
+++ b/tools/cli/installers/lib/core/custom-module-cache.js
@@ -51,7 +51,19 @@ class CustomModuleCache {
   }
 
   /**
-   * Calculate hash of a file or directory
+   * Stream a file into the hash to avoid loading entire file into memory
+   */
+  async hashFileStream(filePath, hash) {
+    return new Promise((resolve, reject) => {
+      const stream = require('node:fs').createReadStream(filePath);
+      stream.on('data', (chunk) => hash.update(chunk));
+      stream.on('end', resolve);
+      stream.on('error', reject);
+    });
+  }
+
+  /**
+   * Calculate hash of a file or directory using streaming to minimize memory usage
    */
   async calculateHash(sourcePath) {
     const hash = crypto.createHash('sha256');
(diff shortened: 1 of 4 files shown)


dickymoore:
@coderabbitai review

bmadcode:
@dickymoore good idea here, but going to close as will be removing all need for workflow.xml once the final workflows are converted.

Issue #1074, opened 2025-12-09
BMAD V6 Alpha 15 Install Failure: Module 'custom' not found in any source location


raegislabs:
**Describe the bug**
✖ Installation failed
Installation failed: Module 'custom' not found in any source location
Error: Module 'custom' not found in any source location
    at ModuleManager.install (/Users/xxx/.npm/_npx/2e2f2306f34e978a/node_modules/bmad-method/tools/cli/installers/lib/modules/manager.js:419:13)
    at async Installer.installModuleWithDependencies (/Users/johnpates/.npm/_npx/2e2f2306f34e978a/node_modules/bmad-method/tools/cli/installers/lib/core/installer.js:1618:5)
    at async Installer.install (/Users/xxx/.npm/_npx/2e2f2306f34e978a/node_modules/bmad-method/tools/cli/installers/lib/core/installer.js:979:15)
    at async Command.action (/Users/xxx/.npm/_npx/2e2f2306f34e978a/node_modules/bmad-method/tools/cli/commands/install.js:48:22)

**Steps to Reproduce**
Install Bmad v6 over a previous alpha version install (modify settings)

**Expected behavior**
Custom agents and workflows moved to the new location. Or at least install can complete without an error

**Additional context**
I have BMAD v6 alpha 12 installed with custom agents and workflows in the old default .bmad/custom (I think)

raegislabs:
issue still persists in alpha 16.

bmadcode:
This issue has been inactive for over 30 days and is being closed as part of backlog cleanup.

If this issue still applies to the latest beta version, please reopen it with additional details about:
- Whether you've tested with the latest version
- Current impact or severity
- Steps to reproduce if it's a bug

PR #1079, opened 2025-12-09
Look in the local cache for module installer

kaheidt:
Looks in the local _cfg custom module cache location for the module, in case the module is one that was sourced from an external location.

Without this, the `installer.js` for custom modules is not executed.

### DIFF
--- a/tools/cli/installers/lib/modules/manager.js
+++ b/tools/cli/installers/lib/modules/manager.js
@@ -1122,7 +1122,13 @@ class ModuleManager {
     if (moduleName === 'core') {
       sourcePath = getSourcePath('core');
     } else {
-      sourcePath = await this.findModuleSource(moduleName);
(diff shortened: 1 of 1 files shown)


kaheidt:
Note: I believe this also resolves #1074

bmadcode:
@kaheidt thanks for this - I temp remove the functionality and and rebuilding the installer from the ground up for better custom support also.

kaheidt:
> @kaheidt thanks for this - I temp remove the functionality and and rebuilding the installer from the ground up for better custom support also.

Sounds good. If you need a second pair of eyes and someone that has some custom modules and workflows that I'm distributing internally manually, please feel free to tag me.

bmadcode:
@kaheidt definitely! Are you in the discord - if so also please msg me there - I have some general questions of what types of features people are looking for regarding sharing of custom agents workflows and modules.

PR #1086, opened 2025-12-09
Enhance config collector to support static fields

kaheidt:
Refactor config collection to handle both interactive and static fields. Update logic to process new static fields and merge answers accordingly.

Currently the configuration-collector only collects information that's based on question/answer interaction when a module is being installed. It doesn't collect static "answer-only" data from the module.yaml and put it in the config.yaml. This causes fields like the [static items in the module template](https://github.com/bmad-code-org/BMAD-METHOD/blob/main/src/modules/bmb/workflows/create-module/templates/module.template.yaml#L47-L53) to not be represented in the final config.yaml


bmadcode:
thanks @kaheidt good catch here.

Issue #1124, opened 2025-12-13
DSL: template-output tag is overloaded for two different purposes


alexeyv:
## Description

The `<template-output>` tag is used for two semantically different purposes, creating confusion:

### Pattern 1: Template-Driven (actual template filling)
```xml
<template-output file="{default_output_file}">story_header</template-output>
```
- Has `file=` attribute
- Fills sections of an actual template file (`template.md`)
- Generates document output
- Name makes sense: "template output"

### Pattern 2: Data Return (no template involved)
```xml
<template-output>risks = {{risks}}</template-output>
<template-output>next_workflow_id = {{next_workflow_id}}</template-output>
```
- No `file=` attribute
- No template file exists
- Returns key-value pairs to calling workflow
- Name is misleading: there's no "template" being filled

## The Problem

When a workflow like `sprint-status` uses data mode to return values, it uses `<template-output>` even though:
- No `template.md` file exists
- No `template:` field in `workflow.yaml`
- No file is being generated
- It's purely returning data to a caller

This overloaded naming makes the workflow DSL harder to understand and maintain.

## Suggested Fix

Introduce a separate tag for data return pattern:

```xml
<!-- For actual template filling -->
<template-output file="{output}">section_name</template-output>

<!-- For returning data to caller -->
<return-value>risks = {{risks}}</return-value>
<!-- or -->
<output>risks = {{risks}}</output>
<!-- or -->
<export>risks = {{risks}}</export>
```

Alternatively, update documentation to clearly explain the two patterns and when to use each.

## Affected Files

- `src/core/tasks/workflow.xml` (defines the tag behavior)
- Multiple workflow instructions files using `<template-output>` for data return:
  - `sprint-status/instructions.md` (data mode)
  - `workflow-status/instructions.md` (data mode)
  - Others using service/data patterns

## Found During

Adversarial code review of sprint-status workflow.

alexeyv:
Closing - decided not to proceed with this change

PR #1142, opened 2025-12-15
fix(core): introduce <output> tag for data emission in Core and BMM

alexeyv:
## Summary

Partially addresses #1124 - The `<template-output>` tag was overloaded for two semantically different purposes.

### The Problem

`<template-output>` was used for:
1. **Template filling** (with `file=` attribute) - actual template file generation
2. **Data emission** (without `file=` attribute) - returning key-value data

This made the DSL confusing.

### The Solution

Introduce a new `<output>` tag for the data emission pattern:

| Tag | Purpose | Has `file=`? | Writes file? | Pauses for user? |
|-----|---------|--------------|--------------|------------------|
| `<template-output file="path">` | Fill template sections | Yes | Yes | Yes |
| `<output>` | Emit key=value data | No | No | No |

### Changes

- Updated `workflow.xml` DSL definition with new `<output>` tag
- Converted occurrences in **Core and BMM modules only**
- 8 files changed

### Scope Limitation

**BMGD and CIS modules intentionally excluded** - see #1143

Those modules have a CHECKPOINT PROTOCOL that references `<template-output>` tags, and many `<template-output>` tags without `file=` attributes. The interaction between these is ambiguous and needs investigation before changes.

## Test plan

- [x] All tests pass (`npm run test`)
- [x] Lint and format checks pass

### DIFF
--- a/src/core/tasks/workflow.xml
+++ b/src/core/tasks/workflow.xml
@@ -5,7 +5,7 @@
     <mandate>Always read COMPLETE files - NEVER use offset/limit when reading any workflow related files</mandate>
     <mandate>Instructions are MANDATORY - either as file path, steps or embedded list in YAML, XML or markdown</mandate>
     <mandate>Execute ALL steps in instructions IN EXACT ORDER</mandate>
-    <mandate>Save to template output file after EVERY "template-output" tag</mandate>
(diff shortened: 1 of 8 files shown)


alexeyv:
Closing - decided not to proceed with this change

PR #1178, opened 2025-12-22
Fix name to match link from the main README.md

jhattat:


bmadcode:
No longer applicable as all docs moved and renamed @jhattat - thank you

PR #1182, opened 2025-12-22
refactor(bmm): convert quick-dev workflow to sharded format with adversarial review

alexeyv:
## Summary

- Convert monolithic `instructions.md` to 6 modular step files for better maintainability
- Add `review-adversarial-general.xml` core task for adversarial code reviews
- Replace `workflow.yaml` with `workflow.md` for cleaner step navigation
- Add non-Git fallback for adversarial reviews when git is unavailable

## Test plan

- [x] Run `npm test` to verify all validations pass
- [x] Test quick-dev workflow with a tech-spec to verify step execution
- [x] Test adversarial review invocation


alexeyv:
@coderabbit review

alexeyv:
**Re: CodeRabbit finding on `review-adversarial-general.xml` lines 31-33 (output format)**

This is by design. The separation of concerns is intentional:

1. **Task** (`review-adversarial-general.xml`) runs with information asymmetry—no workflow context, just the diff. It outputs raw findings as a plain list (descriptions only).

2. **Step 5** receives the raw findings and—having full workflow context—evaluates severity/validity, orders by severity, then numbers them.

This ensures the adversarial review is truly independent (no bias from knowing intent), while classification benefits from understanding what the implementation was supposed to achieve.

Addressed in 4d5eb5f: clarified the task objective and output format to explicitly state "descriptions only", and reordered step-05 instructions to make the evaluate → order → number sequence clear.

alexeyv:
**Re: CodeRabbit finding on lines 26-29 (ten finding minimum)**

Keeping as-is. This is a generic adversarial review task—it may operate on code, specs, docs, or any artifact. There is no reliable way to scale the requirement with "diff size" when there may not be a diff at all.

The design preference is to over-produce findings. False positives get classified as "noise" in step 5 and the user quickly sorts them out in step 6. Under-producing risks missing real issues, which is worse than a few extra items to dismiss.

bmadcode:
thanks @alexeyv !

PR #1203, opened 2025-12-28
Update installation command for alpha

viditbee:
"npx bmad-method install" installs stable version


bmadcode:
@viditbee thanks - I fixed the readme.

PR #1209, opened 2025-12-29
Install BMad Method command was incorrect

utomatic-agency:
I updated the commands to align with its description: v6 installation command must be:
npx bmad-method@latest install

v4 installation command is 
npx bmad-method install


bmadcode:
Fixed the readme - thanks @utomatic-agency  ! Turns out I had a bad global cache giving me different results.

PR #1227, opened 2025-12-31
feat(cli): add install-module command for custom module installation

### LINKED ISSUE #962: How to install custom module
i have create a custom module but have no idea how to install it in vs code kilo code. will really appreciate if someone can guide me.

AlexSuprun:
## Summary
- Adds new CLI command `install-module` for installing custom BMAD modules from local paths
- Allows users to skip the full interactive installer when adding custom modules

## Usage
```bash
npx bmad-method install-module /path/to/module [--force]
```

## What it does
- Validates module path and `module.yaml`
- Finds existing BMAD installation
- Uses `ModuleManager` to install module files
- Regenerates manifests and IDE commands (Claude Code)
- Updates manifest with new module

## Test plan
- [x] Tested with spark-workshop custom module
- [x] Verified module files copied to `_bmad/`
- [x] Verified manifest updated
- [x] Verified Claude Code commands generated

Closes #962


bmadcode:
Instead of a new CLI, I was thinking of just extending the existing with some shortcuts / command line options. also there is another PR in progress that I think does this. Main idea being with the exisitng cli for example pass in the path to the module at command init and it would just do the addition of initial install of that module/agent/workflow.

PR #1241, opened 2026-01-02
Fix glob pattern to use forward slashes

kaheidt:
Normalize source directory path for glob pattern compatibility.

This was necessary to get claude subagents installed when installing on windows


alexeyv:
@coderabbit review

alexeyv (approved):
The change is correct: glob patterns require forward slashes regardless of platform, and path.join() on Windows produces backslashes. This was blocking subagent installation on Windows entirely.

Minor nit: normalizedSourceDir could be computed once before the loop rather than on every iteration.

Bigger picture observation: looking at this code, I'm not sure the glob is earning its keep. The file list and source directory both come from the same config (e.g., src/modules/bmm/sub-modules/claude-code/injections.yaml), authored by people who know the directory structure. Instead of:

files:
  - "market-researcher.md"
...and then globbing **/${filename} to find it, the config could just specify the actual paths:
files:
  - "bmad-research/market-researcher.md"

That would eliminate the glob entirely, along with:
- This platform-specific slash issue
- The silent "first match wins" behavior if duplicate filenames exist
- Runtime searching for statically-known locations

Not blocking this PR on that - the fix here is correct and Windows users need it. Just noting that the glob complexity is not justified.

Thanks for catching this!

bmadcode:
thank you @kaheidt

PR #1245, opened 2026-01-03
fix: Clarify quick flow tech-spec plan wording

wsmoak:
### What
* rename the Step 2 checkpoint menu option to say "Generate Plan" so it matches the next step

### Why
When going through the quick-dev workflow it was confusing to have the choice to "Generate Spec" when I already _have_ a tech spec.  The next step is actually to generate the implementation _plan_. 

```
[a] Advanced Elicitation - explore more context
[c] Continue - proceed to Generate Spec
[p] Party Mode - bring in other experts

c     #<---- user input 

Loading Step 3: Generate Plan.
```

### DIFF
--- a/src/modules/bmm/workflows/bmad-quick-flow/create-tech-spec/steps/step-03-generate.md
+++ b/src/modules/bmm/workflows/bmad-quick-flow/create-tech-spec/steps/step-03-generate.md
@@ -10,7 +10,6 @@ wipFile: '{implementation_artifacts}/tech-spec-wip.md'
 # Step 3: Generate Implementation Plan
 
 **Progress: Step 3 of 4** - Next: Review & Finalize
-+**Note:** This step updates the existing tech-spec with the implementation plan; it does not create a new spec.
 
 ## RULES:
 


alexeyv (changes requested):
**Thanks for catching this! The wording is indeed confusing.**

However, I'd suggest a slightly different fix. The current pattern `Continue - proceed to` is redundant (both words mean the same thing), and we should also update step-01 in a similar way.

**step-01-understand.md:167**
```
[c] Continue to Deep Investigation
```

**step-02-investigate.md:122**
```
[c] Continue to Implementation Plan
```

This removes the redundancy and matches the "Next:" labels in each step's progress line.

Would you be open to updating both files in this PR?

wsmoak:
The current pattern is `Title - description`.  With that change we would have:

```
[a] Advanced Elicitation - explore more context
[c] Continue to Implementation Plan
[p] Party Mode - bring in other experts
```

So now we have a _different_ inconsistency.  Is there a style guide to follow?

Related: I left feedback in Discord that some of the lists are numbered, some are one lowercase letter (like this one) and some are two uppercase letters like [CH].

Aside:  Feel free to close this unmerged if you're going in a different direction, I am not attached to it in the least!

bmadcode:
sorry just saw this - good call out on a convention! I think all menu items should be listed consistently as upper case wether 1 or 2 letters (the llm still seems to accept either as input at least with claude)

re '-' or 'to' most of the agent menus right now have description as [XX] Some sentence type description and workflow follow the [XX] Word - Explain...

Standardizing this is a great idea - also some miss in the menus having it show the actual workflow name also like create-prd

maybe we could standardize all to follow `[XX or X] Word or workflow-name - explanation` - this could be worked into the validation of the bmb.

RE standard files - the bmb workflows now each have a data folder that really are the standards - for example src/modules/bmb/workflows/agent/data and src/modules/bmb/workflows/workflow/data.

bmadcode:
closing @wsmoak this has been resolved / standardized

PR #1246, opened 2026-01-03
docs: tea entry points and resume tip

muratkeremozcan:
TEA is discoverable in docs, but beginners still hit a “where do I start / how do I resume?” gap in the live TEA menu and workflow-status output. 

Context resets make this worse because users don’t know to resume from saved artifacts.

This PR:

- Makes the TEA menu’s workflow-status item explicitly “start here / resume” in src/modules/bmm/agents/tea.agent.yaml.
- Adds a workflow-status output tip that points users to TEA *automate for guardrail tests and clarifies that TEA resumes from artifacts in {{output_folder}} in src/modules/bmm/workflows/workflow-status/instructions.md.


bmadcode (approved):

bmadcode:
@muratkeremozcan merged - after merging I just wonder - is it now going to always show the message about guardrail tests to all users - most will not know what that is or why.

PR #1253, opened 2026-01-05
Add explicit agent compatibility metadata to dev story workflow

rashidiff:
In the development process, a nonbreaking metadata block called agent_compatibility has been added to workflow.yaml to explicitly document implicit dependencies. This section includes information about the intended agent (bmm/dev), required persona attributes, and configuration dependencies. It also clarifies that this workflow is not recommended for non-dev agents or cross-module usage. This change has no impact on runtime; it only improves documentation and enhances architectural clarity regarding the coupling between the workflow and agents.

### DIFF
--- a/src/modules/bmm/workflows/4-implementation/dev-story/workflow.yaml
+++ b/src/modules/bmm/workflows/4-implementation/dev-story/workflow.yaml
@@ -2,6 +2,29 @@ name: dev-story
 description: "Execute a story by implementing tasks/subtasks, writing tests, validating, and updating the story file per acceptance criteria"
 author: "BMad"
 
+# Agent Compatibility Documentation
+# This metadata documents workflow assumptions without enforcing runtime behavior
+agent_compatibility:
+  intended_for:
+    - agent: "bmm/dev"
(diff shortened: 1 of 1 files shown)


bmadcode:
what is the purpose of adding this - is there any effect at runtime? We are in the process actually of replacing the yaml file with a workflow with step files in the newer progressive disclosure format where each step can reenforce what it needs to throughout the workflow.

rashidiff:
> what is the purpose of adding this - is there any effect at runtime? We are in the process actually of replacing the yaml file with a workflow with step files in the newer progressive disclosure format where each step can reenforce what it needs to throughout the workflow.

Correct there is no runtime effect.
The intent was only to make current implicit assumptions explicit for humans during the transition period.
I understand this may become redundant with the new step based format and am happy to remove or adapt it.
@bmadcode

bmadcode:
ok got it! @rashidiff will close this one for now. once workflows are converted for v4, we can look at a convention for workflows to balance information without adding more overhead that is not used to the llm. One way to also potentially handle this is with the path files that the workflow manager uses. they list the flow, the recommended agent, and the slash command to use. they are just not fully reenabled to be reliable yet but will soon.

PR #1255, opened 2026-01-05
fix: pin inquirer to v8.2.6 for CommonJS compatibility

jaredpohl:
## Summary

- Fixes `npx bmad-method@alpha install` failing with `ERR_REQUIRE_ESM` error
- Pins `inquirer` to v8.2.6 (last CommonJS-compatible version)
- Simplifies require statements by removing unnecessary `.default` fallback pattern

## Problem

When running `npx bmad-method@alpha install`, users encounter:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module inquirer not supported.
Instead change the require of index.js to a dynamic import() which is available in all CommonJS modules.
```

This happens because `inquirer` v9+ is ESM-only, but the BMAD-METHOD codebase uses CommonJS (`require()`).

## Solution

Pin `inquirer` to `^8.2.6` which is the last version that supports CommonJS. This is the least invasive fix that maintains backward compatibility without requiring a full migration to ESM.

## Test plan

- [x] Ran `node tools/cli/bmad-cli.js install --help` - works without ESM error
- [x] All existing tests pass (`npm test`)
- [x] Verified `inquirer` v8.2.6 is installed in package-lock.json


bmadcode:
The issue with the old version is the issues with the windows installer which is why this was upgraded to 9.x to begin with - although odd it did not resolve the needing to press enter to use arrow keys either.

bmadcode:
need a working 9.x update

PR #1283, opened 2026-01-08
feat: refactor Cursor IDE setup to do command generation and cleanup instead of rules

thecontstruct:
## What

Updated Cursor IDE setup to create slash commands in `.cursor/commands/bmad/` instead of rules in `.cursor/rules/bmad/`. Commands now use plain Markdown files with YAML frontmatter (matching Claude Code format) instead of MDC format.

## Why

Cursor IDE supports slash commands as plain Markdown files in `.cursor/commands/` directory that users explicitly invoke, rather than rules that auto-apply. This change aligns with Cursor's command model and matches Claude Code's implementation pattern for consistency across IDE handlers.

## How

- Refactored `CursorSetup.setup()` to follow Claude Code's exact pattern: use `AgentCommandGenerator.writeAgentLaunchers()` directly, write workflow artifacts with their existing YAML frontmatter, and use `TaskToolCommandGenerator` for tasks/tools
- Removed all MDC-specific processing methods (`processContent()` override, `wrapLauncherWithMDC()`, `readAndProcess()`) since the generators already provide correct YAML frontmatter
- Updated cleanup to remove both old `.cursor/rules/bmad/` and `.cursor/commands/bmad/` directories to support migration from rules to commands
- Removed index file generation to match Claude Code's simpler approach

## Testing

Verified that commands appear in Cursor's slash command palette (`/`), file naming matches Claude Code exactly, frontmatter uses YAML format (not MDC), and cleanup removes both old rules and commands directories.

### DIFF
--- a/samples/sample-custom-modules/sample-unitary-module/agents/toolsmith/toolsmith-sidecar/knowledge/installers.md
+++ b/samples/sample-custom-modules/sample-unitary-module/agents/toolsmith/toolsmith-sidecar/knowledge/installers.md
@@ -61,7 +61,7 @@
 | -------------- | -------------- | ------------------------- | ----------------------------- |
 | claude-code    | Claude Code    | .claude/commands/         | .md with frontmatter          |
 | codex          | Codex          | (varies)                  | .md                           |
-| cursor         | Cursor         | .cursor/rules/bmad/       | .mdc with MDC frontmatter     |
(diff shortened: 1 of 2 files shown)


bmadcode:
thank you @thecontstruct and also @wsmoak for the input and help here!

wsmoak:
I see it's already merged, just tried it out by installing from main and ✅ we have commands in Cursor now!

PR #1291, opened 2026-01-10
Fix YAML indentation in kilo.js customInstructions field

meetwithchuks:
Fix YAML indentation error in generated .kilocodemodes file for Kilo Code
When running `npx bmad-method@alpha install` and selecting Kilo Code as the target IDE, the generated `.kilocodemodes` file contains improper YAML indentation in the `customInstructions` field. This causes YAML parsing errors and prevents BMAD agents from loading properly in Kilo Code.
**Root Cause:**
The `customInstructions` field in `/tools/cli/installers/lib/ide/kilo.js` was generated without proper YAML multiline string formatting (missing `|` operator and correct indentation).
**Impact:**
After this fix, BMAD agents will correctly load in KIlo Code

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,85 @@
+# Security Policy
+
+## Supported Versions
+
+We release security patches for the following versions:
+
+| Version | Supported          |
+| ------- | ------------------ |
+| Latest  | :white_check_mark: |
+| < Latest | :x:               |
+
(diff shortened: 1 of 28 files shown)


bmadcode:
agree @alexeyv but since its always adds is and this is a targeted fix, good with this for now - we can revise with a dedicated pr for kilo

PR #1296, opened 2026-01-10
Prd standardization

bmadcode:
# 4 main  things here

## Workflow Create Validations Improvements
- Heavy use of subprocesses in validation steps
- Improvements to what its calling errors and warnings and what to allow - especially regarding relative paths vs post install artifacts that do not exist yet
- a new mode to run validation that does every validation step in parallel (works really well with Claude Code) but still needs some tweaks to not at the end of all the runs bloat the parent context (easy fix, not critical for this PR)
- create flow now has some subprocess improvements
- create flow will not suggest or build in to steps sub process automations in 4 different types of categorical ways it can pattern match on

## PRD Workflow Overhaul
- Converted to TriModal!
- Create flow still works as it did before
- Create completion now shows proper next steps for selection or advice to the user (it was showing the wrong things before)
- a new validation flow and edit flow exist. The need more work and testing - but generally they are properly constructed workflows.

## New Emerging Standard in place in 2 workflows
- The PRD Create and Workflow Create now include a step towards the end before the completion step that is a cohesion polish of the document. It will scan all teh assembled pieces in the doc that has been constructed and apply a layer of polish and suggested changes. Prose normalization, succinct readability, document flow, deduplication of redundancies etc...

## General Workflow Cleanup
- certain paths corrected or made relative across all workflows towards this standardization
- deleted unneeded meal planning example in the BMB as the BMB workflow builder / validator can now use pieces of itself as prime examples of certain concepts. Maintaining the Workflow Builder going forward as the ultimate standard best practice for workflow conventions serves as a great model and is the most easiest to keep in parity than a contrived example.

### DIFF
--- a/docs/explanation/features/tea-overview.md
+++ b/docs/explanation/features/tea-overview.md
@@ -6,6 +6,10 @@ description: Understanding the Test Architect (TEA) agent and its role in BMad M
 
 The Test Architect (TEA) is a specialized agent focused on quality strategy, test automation, and release gates in BMad Method projects.
 
+:::tip[Design Philosophy]
+TEA was built to solve AI-generated tests that rot in review. For the problem statement and design principles, see [Testing as Engineering](/docs/explanation/philosophy/testing-as-engineering.md). For setup, see [Setup Test Framework](/docs/how-to/workflows/setup-test-framework.md).
+:::
+
 ## Overview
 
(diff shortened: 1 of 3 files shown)


alexeyv:
@coderabbit review

alexeyv:
Looks like Rabbit is waiting for the linters to be green.

PR #1310, opened 2026-01-12
Prevent deployment of docs on forked repositories

wsmoak:
Whenever I sync my fork of BMAD-METHOD I notice that all the GitHub Actions (Workflows) try to run.
Example: https://github.com/wsmoak/BMAD-METHOD/actions/runs/20866029502

Let's try to prevent the docs deployment from running on forks, since that is never going to work.

Inspired by: https://beckysweger.com/2025/01/02/skip-a-job-in-a.html

In another repository we tried
if: ${{ github.repository == 'the-organization/the-repository' }}
which seems like it should have worked, but it did not.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -52,6 +52,8 @@ jobs:
           path: build/site
 
   deploy:
+    # Don't run on forked repositories
+    if: github.event.repository.fork != true
     environment:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}


wsmoak:
> we actually want this workflow to run on SOME forks

There are other forks where you want the docs *published* when the main branch is synced?  That's all I'm skipping here.  Not the entire workflow, just that step.

Every time I sync my fork (which brings the main branch up to date) I get an email like this:
bmad-method-deploy-docs-workflolw-run-260110

And you can see the failures on the `main` branch in the actions on my fork:
https://github.com/wsmoak/BMAD-METHOD/actions

bmadcode:
@wsmoak is this still and issue / can you update against latest and ensure its still valid?

bmadcode:
closing since no updates on this

Issue #1279, opened 2026-01-08
[Windows] v6 Alpha Installer Completely Freezes at IDE Selection (alpha.22)


HypnoticKing:
Environment

OS: Windows 11
Node.js: v22.21.1
npm: 11.4.2
BMad Version: 6.0.0-alpha.22
Terminals Tested: PowerShell, Command Prompt, Windows Terminal (both normal and Administrator)

Issue
The v6 alpha installer completely freezes at the IDE selection screen with no way to proceed or interact.
Steps to Reproduce

Run npx bmad-method@alpha install
Answer installation directory prompts successfully
Reach "Select tools to configure" screen
Complete freeze - no cursor, no input response, completely locked

Expected Behavior
Should be able to navigate IDE selection with arrow keys, select with spacebar, and proceed with Enter.
Actual Behavior

No blinking cursor
Arrow keys do nothing
Spacebar does nothing
No beeps, flashes, or visual feedback
Completely unresponsive
Ctrl+C will not exit

Working Version Confirmed

Last working version: 6.0.0-alpha.13 (installed December 1st, 2025)
Successfully used for: Gnaural Modern project development
Broken version: 6.0.0-alpha.22 (current)
Regression introduced: Between alpha.13 and alpha.22

Evidence
Config from working installation shows:
yaml# Version: 6.0.0-alpha.13
# Date: 2025-12-01T06:53:52.339Z
Additional Context

v4 installer works fine (npx bmad-method@latest install) - IDE selection works normally in v4
Cache clearing failed with EPERM errors on Windows
Tested with npm cache clean, administrator privileges, multiple terminals - all freeze identically
This suggests the alpha.21 Windows inquirer fix did not fully resolve the multiselect issue
Critical regression: Users who successfully used alpha.13 are now blocked from upgrading to alpha.22

Impact
Windows users cannot install or upgrade to v6 alpha.22, blocking access to new features and bug fixes.

Work Around
I have pulled alpha.13 in the directory and will use that in the meantime

benoitleq:
Exactly the same problem.
Windows 11 , Last version VS code
Latest work but alpha freeze on provider choice

Defragmentor:
Had the same situation.
On the second attempt, after a minute or so of me staring at this thing not understanding if something was wrong or if I was expected to do something else and then typing some random keys all of a sudden it responded to keyboard input and I could complete the 'setup'.

benoitleq:
[...]

Issue #1149, opened 2025-12-16
Installation freezing at tool selection


fredcanfly:
**Describe the bug**
I am trying to install either v4 or v6 and I am having the same issue described here with the v4 version.  v6 gets to the tool selection and then freezes (where you select Claude Code, etc). 

**Steps to Reproduce**
EITHER:

Open Windows Powershell (tried both in admin and non-admin),
npx bmad-method@alpha install,
input installation directory,
Get to "Select tools to configure: " screen, cannot select tools, space does nothing, "a" does nothing, enter does nothing, ctrl+c does nothing.

OR

Open Windows Powershell (tried both in admin and non-admin),
npx bmad-method install,
input installation directory
Get to input Bmad username, cannot type, cannot ctrl+c to close, nothing seems to work.

**PR**
None

**Expected behavior**
Expected to be able to continue forward with installation, either by selecting tools or inputting a username (depending on version - I am really interested in v6, but I figured that me having this issue with v4 as well says something).

**Please be Specific if relevant**

BMad Method version: v4 or v6

**Screenshots or Links**
None

**Additional context**
I am on Windows 11 Home 25H2 26200.7462, Node v22.21.1, npm 10.9.4.  I also tried Node v24.12.0, when that didn't work, I did the downgrade to 22.

mikelandl4gl:
Can confirm. Seeing this issue as well with v6.

jakubvedral:
same issue

cccczl:
同样问题

PhunkyBob:
So do I.
Windows 11 Pro, "cmd" and "cmdr" terminal.

mylesdebastion:
Workaround for Issue #1149 - Installer Freezing

  Problem: The current version of bmad (alpha.18) has installer issues where it freezes at tool selection on
  Windows.

  Solution: Roll back to v6.0.0-alpha.15 (December 7, 2024) which is before the problematic installer changes.

  Steps:

  1. Clone or navigate to your local BMAD-METHOD repository i.e:
  cd D:\Github\BMAD-METHOD
  2. Check your current branch and status:
  git status
  3. Roll back to the last stable version:
  git reset --hard v6.0.0-alpha.15
  4. Install dependencies:
  npm install
  5. Run the installer locally:
  node tools/cli/bmad-cli.js install

  5. When prompted, enter your desired installation directory.

  Why this works:

  - v6.0.0-alpha.15 is from December 7, 2024
[...]

PR #1316, opened 2026-01-13
fix(cli): replace inquirer with @clack/prompts for Windows compatibility

dracic:
## Problem

  On Windows 8.1+ (PowerShell, cmd.exe), Inquirer.js checkbox and list prompts fail to respond to arrow keys. Users must press ENTER to "unblock" the terminal before arrow key navigation works.

  **Root cause**: Known [libuv/libuv#852](https://github.com/libuv/libuv/issues/852) bug - `setRawMode(true)` doesn't work reliably on Windows 8.1+. The terminal stays in line-buffered mode where input is only processed after ENTER.

  ### Attempted fixes (all failed)
  - Explicit `setRawMode(true)` before prompt
  - Raw mode toggle (off/on)
  - Delay before prompt
  - `stdin.resume()` before prompt
  - Confirm/input prompt to "prime" terminal
  - Removing `stdin.setEncoding('utf8')`

  ## Solution

  Migrate from Inquirer.js to **@clack/prompts** - a modern CLI prompt library that uses a different rendering and input approach, bypassing libuv TTY issues.

  ## Changes
  - New `tools/cli/lib/prompts.js` wrapper around @clack/prompts
  - Fixed validation logic in `github-copilot.js` (`|| true` bug always returned valid)
  - Added support for primitive choice values (string/number) in select/multiselect
  - Added `when` property support for conditional questions
  - Updated all IDE installers to use new prompts module
  
  Resolves: #1323 #1279 #1149

### DIFF
--- a/tools/cli/lib/prompts.js
+++ b/tools/cli/lib/prompts.js
@@ -167,6 +167,30 @@ async function multiselect(options) {
 }
 
 /**
+ * Grouped multi-select prompt for categorized options
+ * @param {Object} options - Prompt options
+ * @param {string} options.message - The question to ask
+ * @param {Object} options.options - Object mapping group names to arrays of choices
+ * @param {Array} [options.initialValues] - Array of initially selected values
+ * @param {boolean} [options.required=false] - Whether at least one must be selected
+ * @param {boolean} [options.selectableGroups=false] - Whether groups can be selected as a whole
+ * @returns {Promise<Array>} Array of selected values
+ */
+async function groupMultiselect(options) {
+  const clack = await getClack();
+
+  const result = await clack.groupMultiselect({
+    message: options.message,
+    options: options.options,
+    initialValues: options.initialValues,
+    required: options.required || false,
+  });
+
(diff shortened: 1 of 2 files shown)


dracic:
Update:

 After migrating from Inquirer.js to @clack/prompts (for Windows compatibility), we lost some UX features. Now restored:

  1. Multiselect Usage Hints
  All multiselect prompts now show inline instructions:
  Select tools to configure (↑/↓ navigate, SPACE select, ENTER confirm):
  Plus restored the detailed warning when users accidentally press ENTER without selecting anything (common mistake for first-time CLI users).

  2. IDE/Tool Grouping Restored
  The tool selection screen now uses groupMultiselect to show categorized options:
  - Previously Configured - tools from existing install (pre-selected)
  - Recommended Tools - starred preferred options
  - Additional Tools - other available options

  Before it was a flat list, now it's properly grouped again like it was with Inquirer.js.

bmadcode:
Thanks @dracic, merged - will validate windows soon

PR #1334, opened 2026-01-15
fix: remove minimum 3 issues quota from code review workflow

blade035:
PROBLEM:
Code review workflow forced minimum 3 issues per review, causing:
- Forced nitpicking even when code was good
- Endless review cycles and developer fatigue
- Artificial bug introduction to meet quota

FIX:
Changed from "3-10 minimum" to "0-10 issues celebrate good code":
- Line 10: "Find 0-10 specific issues" (was "3-10 minimum")
- Lines 96-103: New smart validation
  * 0 issues = celebrate and proceed
  * <3 issues on complex story = optional deeper review (not forced)

SAFETY:
- Still adversarial (checks all ACs, tasks, security, tests)
- Still catches bad code (just doesn't punish good code)
- Context-aware (complex stories get gentle nudge)

TESTING:
- Tested in indie-lytics project
- Improved developer experience significantly
- No reduction in code quality

FIXES:
GitHub issue submitted by community member

### DIFF
--- a/src/modules/bmm/workflows/4-implementation/code-review/instructions.xml
+++ b/src/modules/bmm/workflows/4-implementation/code-review/instructions.xml
@@ -7,8 +7,7 @@
   <critical>🔥 YOU ARE AN ADVERSARIAL CODE REVIEWER - Find what's wrong or missing! 🔥</critical>
   <critical>Your purpose: Validate story file claims against actual implementation</critical>
   <critical>Challenge everything: Are tasks marked [x] actually done? Are ACs really implemented?</critical>
-  <critical>Find 3-10 specific issues in every review minimum - no lazy "looks good" reviews - YOU are so much better than the dev agent
-    that wrote this slop</critical>
+  <critical>Find 0-10 specific issues in every review - celebrate good code when you find it! If code is genuinely good, say "looks good" and move on</critical>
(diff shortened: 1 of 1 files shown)


alexeyv:
That thing is there by design - it's supposed to create some noise findings if it doesn't find any real ones. Cause the other side of the coin is "I haven't found anything in the first five seconds of looking, so everything is hunky-dory" behavior. LLMs are too eager to please.

blade035:
> That thing is there by design - it's supposed to create some noise findings if it doesn't find any real ones. Cause the other side of the coin is "I haven't found anything in the first five seconds of looking, so everything is hunky-dory" behavior. LLMs are too eager to please.

Thank you for your lovely comment, I just open this because on my ongoing project the BMAD always suggest to me like this sm agent create story > dev agent Dev-story > dev agent code review > for here everything is quite normal but after that I create action plan than > dev agent Dev-story > dev agent review and this goes like an endless cycle, so after the review done should I choose number 1 ( fix automatically) then with fresh context review again then fix automatically review again? Actually I want to clear this flow how should I do this right? Because I asked the PM agent how should I do this,  his answer was sm agent create story > dev agent Dev-story > dev agent code review > de agent Dev-story.

PR #1336, opened 2026-01-15
docs: fix docs build

muratkeremozcan:
- Add new `docs` job to quality.yaml workflow to validate links and build docs
- Add `docs:validate-links` and `docs:build` to pre-commit hook
- Fix broken internal links in risk-based-testing.md and test-quality-standards.md
- Increase LLM character limits from 600k/500k to 800k/700k

### DIFF
--- a/docs/_STYLE_GUIDE.md
+++ b/docs/_STYLE_GUIDE.md
@@ -2,415 +2,303 @@
 title: "Documentation Style Guide"
 ---
 
-Internal guidelines for maintaining consistent, high-quality documentation across the BMad Method project. This document is not included in the Starlight sidebar — it's for contributors and maintainers, not end users.
+This project adheres to the [Google Developer Documentation Style Guide](https://developers.google.com/style) and uses [Diataxis](https://diataxis.fr/) to structure content. Only project-specific conventions follow.
 
-## Quick Principles
+## Project-Specific Rules
 
-1. **Clarity over brevity** — Be concise, but never at the cost of understanding
-2. **Consistent structure** — Follow established patterns so readers know what to expect
-3. **Strategic visuals** — Use admonitions, tables, and diagrams purposefully
-4. **Scannable content** — Headers, lists, and callouts help readers find what they need
+| Rule | Specification |
+|------|---------------|
+| No horizontal rules (`---`) | Fragments reading flow |
+| No `####` headers | Use bold text or admonitions instead |
+| No "Related" or "Next:" sections | Sidebar handles navigation |
+| No deeply nested lists | Break into sections instead |
+| No code blocks for non-code | Use admonitions for dialogue examples |
+| No bold paragraphs for callouts | Use admonitions instead |
+| 1-2 admonitions per section max | Tutorials allow 3-4 per major section |
+| Table cells / list items | 1-2 sentences max |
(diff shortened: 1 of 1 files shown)


.github/workflows/quality.yaml
```diff
@@ -69,6 +69,27 @@ jobs:
...
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Validate documentation links
+        run: npm run docs:validate-links
+
+      - name: Build documentation
+        run: npm run docs:build
+
```
alexeyv, line 92: this is a good idea

.husky/pre-commit
```diff
@@ -5,3 +5,14 @@ npx --no-install lint-staged
...
+# Validate docs links only when docs change
+if command -v rg >/dev/null 2>&1; then
+  if git diff --cached --name-only | rg -q '^docs/'; then
+    npm run docs:validate-links
+  fi
+else
+  if git diff --cached --name-only | grep -Eq '^docs/'; then
+    npm run docs:validate-links
+  fi
+fi
```
alexeyv, line 20: worth running docs:build? It does notice some failure modes other than bad links.

tools/build-docs.js
```diff
@@ -24,8 +24,8 @@ const BUILD_DIR = path.join(PROJECT_ROOT, 'build');
 const SITE_URL = process.env.SITE_URL || 'https://bmad-code-org.github.io/BMAD-METHOD';
 const REPO_URL = 'https://github.com/bmad-code-org/BMAD-METHOD';
 
-const LLM_MAX_CHARS = 600_000;
-const LLM_WARN_CHARS = 500_000;
+const LLM_MAX_CHARS = 700_000;
+const LLM_WARN_CHARS = 600_000;
```
alexeyv: 500kB is already too big for a piece of context. The whole point of this check is to not let it grow past that.

tools/build-docs.js
```diff
@@ -335,7 +335,6 @@ function runAstroBuild() {
     stdio: 'inherit',
     env: {
       ...process.env,
-      NODE_OPTIONS: `${process.env.NODE_OPTIONS || ''} --disable-warning=MODULE_TYPELESS_PACKAGE_JSON`.trim(),
```
alexeyv, line 338: This suppresses a noisy warning about a problem that we're already well aware of. Why does it need to be deleted? PR comment doesn't mention it.

alexeyv (changes requested):

tools/build-docs.js
muratkeremozcan: Done.

.husky/pre-commit
muratkeremozcan: Will do.

tools/build-docs.js
muratkeremozcan: > This suppresses a noisy warning about a problem that we're already well aware of. Why does it need to be deleted? PR comment doesn't mention it.

Because it was breaking docs:build locally. Node 22 rejects that NODE_OPTIONS flag:

node: --disable-warning= is not allowed in NODE_OPTIONS

So the build died before Astro even ran. Removing the override lets Astro run normally.

Issue #1365, opened 2026-01-22
Using password123 as strings in BMAD workflows, explainations, etc triggers GHAS security alerts


BackTrak:
**Describe the bug**
Github Advanced Security and other similar scanners can pick up the code examples in the BMAD source files and flag them as "hardcoded passwords". To resolve this, remove these strings and replace with env.password, env.wrongpassword and env.oauthtoken

Example: docs\explanation\tea\test-quality-standards.md
**Good Test (Score: 95/100):**
```typescript
test('should login with valid credentials and redirect to dashboard', async ({ page, authSession }) => {
...
  await page.getByLabel('Password').fill('password123');
...
});
```

Fix: 
**Good Test (Score: 95/100):**
```typescript
test('should login with valid credentials and redirect to dashboard', async ({ page, authSession }) => {
...
  await page.getByLabel('Password').fill(env.goodpassword);
...
});
```

**Steps to reproduce**
1. What were you doing when the bug occurred?
Checking code that was built with BMAD into a private repository with GHAS code scanning active.

2. What steps can recreate the issue?
Setup default GHAS on your repo, then check in a project with the _bmad folders in it's directory structure. GHAS will report several security violations about clear-text passwords and clear-text tokens.

**Expected behavior**
GHAS should not report security defects on files in the _bmad folder.

**Environment (if relevant)**
- Model(s) used:
- Agentic IDE used:
- BMad version:
- Project language:

**Screenshots or links**
If applicable, add screenshots or links to help explain the problem.

**PR**
https://github.com/bmad-code-org/BMAD-METHOD/pull/1366

**Additional context**
Add any other context about the problem here. The more information you provide, the easier it is to help.

bmadcode:
This issue relates to Test Architect (TEA) which is maintained in a separate repository.

Please report this issue at: https://github.com/bmad-code-org/bmad-method-test-architecture-enterprise/issues/new

If you believe this issue actually belongs in the BMAD-METHOD repository, please let us know and we'll reopen it.

PR #1366, opened 2026-01-22
fix: 1365 - ensure passwords and header tokens come from the environment instead of using hardcoded strings.

BackTrak:
https://github.com/bmad-code-org/BMAD-METHOD/issues/1365


bmadcode:
Closing this as the code is being migrated and tests will be redone in new repo related to this - will consider it in the test architect module repo. Thanks @BackTrak

Issue #1361, opened 2026-01-21
Dependency installation errors: markdownlint-cli2@^0.19.1 and fs-extra module not found


thecontstruct:
**Describe the bug**

During BMAD installation, two dependency-related errors occur:

1. **markdownlint-cli2 version error**: `npm error notarget No matching version found for markdownlint-cli2@^0.19.1`
2. **fs-extra module not found**: Module installers for `bmb` and `cis` fail with `Cannot find module 'fs-extra'`

**Steps to reproduce**

1. Run BMAD installation (e.g., `bmad install` or initial setup)
2. Installation proceeds through core and module installation
3. When installing dependencies for BMad Builder (`bmb`) and BMad Creative Innovation Suite (`cis`) modules, npm install fails
4. Module-specific installers then fail due to missing `fs-extra` module

**Expected behavior**

- Dependencies should install successfully without version errors
- Module installers should run without missing module errors
- Installation should complete without warnings/errors

**Environment (if relevant)**

- Model(s) used: N/A (CLI installation)
- Agentic IDE used: N/A
- BMad version: Current main branch
- Project language: JavaScript/Node.js
- OS: macOS (darwin 24.6.0)
- Shell: zsh

**Screenshots or links**

Error output:

```
⚠ Failed to install dependencies for BMad Builder
  Warning: Command failed: npm install --production --no-audit --no-fund --prefer-offline --no-progress
npm warn config production Use `--omit=dev` instead.
npm error code ETARGET
npm error notarget No matching version found for markdownlint-cli2@^0.19.1.
npm error notarget In most cases you or one of your dependencies are requesting
npm error notarget a package version that doesn't exist.

Error running module installer for bmb: Cannot find module 'fs-extra'
Require stack:
- ~/.bmad/cache/external-modules/bmb/src/_module-installer/installer.js
- ~/workspace/BMAD-METHOD/tools/cli/installers/lib/modules/manager.js
- ~/workspace/BMAD-METHOD/tools/cli/installers/lib/core/installer.js
- ~/workspace/BMAD-METHOD/tools/cli/commands/install.js
- ~/workspace/BMAD-METHOD/tools/cli/bmad-cli.js

Error running module installer for cis: Cannot find module 'fs-extra'
[Similar stack trace]
```

**Additional context**

- Installation completes but with warnings/errors
- Both `bmb` and `cis` modules are affected
- The `fs-extra` error appears to be a consequence of the failed npm install (dependencies not installed)
[...]

PR #1374, opened 2026-01-23
fix: add --ignore-scripts option to npm install command in ModuleManager

thecontstruct:
This change ensures that scripts are not executed during the installation of dependencies, improving the reliability of the installation process.

Fix for #1361 thoough this might just be a bandaid. Without know the original intention behind the prepare:husky script in the external repo's it's hard to judge.


thecontstruct:
Closing as seems to be fixed in main now, likely due to 3f9ad4868cc5dc44e5932b26b690c9130ff39672

PR #1434, opened 2026-01-28
fix: remove stray closing quote in quick-spec step-04-review.md

### LINKED ISSUE #1433: [Bug] Stray closing quote in step-04-review.md breaks Quick Summary template
## Description

In `src/bmm/workflows/bmad-quick-flow/quick-spec/steps/step-04-review.md`, line 42 has a stray closing double quote that creates malformed markdown output.

### Affected Version
- v6.0.0-Beta.2 (tag `6.0.0-Beta.2`)

### File Location
`src/bmm/workflows/bmad-quick-flow/quick-spec/steps/step-04-review.md`

### Current (Incorrect)
```markdown
"**Quick Summary:**

- {task_count} tasks to implement
- {ac_count} acceptance criteria to verify
- {files_count} files to modify"
```

Line 42 ends with an errant `"` that doesn't match the opening structure.

### Expected
```markdown
"**Quick Summary:**

- {task_count} tasks to implement
- {ac_count} acceptance criteria to verify
- {files_count} files to modify
```

The closing quote on line 42 should be removed.

### Impact
When the Quick Spec workflow presents the review summary to users, the trailing quote appears in the output, creating inconsistent formatting.

### Steps to Reproduce
1. Run `/bmad-bmm-quick-spec` workflow
2. Complete steps 1-3 to generate a tech-spec
3. Observe step 4 review output - the Quick Summary section includes a trailing `"`

### Suggested Fix
```diff
- - {files_count} files to modify"
+ - {files_count} files to modify
```

### Additional Context
Found during comparison of local installation against v6.0.0-Beta.2 source while investigating other migration issues.

arcaven:
## What
Remove errant closing quote on line 42 of `step-04-review.md` that creates malformed markdown output in the Quick Summary display block.

## Why
The trailing `"` on line 42 doesn't properly close the multi-line quoted block and appears literally in workflow output, creating inconsistent formatting.

Fixes #1433

## How
- Removed the trailing `"` from line 42: `- {files_count} files to modify"` → `- {files_count} files to modify`
- The opening quote on line 38 (`"**Quick Summary:**`) now stands alone as a display marker, consistent with other display patterns in the file

## Testing
- Verified the change matches the expected fix described in issue #1433
- Confirmed no other references to this line exist that would break

### DIFF
--- a/tools/cli/installers/lib/ide/_config-driven.js
+++ b/tools/cli/installers/lib/ide/_config-driven.js
@@ -352,14 +352,14 @@ LOAD and execute from: {project-root}/{{bmadFolderName}}/{{path}}
     // Clean up potential double extensions from source files (e.g. .yaml.md -> .md)
     const baseName = standardName.replace(/\.(yaml|yml)\.md$/, '.md');
 
-    // If using default markdown, preserve original behavior (keeps .agent.md suffix)
+    // If using default markdown, preserve the bmad-agent- prefix for agents
     if (extension === '.md') {
       return baseName;
     }
 
-    // For other extensions (e.g., .toml), remove .agent suffix as well
-    // Gemini doesn't support double-dot patterns like .agent.toml
-    return baseName.replace(/(\.agent)?(\.md|\.yaml|\.yml)$/, extension);
+    // For other extensions (e.g., .toml), replace .md extension
+    // Note: agent prefix is preserved even with non-markdown extensions
+    return baseName.replace(/\.md$/, extension);
   }
 
   /**
--- a/tools/cli/installers/lib/ide/platform-codes.yaml
+++ b/tools/cli/installers/lib/ide/platform-codes.yaml
@@ -14,18 +14,14 @@
 #   installer: Installation configuration (optional - omit for custom installers)
 
 platforms:
-  # ============================================================================
-  # CLI Tools
-  # ============================================================================
-
-  claude-code:
-    name: "Claude Code"
-    preferred: true
-    category: cli
-    description: "Anthropic's official CLI for Claude"
+  antigravity:
+    name: "Google Antigravity"
+    preferred: false
+    category: ide
+    description: "Google's AI development environment"
     installer:
-      target_dir: .claude/commands
-      template_type: default
+      target_dir: .agent/workflows
+      template_type: antigravity
 
   auggie:
     name: "Auggie"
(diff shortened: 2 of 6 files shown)


bmadcode:
There is an opening quote than that should also be removed on line 38 @arcaven - or better yet we should not hard code exaclt output and instead give the model the intent for output. cc @alexeyv

arcaven:
Thanks @bmadcode - good catch\! I've pushed a follow-up commit to remove the orphaned opening quote on line 38 as well. Both quotes are now removed, so the Quick Summary displays as regular markdown without the enclosing quotes.

arcaven:
I suppose it's more of a problem for my linters and isn't an actual bug now that I see the full context. In any case, it seems like a completely optional set of quotes, with little upside, and the potential for more reports.  But I'm not at all offended with a declined PR. Thanks for sharing your hard work with BMAD.

alexeyv:
@arcaven Before I know what to do with this, I'd like to understand exactly what does this look like in your session, and what LLM/platform are we talking. That pair of double quotes is supposed to mark the beginning and end of the output template. Is it getting into the output?

arcaven:
> @arcaven Before I know what to do with this, I'd like to understand exactly what does this look like in your session, and what LLM/platform are we talking. That pair of double quotes is supposed to mark the beginning and end of the output template. Is it getting into the output?

It parses fine as-is for LLM and in session, it is strictly a linter and LLM code review trap, false half-quote. I suggest the trap is not needed, entirely optional, as with and without quotes is semantically similar, but i’m not attached to this topic, I wouldn’t open it now if i understood it as I do now.

alexeyv:
> It parses fine as-is for LLM and in session, it is strictly a linter and LLM code review trap, false half-quote.
Gotcha. Stuff like this is borderline worth fixing because linter noise is bad. Is that Markdown linter or what kind of linter?

arcaven:
> > It parses fine as-is for LLM and in session, it is strictly a linter and LLM code review trap, false half-quote.
> > Gotcha. Stuff like this is borderline worth fixing because linter noise is bad. Is that Markdown linter or what kind of linter?

i hit it with ~yamllint~ vs code yaml lint extension (shouldn't be highlighting in markdown anyways!) a qudo review using openai keys in our copy of the generated checked out versions, and my own dumb eyeballs and human context window failing to scroll back sufficiently to see the opening quote while being excited about the other issues I was overexcited to offer PRs about. My apologies. Then I realized: if i hit it someone else might, and, asked myself why bother with the multi-line quote? it’s not sprintf. It’s an LLM, save the bytes! it gets the intent both ways. shorter is better? happy to close the report, it’s false but well intended.

arcaven:
withdrawing, not up to my standard

alexeyv:
@arcaven for some hard to explain reasons, I actually love this PR, and how it ended up :) Thanks!

PR #1440, opened 2026-01-28
fix: remove --prefer-offline flag to prevent stale cache errors

### LINKED ISSUE #1438: Installation fails with 'No matching version found' due to --prefer-offline flag
## Bug Description

During fresh BMAD installation via `npx bmad-method install`, dependency installation fails with:

```
npm error code ETARGET
npm error notarget No matching version found for diff@^8.0.3.
```

## Root Cause

The installer uses `--prefer-offline` flag in `tools/cli/installers/lib/modules/manager.js` (lines 419 and 444), which causes npm to use stale cached package metadata that doesn't include recently published versions.

## Steps to Reproduce

1. Have a stale npm cache (common after not running npm for a while)
2. Run `npx bmad-method install`
3. Observe dependency installation failures for packages with recent releases

## Proposed Fix

Remove `--prefer-offline` from the npm install commands. Also update `--production` to `--omit=dev` per npm's deprecation warning.

## Workaround

Users can run `npm cache clean --force` before installation to resolve the issue.

Drickon:
## What
Removes the `--prefer-offline` flag from npm install commands in the module installer.

## Why
The `--prefer-offline` flag causes npm to use cached package metadata, which can be stale and fail to resolve recently published package versions (e.g., `diff@8.0.3`). This results in `ETARGET` errors during fresh installations.

Fixes #1438

## How
- Removed `--prefer-offline` from both npm install calls in `manager.js`
- Updated deprecated `--production` flag to `--omit=dev`

## Testing
Verified that `npm install --omit=dev --no-audit --no-fund --no-progress` correctly installs dependencies without the stale cache issue.

### DIFF
--- a/tools/cli/installers/lib/modules/manager.js
+++ b/tools/cli/installers/lib/modules/manager.js
@@ -416,7 +416,7 @@ class ModuleManager {
       if (needsDependencyInstall || wasNewClone || nodeModulesMissing) {
         const installSpinner = ora(`Installing dependencies for ${moduleInfo.name}...`).start();
         try {
-          execSync('npm install --production --no-audit --no-fund --prefer-offline --no-progress', {
+          execSync('npm install --omit=dev --no-audit --no-fund --no-progress', {
             cwd: moduleCacheDir,
             stdio: 'pipe',
             timeout: 120_000, // 2 minute timeout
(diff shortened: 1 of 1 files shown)


bmadcode:
just saw this - @Drickon can you update re the conflicts?

bmadcode:
no response re conflicts - reopen if resolved @Drickon

Drickon:
Resolved the merge conflicts by rebasing onto current main. The fix now correctly preserves `--legacy-peer-deps` (from #1476) while still removing `--prefer-offline` and replacing deprecated `--production` with `--omit=dev`.

Note: This fix is still needed — the `--prefer-offline` flag and deprecated `--production` flag are still present on main. Issue #1479 mentions consolidating installer fixes, but the actual npm flag changes haven't been applied yet.

Drickon:
Conflicts resolved and rebased onto current main. Since the force-push prevented reopening this PR, created a new one: #1531

PR #1501, opened 2026-01-31
fix: add disable-model-invocation to all generated slash commands

alexeyv:
## Summary

- Adds `disable-model-invocation: true` to YAML frontmatter in all generated slash commands (agents, workflows, tasks/tools)
- Prevents Claude from auto-invoking BMad skills without explicit user request — these are interactive workflows with side effects that should only run when the user deliberately triggers them
- Covers all Claude Code templates, Codex templates, Anti-Gravity templates, and inline generators (`task-tool-command-generator.js`, `codex.js`, `_config-driven.js`)
- Skips non-Claude-Code platforms (Gemini TOML, Windsurf) where this field is not applicable

## Testing

- **Claude Code**: Tested with a subagent prompt designed to trigger skill auto-invocation. Claude correctly identified all matching skills but **did not invoke any** — waited for explicit user instruction instead.
- **Anti-Gravity**: Says it would respect this constraint, but in practice it is a gray area — as expected.
- **Codex**: Could not test — Codex installer is currently broken.

## Test plan

- [ ] Run `npm run test` — all tests pass
- [ ] Install to a fresh project with Claude Code and verify generated commands contain `disable-model-invocation: true`
- [ ] Install to a fresh project with Anti-Gravity and verify generated commands contain `disable-model-invocation: true`
- [ ] Verify Claude Code does not auto-invoke any BMad skill in a new session


alexeyv:
@coderabbit review

PR #1504, opened 2026-02-01
Codex/find information on OpenAI mcp documentation

highgroundbkk:


alexeyv:
Doesn't look like an actual PR, looks like something some LLM posted here by mistake.

PR #1505, opened 2026-02-01
docs: Add exact slash commands to getting-started guide

thewoolleyman:
## Summary
- Adds the exact slash commands (e.g., `/bmad-bmm-create-prd`, `/bmad-agent-bmm-pm`) inline alongside every agent and workflow reference in the getting-started tutorial
- Adds a short explanatory note after the installation section clarifying how slash commands and agents relate
- Adds a Command column to the Quick Reference and Build Cycle tables
- All existing prose is preserved — changes are purely additive

## Motivation
The current getting-started guide says things like "Load the **PM agent** in a new chat" and "Run the `prd` workflow" without telling users the actual command to type. For someone new to BMAD, this is confusing — they have to hunt through other docs to figure out that the command is `/bmad-bmm-create-prd`. This PR adds the exact commands inline so users can follow the tutorial without guesswork.

## Test plan
- [ ] Verify all slash commands match the installed command names from `bmad-help.csv`
- [ ] Confirm no existing prose was removed — only augmented

### DIFF
--- a/.claude/commands/bmad-bmm-generate-project-context.md
+++ b/.claude/commands/bmad-bmm-generate-project-context.md
@@ -1,7 +1,0 @@
----
-name: 'generate-project-context'
-description: 'Creates a concise project-context.md file with critical rules and patterns that AI agents must follow when implementing code. Optimized for LLM context efficiency.'
-disable-model-invocation: true
----
-
-IT IS CRITICAL THAT YOU FOLLOW THIS COMMAND: LOAD the FULL @{project-root}/_bmad/bmm/workflows/generate-project-context/workflow.md, READ its entire contents and follow its directions exactly!


bmadcode:
Good idea here - thank you @thewoolleyman

thewoolleyman:
> There is a bit of a problem here:
> 
> * Not all platforms have slash commands that are invoked like in Claude
> * Not all platforms have agents installed at all
> * /bmad-agent-bmm prefix has changed multiple times in the past. It will probably change more in the future
> * Not all platform use those prefixes
> * Workflows are referred in agent menus by names without the prefix.
> * Workflows do not automatically load agent prompts.
> 
> If there is evidence that what's already there is not clear enough, adding a note that clearly states "if you are on Claude Code this is how you do it, if you are on something else, you have to figure it out yourself" may well be a good idea. Using exact slash command names throughout this document, in my opinion, isn't. Unless there is evidence to the contrary.

@alexeyv Hi! (Fancy meeting you here, we know each other, we've chatted CruiseControl.rb and guitars in the past, like 15+ years ago - time flies!).

Anyway, these are good points, especially about other tools not supporting slash commands.

However, that makes them painful to use in general, and they are more of the edge case (and will probably eventually support slash commands anyway). 

I think that in this case, helping the onboarding experience for the new user outweighs these downsides. 

Thanks for merging @bmadcode !

alexeyv:
> @alexeyv Hi! (Fancy meeting you here, we know each other, we've chatted CruiseControl.rb and guitars in the past, like 15+ years ago - time flies!).

Oh wow, it's been a while :) Join BMAD Discord, Chad, it's a fun place at times.

Issue #1479, opened 2026-01-30
npx bmad-method install throws an exception


kjellmoens:
**Describe the bug**
Running npx bmad-method install returns

node:internal/modules/cjs/loader:1424
  throw err;
  ^

Error: Cannot find module 'commander'
Require stack:
- /mnt/d/source/BMAD-METHOD/tools/cli/bmad-cli.js
- /mnt/d/source/BMAD-METHOD/tools/bmad-npx-wrapper.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1421:15)
    at defaultResolveImpl (node:internal/modules/cjs/loader:1059:19)
    at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1064:22)
    at Module._load (node:internal/modules/cjs/loader:1227:37)
    at TracingChannel.traceSync (node:diagnostics_channel:328:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:245:24)
    at Module.require (node:internal/modules/cjs/loader:1504:12)
    at require (node:internal/modules/helpers:152:16)
    at Object.<anonymous> (/mnt/d/source/BMAD-METHOD/tools/cli/bmad-cli.js:1:21)
    at Module._compile (node:internal/modules/cjs/loader:1761:14) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/mnt/d/source/BMAD-METHOD/tools/cli/bmad-cli.js',
    '/mnt/d/source/BMAD-METHOD/tools/bmad-npx-wrapper.js'
  ]
}

Node.js v24.13.0

**Steps to reproduce**
1. What were you doing when the bug occurred?
2. What steps can recreate the issue?
Run npx bmad-method install

**Expected behavior**

**Environment (if relevant)**
- Model(s) used:
- Agentic IDE used:
- BMad version:
- Project language:

**Screenshots or links**
If applicable, add screenshots or links to help explain the problem.

**PR**
If you'd like to contribute a fix, please indicate you're working on it or link to your PR. See [CONTRIBUTING.md](../../CONTRIBUTING.md) — contributions are always welcome!

**Additional context**
Add any other context about the problem here. The more information you provide, the easier it is to help.

bmadcode:
fixed

PR #1531, opened 2026-02-04
fix: remove --prefer-offline flag and update deprecated --production to --omit=dev

### LINKED ISSUE #1438: Installation fails with 'No matching version found' due to --prefer-offline flag

Drickon:
## What
Removes the `--prefer-offline` flag and replaces deprecated `--production` with `--omit=dev` in npm install commands in the module installer.

## Why
- **`--prefer-offline`** causes npm to use cached package metadata, which can be stale and fail to resolve recently published package versions (e.g., `diff@8.0.3`). This results in `ETARGET` errors during fresh installations.
- **`--production`** is deprecated by npm in favor of `--omit=dev`.

Fixes #1438

Replaces #1440 (closed due to merge conflicts from force-push; conflicts now resolved against current main).

## How
- Removed `--prefer-offline` from both npm install calls in `manager.js`
- Updated deprecated `--production` flag to `--omit=dev`
- Preserved `--legacy-peer-deps` from #1476

## Note on #1479
Issue #1479 consolidates multiple installer failures, but the `--prefer-offline` and `--production` flag fixes **have not yet been applied** to main. This PR provides the targeted fix for these specific npm flag issues.

## Testing
Verified that `npm install --omit=dev --no-audit --no-fund --no-progress --legacy-peer-deps` correctly installs dependencies without the stale cache issue.

### DIFF
--- a/tools/cli/installers/lib/modules/manager.js
+++ b/tools/cli/installers/lib/modules/manager.js
@@ -417,7 +417,7 @@ class ModuleManager {
       if (needsDependencyInstall || wasNewClone || nodeModulesMissing) {
         const installSpinner = ora(`Installing dependencies for ${moduleInfo.name}...`).start();
         try {
-          execSync('npm install --production --no-audit --no-fund --prefer-offline --no-progress --legacy-peer-deps', {
+          execSync('npm install --omit=dev --no-audit --no-fund --no-progress --legacy-peer-deps', {
             cwd: moduleCacheDir,
             stdio: 'pipe',
             timeout: 120_000, // 2 minute timeout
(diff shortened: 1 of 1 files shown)


bmadcode (approved):

bmadcode:
TY @Drickon

PR #1537, opened 2026-02-05
fix: trim activation header to avoid YAML formatting issues in kilo installer

### LINKED ISSUE #1503: 🐞 Bug Report: BMAD workflows / commands not working in Kilo Code IDE
Hi BMAD team,

I’m experiencing an issue using **BMAD Method** with the **Kilo Code IDE (kilo.ai)**.

While the **modes** (e.g. `Analyst`, `Architect`, `Dev`, etc.) load and work correctly (through the mode switcher and the /[mode-name] command) **none of the BMAD workflows or / commands seem to function** inside Kilo Code. Unfortunately not even the /bmad-help command works!

This makes it impossible to use BMAD as intended beyond switching modes.

---

### ✅ Expected Behavior
- BMAD workflows and commands should be recognized and executable in the IDE, similar to supported environments.
- Commands such as workflows or other BMAD actions should trigger the expected behavior.

---

### ❌ Actual Behavior
- Only the BMAD modes (Analyst, Architect, Dev, etc.) work.
- Workflows and commands are ignored or not recognized.
- No visible error message is shown in the IDE.

---

### 🧪 Environment
- **IDE:** Kilo Code (kilo.ai)
- **Kilo Version:** 5.2.2
- **BMAD Version:** 6.0.0 Beta 4
- **OS:** Windows 10 Enterprise

---

### ℹ️ Additional Notes
- This might be related to how Kilo Code handles command parsing, slash commands, or workflow definitions.
- If Kilo Code requires specific configuration steps for BMAD compatibility, documentation or guidance would be appreciated.

Thanks for your work on BMAD!

Image

Image

dracic:
Resolves #1503


dracic:
image

Missed that...

dracic:
I'm going to add kilo workflows too. Currently:

  - kilo.js only handles agents → creates custom modes in .kilocodemodes
  - No workflow support is implemented

dracic:
image

dracic:
image

bmadcode:
if kilo uses commands now and not that kilomodes file we probably do not need this separate kilo file? Will merge for now to fix the issue, but we should refactor this file away soon

dracic:
@bmadcode it uses both. .kilocodemodes for custom modes (dev, sm etc) and workflows for slash commands in chat window. Skills are here too, but knowing how huge and detailed the prompt is for each of the default modes in Cline-based tools, and as long as that remains the case, I don't think it's a good idea to remove custom modes. At least one "bmad" custom mode.

PR #1601, opened 2026-02-08
fix: enforce array type for agent principles in schema

### LINKED ISSUE #1594: Tighten agent schema: principles field allows both string and array
## Problem

The agent YAML schema defines `principles` as a union type:

```javascript
principles: z.union([
  createNonEmptyString('agent.persona.principles'),
  z.array(createNonEmptyString('agent.persona.principles[]'))
])
```

But the template (`agent.customize.template.yaml`) defines it as an array:

```yaml
principles: []
```

This mismatch allows contributors to use either format, causing ongoing inconsistency. PR #1593 normalizes existing files to arrays, but without schema enforcement the drift will recur.

## Suggested fix

Update the schema to array-only:

```javascript
principles: z.array(createNonEmptyString('agent.persona.principles[]'))
  .min(1, { message: 'agent.persona.principles must include at least one entry' }),
```

## Context

- PR #1593 normalizes 9 agent files from `principles: |` (string) to `principles:` (array)
- The `qa.agent.yaml` (added in v6.0.0-Beta.7) already used array format correctly
- All other agent files used block scalar strings, which passed validation due to the union

PabloLION:
## Summary

Tightens the agent YAML schema to enforce array-only for the `principles` field, matching the template definition.

Closes #1594

## Problem

The schema currently uses a union type that accepts both strings and arrays:

```javascript
principles: z.union([
  createNonEmptyString('agent.persona.principles'),
  z.array(createNonEmptyString('agent.persona.principles[]'))
    .min(1, { ... }),
])
```

But the template (`agent.customize.template.yaml`) defines `principles: []` — an array. This permissiveness allows contributors to use block scalar strings (`principles: |`) which parse as a single string instead of discrete items.

## Fix

- Remove the string option from the union, leaving only the array validator
- Add test fixture (`string-principles.agent.yaml`) that verifies string-format principles now correctly fails validation

### Before
```javascript
principles: z.union([
  createNonEmptyString('agent.persona.principles'),
  z.array(...).min(1, ...),
])
```

### After
```javascript
principles: z.array(createNonEmptyString('agent.persona.principles[]'))
  .min(1, { message: 'agent.persona.principles must include at least one entry' }),
```

## Dependencies

This PR should be merged **after** #1593 (which converts existing agent files from string to array format). Otherwise, the existing agent files will fail validation.

## Test plan

- [ ] Existing valid fixture (`complete-persona.agent.yaml`) still passes (uses array)
- [ ] New invalid fixture (`string-principles.agent.yaml`) correctly fails
- [ ] Existing invalid fixture (`empty-principles-array.agent.yaml`) still fails
- [ ] All agent YAML files pass validation (after #1593 is merged)

### DIFF
--- a/test/fixtures/agent-schema/invalid/persona/string-principles.agent.yaml
+++ b/test/fixtures/agent-schema/invalid/persona/string-principles.agent.yaml
@@ -1,0 +1,26 @@
+# Test: Principles as block scalar string instead of array
+# Expected: FAIL
+# Error path: agent.persona.principles
+# Error expected: array
+# Error received: string
+
+agent:
+  metadata:
+    id: string-principles
+    name: String Principles
+    title: String Principles
+    icon: ❌
+    hasSidecar: false
+
+  persona:
+    role: Test agent
+    identity: Test identity
+    communication_style: Test style
+    principles: |
+      - This is a string, not an array
+      - Each line is part of one string
+
+  menu:
+    - trigger: help
+      description: Show help
+      action: display_help
--- a/tools/schema/agent.js
+++ b/tools/schema/agent.js
@@ -241,12 +241,9 @@ function buildPersonaSchema() {
       role: createNonEmptyString('agent.persona.role'),
       identity: createNonEmptyString('agent.persona.identity'),
       communication_style: createNonEmptyString('agent.persona.communication_style'),
-      principles: z.union([
-        createNonEmptyString('agent.persona.principles'),
-        z
-          .array(createNonEmptyString('agent.persona.principles[]'))
-          .min(1, { message: 'agent.persona.principles must include at least one entry' }),
-      ]),
+      principles: z
+        .array(createNonEmptyString('agent.persona.principles[]'))
+        .min(1, { message: 'agent.persona.principles must include at least one entry' }),
     })
     .strict();
 }


PabloLION:
@coderabbitai Re: "All 9 production agent files currently fail the new schema — this PR breaks existing agents"

This is explicitly documented in the PR description under **Dependencies**:

> This PR should be merged **after** #1593 (which converts existing agent files from string to array format). Otherwise, the existing agent files will fail validation.

The merge order is intentional: #1593 migrates the data, #1601 tightens the schema. This is standard practice for schema migrations — data first, constraint second.

Re: the template using `principles: []` vs `.min(1)` — the template is a scaffolding starting point showing the field type. The `.min(1)` constraint exists to ensure agents in production have at least one principle. These serve different purposes and are not contradictory.

Re: the `startsWith('src/')` redundant guard — agreed it's pre-existing, out of scope for this PR.

PabloLION:
## Downstream impact: net positive

We maintain [bmad-plugin](https://github.com/PabloLION/bmad-plugin), a Claude Code plugin that wraps BMAD-METHOD. Our toolchain parses agent YAML files to generate plugin-compatible agent definitions.

When `qa.agent.yaml` was introduced in v6.0.0-Beta.7 with the correct array format for `principles`, while the 9 older agents still used `principles: |` (block scalar string), our `generate-agents` pipeline crashed:

```
TypeError: principles.startsWith is not a function
```

We've already adapted our code to match ([permalink](https://github.com/PabloLION/bmad-plugin/blob/2430777404df680cbd17d61ddc96619f1ec9d8e9/scripts/generate-agents.ts#L106-L108)). The result was a **net simplification** — our `formatPrinciples()` went from 12 lines of string-parsing heuristics (detecting `"- "` prefixes, splitting on periods, reconstructing lists) down to a single `.map()` call:

```diff
-function formatPrinciples(principles: string): string {
-  if (principles.startsWith('- ')) { return principles; }
-  const sentences = principles.split(/\.\s+/).map(...)...
-  return sentences.map((s) => `- ${s}`).join('\n');
-}
+function formatPrinciples(principles: string[]): string {
+  return principles.map((p) => `- ${p}`).join('\n');
+}
```

With proper YAML arrays, downstream consumers no longer need to guess whether principles is a string-encoded list or a real array. The schema enforcement in this PR would prevent regressions.

Since BMAD-METHOD is still in beta, this kind of type tightening seems like the right time to do it. We'd appreciate PRs #1593 + #1601 being included in the current cycle or the next patch.

bmadcode:
this is going to change with another PR, closing this one - thank you @PabloLION

PR #1614, opened 2026-02-09
fix: make installer prompt box compatible with current clack

dickymoore:
## What
Fixes a runtime crash during `bmad-cli install` when running with current `@clack/prompts` where `clack.box` is not exported, and adds a dedicated install smoke E2E test.

## Why
The crash blocks non-interactive installer E2E validation and prevents reliable end-to-end install testing.

## How
- Updated `tools/cli/lib/prompts.js` `box()` wrapper to gracefully degrade: `box` -> `note` -> `log.message` -> console fallback.
- Added `test/test-cli-install-e2e.sh` to run install in a temp project and assert expected BMAD layout is created.
- Added `npm run test:e2e:install` script in `package.json`.

## Testing
- `npm run test:e2e:install`
- `npm test`

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "rebundle": "node tools/cli/bundlers/bundle-web.js rebundle",
     "test": "npm run test:schemas && npm run test:refs && npm run test:install && npm run validate:schemas && npm run lint && npm run lint:md && npm run format:check",
     "test:coverage": "c8 --reporter=text --reporter=html npm run test:schemas",
+    "test:e2e:install": "bash test/test-cli-install-e2e.sh",
     "test:install": "node test/test-installation-components.js",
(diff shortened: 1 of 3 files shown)


dracic:
clack.box is exported and works correctly in the current and only stable v1 release. There is no version of @clack/prompts satisfying ^1.0.0 where box is  missing. How can I reproduce this?

dickymoore:
Thanks for flagging this @dracic. You’re right: my local environment had an out-of-spec dependency state (node_modules drift), and I treated that as a product bug. That was my mistake. In a clean install with the locked @clack/prompts v1 dependency, clack.box is available as expected. I’m closing this PR to avoid adding unnecessary compatibility code for an invalid setup. Appreciate you calling this out. Dicky

PR #1651, opened 2026-02-14
Fix: --custom-content flag and workflow config.yaml copying

### LINKED ISSUE #1623: [BUG] --custom-content CLI flag fails: missing sources/selectedFiles in config
## Description

The `--custom-content` CLI flag for non-interactive installation fails when used with custom modules. The installer throws:

```
Source for module 'pantheon' is not available. It will be retained but cannot be updated without its source files.
```

## Steps to reproduce

1. Have a custom module with a valid `module.yaml` (e.g., `~/git/pantheon/src/module.yaml`)
2. Run: `npx bmad-method install --directory . --modules bmm,pantheon --tools claude-code --custom-content ~/git/pantheon/src --user-name Test --yes`
3. Installation fails with the error above

## Expected behavior

The installer should recognize the custom content path and install the custom module successfully, the same way the interactive flow does.

## Actual behavior

The installer fails because `findModuleSource()` cannot find the custom module. The `customModulePaths` map is never populated because the installer checks for `config.customContent.sources` and `config.customContent.selectedFiles`, but the CLI code path only sets `paths` and `selectedModuleIds`.

### Root cause

In `tools/cli/lib/ui.js`, there are two code paths that build `customContentConfig` from CLI args (the "modify existing" flow ~line 338 and the "fresh install" flow ~line 479). Both produce:

```js
{ hasCustomContent: true, paths: [...], selectedModuleIds: [...] }
```

But the installer (`tools/cli/installers/lib/core/installer.js` ~lines 297-312) expects the config shape produced by the **interactive** path (`promptCustomContentSource()`):

```js
{ hasCustomContent: true, selected: true, sources: [{ path, id, name }], selectedFiles: [...], selectedModuleIds: [...] }
```

The missing `sources`, `selected`, and `selectedFiles` properties cause the installer to skip populating `customModulePaths`, so `findModuleSource()` returns null.

## Environment

- **BMad Version**: 6.0.0-Beta.8
- **AI IDE**: Claude Code
- **OS**: macOS

cidemaxio:
What
When using --custom-content, the installer now populates sources, selected, and selectedFiles in the custom content config so the installer can resolve custom module sources.
Why
The installer expects config.customContent.sources to build customModulePaths, but the --custom-content path only set paths and selectedModuleIds. That left customModulePaths empty and caused findModuleSource() to fail with "Source for module 'X' is not available."
Fixes #1623
How
Added a sources array with { path, id, name } for each custom module when building config from --custom-content
Set selected: true and selectedFiles (paths to each module.yaml) on the custom content config
Applied the same logic in both the modify flow and the new-install flow in tools/cli/lib/ui.js
Incorporated fixes from PR #1624: empty module.yaml handling (skip + warn) and removal of paths from the config to match promptCustomContentSource()
Fixed quick-update with --custom-content: pass through for re-caching, add defensive pathExists checks for cache entries, and fix workflow-level config.yaml copying in manager.js
Testing
Ran npx bmad-method install --directory . --modules bmm,bmb --custom-content /path/to/bmad-story-orchestration --tools claude-code,cursor --yes and confirmed the install completes and custom modules are installed. Re-running the same command (quick-update) also succeeds.

### DIFF
--- a/tools/cli/installers/lib/core/installer.js
+++ b/tools/cli/installers/lib/core/installer.js
@@ -527,28 +527,30 @@ class Installer {
             const cachedModules = await fs.readdir(cacheDir, { withFileTypes: true });
 
             for (const cachedModule of cachedModules) {
-              if (cachedModule.isDirectory()) {
-                const moduleId = cachedModule.name;
+              const moduleId = cachedModule.name;
+              const cachedPath = path.join(cacheDir, moduleId);
 
-                // Skip if we already have this module from manifest
-                if (customModulePaths.has(moduleId)) {
-                  continue;
-                }
+              // Skip if path doesn't exist (broken symlink, deleted dir) - avoids lstat ENOENT
+              if (!(await fs.pathExists(cachedPath)) || !cachedModule.isDirectory()) {
+                continue;
+              }
 
-                // Check if this is an external official module - skip cache for those
-                const isExternal = await this.moduleManager.isExternalModule(moduleId);
-                if (isExternal) {
-                  // External modules are handled via cloneExternalModule, not from cache
-                  continue;
-                }
+              // Skip if we already have this module from manifest
+              if (customModulePaths.has(moduleId)) {
+                continue;
+              }
 
-                const cachedPath = path.join(cacheDir, moduleId);
+              // Check if this is an external official module - skip cache for those
+              const isExternal = await this.moduleManager.isExternalModule(moduleId);
+              if (isExternal) {
+                // External modules are handled via cloneExternalModule, not from cache
+                continue;
+              }
 
-                // Check if this is actually a custom module (has module.yaml)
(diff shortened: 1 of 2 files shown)


dracic:
Is this duplicate of #1624 ?

cidemaxio:
quite possibly a duplicate.

dracic:
Better one PR too many than one too few.

cidemaxio:
I included the changes from the referenced PR.  I also made fixes so that update installs work.

- Empty module.yaml handling (skip + warn) — from #1624
- Removed paths from config to match promptCustomContentSource() — from #1624
- Pass --custom-content through for quick-update so installer can re-cache from source
- Add defensive pathExists checks for cache entries to avoid lstat ENOENT
- Fix workflow-level config.yaml copying in manager.js
- Fix customInfo.path resolution for _customModuleSources

Issue #1704, opened 2026-02-19
[BUG] Installed commands in Gemini CLI cannot find the agent files


rmzzz:
### Description

when starting, e.g. bmb agent in Gemini CLI, the paths cannot be located:
e.g.:
```
 > /bmad-agent-bmb-agent-builder                                                                                                                                
▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄
✦ I have completed the pre-flight checklist. However, I was unable to locate the configuration file at {project-root}/_bmad/bmb/config.yaml or the agent
  definition at {project-root}/_bmad/bmb/agents/agent-builder.md.


  To proceed, I require the correct paths to these files. Once I have them, I can load the configuration, internalize my agent definition, and properly greet
  you by name to begin our session.

```

The issue seems to be caused by the fact that Gemini does not understand the variable {project-root}.

Workaround:
change files references format manually, e.g.:
```
PRE-FLIGHT CHECKLIST:
1.  [ ] IMMEDIATE ACTION: Load and parse @{_bmad/bmb/config.yaml} - store ALL config values in memory for use throughout the session.
2.  [ ] IMMEDIATE ACTION: Read and internalize the full agent definition at @{_bmad/bmb/agents/module-builder.md}.
3.  [ ] CONFIRM: The user's name from config is {user_name}.

``` 

### Steps to reproduce

1. Run 'npx bmad-method install'
2. Select option "Gemini"
3. Run "gemini"
4. Try to start an agent using slash command
5. See the error

### Expected behavior

The agent files should be loaded correctly

### Actual behavior

Gemini fails to load agent files

### Which module is this for?

BMad Method (BMM) - Core Framework

### BMad Version

6.0.1

[...]

PR #1706, opened 2026-02-19
fix: file references in Gemini commands templates

rmzzz:
## What
fixed files refs in Gemini commands
issue: https://github.com/bmad-code-org/BMAD-METHOD/issues/1704

## Why
because Gemini CLI was not able to load agent files, i.e. could not recognize {project-root} variable.

## How
- changed file refs syntax according to https://geminicli.com/docs/cli/custom-commands/

## Testing
- exec `npm run bmad:install`
- start `gemini`
- try any agent, e.g. `/bmad-agent-bmm-dev`
- check that agent is successfully loaded and agent menu is shown

### DIFF
--- a/tools/cli/installers/lib/ide/templates/combined/gemini-agent.toml
+++ b/tools/cli/installers/lib/ide/templates/combined/gemini-agent.toml
@@ -3,12 +3,12 @@ prompt = """
 CRITICAL: You are now the BMad '{{name}}' agent.
 
 PRE-FLIGHT CHECKLIST:
-1.  [ ] IMMEDIATE ACTION: Load and parse {project-root}/{{bmadFolderName}}/{{module}}/config.yaml - store ALL config values in memory for use throughout the session.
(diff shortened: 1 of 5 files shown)


dracic:
We had file injection before, somehow it was kicked out from Gemini CLI custom commands definition. But if we go back let's keep in mind that: _You can directly embed the content of a file or a directory listing into your prompt using the @{...} syntax._ 

So if the content of a config.yaml is:

`This is my prompt that i'm going to add to my agent definition`

And we have:

`1.  [ ] IMMEDIATE ACTION: Load and parse @{{{bmadFolderName}}/{{module}}/config.yaml} - store ALL config values in memory for use throughout the session.`

The final prompt would be:

`1.  [ ] IMMEDIATE ACTION: Load and parse This is my prompt that i'm going to add to my agent definition - store ALL config values in memory for use throughout the session.`

Practically the whole content of a prompt, except of injecting using the agent file name would be redundant. This notation is not for file references, it is for embedding file content.

rmzzz:
the issue cannot be reproduced anymore, even without any fix -> closing

Issue #1718, opened 2026-02-20
[Master Plan] Standardize and fully enforce file reference conventions


alexeyv:
## Context

BMAD's cross-file reference system has grown organically into multiple inconsistent formats with incomplete validator coverage and until recently no CI enforcement. This issue is the master plan for fixing it completely. It supersedes #1574, #1615, and #1626.

---

## Two workflow formats — different rules apply

BMAD contains two distinct workflow formats with different path conventions. It is important not to conflate them.

### Old format (XML-based)

Files: `workflow.yaml` + `instructions.xml`, executed via `core/tasks/workflow.xml`.
Used by: `create-story`, `dev-story`, `code-review` in `4-implementation/`.

This format uses `{installed_path}` extensively. That token is valid because it is always defined by the calling chain through `workflow.yaml` before any `instructions.xml` is reached. This is a closed, self-consistent system with its own conventions. **The path reference requirements in this issue do not apply to old-format workflows.**

### New format (MD/steps-based)

Files: `workflow.md` + `steps/step-NN.md`.
Used by: all BMM workflows (`create-architecture`, `check-implementation-readiness`, `create-prd`, `quick-dev`, `quick-spec`, `create-ux-design`, `create-epics-and-stories`, etc.).

This is the active format and the subject of this issue. Its path conventions need to be standardized and enforced.

---

## The core requirement (new format only)

`{project-root}` is the accepted starting point. It works in practice across all supported LLM platforms and there is no better alternative.

**After installation, no path should exist anywhere in any new-format installed file that is relative or platform-absolute. Every path must be expressed as `{project-root}/_bmad/...`.**

### Why the topology makes this tractable

New-format workflows have a strictly linear, single-entry calling chain:

```
entry point → workflow.md → step-01.md → step-02.md → ...
```

There is no way to reach a step file without having passed through `workflow.md` first. `workflow.md` defines `{installed_path}` as:

```
installed_path = {project-root}/_bmad/bmm/workflows/.../this-workflow
```

So `{installed_path}` is always available in context for step files — but it is defined in terms of `{project-root}`, which means any path using `{installed_path}` is transitively traceable back to `{project-root}`. This is valid.

What is NOT valid: relative paths (`./`, `../`, bare filenames, bare `_bmad/`). These require the LLM to track its current file location, which is a harder inference with no formal grounding.

### Core tasks

[...]

PR #1720, opened 2026-02-20
feat: add LLM audit prompt for file reference conventions

alexeyv:
## Summary

- Add `tools/audit-file-refs.md` — a repeatable prompt that audits new-format BMAD source files for non-conforming file references using parallel Haiku subagents
- Add `.junie/` to `.prettierignore` (also covered by #1719; whichever merges second will be a no-op)

## What this replaces

Item 1 of the master plan (#1718) originally called for extending `validate-file-refs.js` with additional regex patterns to cover ~110 currently invisible references.

That approach was superseded. Regex pattern enumeration cannot reliably determine whether a given text fragment is a file reference that violates the convention — that requires semantic understanding of context. The regex CI validator stays as a fast enforcement layer for patterns it already knows; this prompt provides the broader coverage.

## How it works

1. Runs `find src/bmm src/core src/utility` to get the file list (excludes `4-implementation/` old-format files)
2. Batches files into groups of ~20
3. Spawns parallel Haiku subagents — each reads its batch and identifies references that do not conform to the `{project-root}/_bmad/...` convention
4. Self-check: sums `FILES CHECKED: N` from all subagents and verifies against the total before reporting
5. Outputs findings grouped by violation type

## Usage

Ask Claude: *"Read and follow tools/audit-file-refs.md"*

Refs #1718

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ cursor
 CLAUDE.local.md
 .serena/
 .claude/settings.local.json
+.junie/
 
 z*/
 
@@ -45,6 +46,7 @@ _bmad-output
 # .augment/ is gitignored except tracked config files — add exceptions explicitly
 .augment/*
 !.augment/code_review_guidelines.yaml
+.codebuddy
 .crush
 .cursor
 .iflow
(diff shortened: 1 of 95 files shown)


bmadcode:
Can this be a .claude/skill instead in the repo similar to some of the others? (eventually we will move skills over to a maintainer repo)

PR #1790, opened 2026-02-28
feat(core): add edge case hunter review task

alexeyv:
## What

New reusable review task (`review-edge-case-hunter.xml`) that exhaustively traces branching paths and boundary conditions in code, reporting only unhandled gaps. Registered in `module-help.csv` as `bmad-review-edge-case-hunter`.

## Why

The existing adversarial review is attitude-driven (cynical persona). Pairing it with a method-driven reviewer that mechanically walks every path produces orthogonal findings with minimal overlap. This second perspective is proven useful but had no dedicated task definition.

## How

- Created `src/core/tasks/review-edge-case-hunter.xml` following the same XML task structure as the adversarial review
- Diff-scoped analysis: scans only diff hunks when a diff is provided, full content otherwise
- Structured JSON output with four fields: location, trigger_condition, guard_snippet, potential_consequence
- No persona — the forcing function is the exhaustive path-walk method, not an attitude
- Added registration row in `src/core/module-help.csv`

## Testing

- All existing tests pass (`npm test`)
- File reference validation confirms the new CSV row points to a valid file


alexeyv:
## Review Response

Thanks for the thorough review. Addressed 4 findings, dismissed the rest with rationale.

### Fixed (commit 5103be4)

1. **Output format array example** (line 17) — Now shows `[{}]` wrapper and documents `[]` as valid output.
2. **Zero findings halt contradiction** (line 60) — Changed from ambiguous "HALT or re-analyze" to deterministic: re-analyze once, then return `[]`. This is critical for subagent consumers that expect JSON.
3. **Empty content handling duplication** (lines 42 vs 61) — Step 1 now defers to halt-conditions instead of defining separate behavior.
4. **"Execute ALL steps" vs halt-conditions** (line 27) — Softened to "unless a halt-condition triggers."

### Dismissed

- **`_bmad/` underscore prefix** — Project convention. ALL installed file IDs use this prefix (`_bmad/core/tasks/...`, `_bmad/bmm/workflows/...`). It's the installation target path, not a private/public indicator.
- **Scope rule false positives** — By design. Diff-scoped analysis is the purpose of this task. The `also_consider` input exists for additional context when needed. The downstream consumer (PR review workflow) already knows findings are diff-scoped.
- **Missing language/content_type inputs** — LLMs reliably detect language from diff headers and code syntax. Adding required inputs adds invocation friction for no measurable accuracy gain.
- **Output size constraints/severity field** — The consumer (PR review skill) handles severity mapping and finding limits. Adding them here would duplicate logic and change the contract.
- **"Mechanically walk" is aspirational** — This is an LLM prompt instruction, not a product claim. Strong directive language produces better coverage than hedged instructions with disclaimers.
- **Boundary list incomplete** — Current list covers the most common static-analysis cases. Adding BOM/locale/symlinks/NFC is diminishing returns for a prompt instruction. Concurrency/timing are valid to flag even if imperfect.
- **Deduplication/grouping** — The consumer handles dedup (merge step in PR review workflow). Adding it within the task creates redundant logic.
- **"Within provided scope" clarification** — The objective already states "scan only the diff hunks." The description is accurate for its context.
- **also_consider format** — It's optional freeform prose by design. Over-specifying format adds friction for a field that's rarely used.

PR #1792, opened 2026-02-28
feat: add Kimi Code CLI support

lasting-yang:
add Kimi Code CLI support


bmadcode:
@lasting-yang branch has conflicts and code rabbit comments that can be addressed.

alexeyv:
We're switching BMAD to "everything is skills", so if Kimi supports skills, please look at how the installer is now (it's really trivial) and submit a PR that makes it part of the new setup. 

If not, let's wait for a week or two until they do - seeing how everybody supports skills now, it shouldn't be a much longer wait.

PR #1793, opened 2026-02-28
fix(workflow): use git for dev-story File List

### LINKED ISSUE #1789: [BUG] dev-story workflow produces systematically incomplete File Lists — infrastructure side-effects never captured
### Description

_I had BMAD do this analysis and write this up itself._   

Component: bmm module — dev-story workflow                                                                                  
  Affected file: _bmad/bmm/workflows/4-implementation/dev-story/instructions.xml
  BMAD version: 6.0.3 (installed 2026-02-25)
  Severity: Medium — causes a recurring MEDIUM finding in every code-review run

  ---
  Description

  The dev-story workflow produces a File List in the Dev Agent Record that reliably omits infrastructure files that change as  predictable side-effects of story implementation. The same class of omission has occurred in every story implemented so far in this project (Stories 1-1, 1-2, 1-3).

  The omitted files are always the same:
  - pnpm-lock.yaml — updated whenever any package is installed or upgraded
  - _bmad-output/implementation-artifacts/<story-key>.md — the story file itself is modified throughout implementation but  never self-listed
  - _bmad-output/implementation-artifacts/sprint-status.yaml — updated by the workflow at start (→ in-progress) and completion
   (→ review)

  These files are genuinely changed in git. The code-review workflow correctly detects them as "files changed but not in story  File List" and raises a MEDIUM finding every time. This is wasted review signal — reviewers have to acknowledge a known-false positive on every single story.

  ---
  Root Cause

  The problem has two parts:

  1. The File List instruction is per-task and agent-driven (Step 8, line 299)

  <action>Update File List section with ALL new, modified, or deleted files (paths relative to repo root)</action>

  This fires after each task completes. The agent updates the File List based on what it believes it changed during that task — which is the application source files it just wrote. It has no awareness that pnpm install (run as a subtask) also mutated   pnpm-lock.yaml.

  2. The Step 9 completeness check is also agent-driven (lines 326, 339, 369)

  <action>Confirm File List includes every changed file</action>
  ...
  - File List includes every new/modified/deleted file (relative paths)
  ...
  <action if="File List is incomplete">HALT - Update File List with all changed files</action>

  The agent self-evaluates "is the File List complete?" against its own mental model of what it changed — which already  excludes the infrastructure files. The HALT gate fires only if the agent knows something is missing. Since it never knew  about pnpm-lock.yaml, it doesn't HALT. The check passes vacuously.

  3. The story template provides no guidance (template.md, line 49)

  ### File List

  The section is a bare heading with no examples, no hints, and no reminder about infrastructure side-effects. The agent has nothing to anchor on beyond the generic instruction in Step 8.

  ---
  Proposed Fix

  Replace the agent's self-assessment in Step 8 and Step 9 with ground-truth git output. After each task and at story  completion, run git status --porcelain and git diff --name-only to get the actual set of changed files, then cross-reference  against the current File List and append any missing entries.

  Step 8 change — after the current File List update action, add:

  <action>Run `git status --porcelain` to get the actual set of changed files on disk</action>
  <action>Run `git diff --name-only` to get unstaged modified files</action>
  <action>Run `git diff --cached --name-only` to get staged files</action>
  <action>Compare git-reported changed files against current File List content</action>
  <action>Append any files present in git output but absent from File List (exclude _bmad/ and _bmad-output/
  directories)</action>

  Step 9 change — replace the current text-instruction check with the same git-driven verification, and make the HALT gate use git output as the source of truth rather than agent recall:

  <action>Run `git status --porcelain` and `git diff --name-only` to enumerate ALL changed files</action>
  <action>Cross-reference against File List — any git-reported file not in File List is an omission</action>
  <action if="git reports files not in File List">Append missing files to File List before continuing</action>

  This eliminates the class of problem entirely — the File List is now derived from git reality rather than agent memory, so infrastructure side-effects (lockfile, sprint tracker, story file) are captured automatically.

  ---
  Exclusion rule

  The code-review instructions already exclude _bmad/ and _bmad-output/ from code review (instructions.xml line 15). The same exclusion should apply to the File List population — the story file and sprint tracker are workflow artifacts, not application source. The fix should exclude _bmad-output/ from the git cross-reference so they don't clutter the File List. Only pnpm-lock.yaml and similar repo-root infrastructure files should be added.

### Steps to reproduce
_Given a story that requires installing new dependencies._

1. /bmad-bmm-create-story
2. /bmad-bmm-dev-story
3. /bmad-bmm-code-review

### Expected behavior

No code review issue reported for files missing from story File List.

### Actual behavior

Code review reports:

>   **M3 — Git-undocumented changes in story File List**
>   The story File List omits three files that have real git changes: pnpm-lock.yaml (modified by pnpm install), the story file  itself, and sprint-status.yaml. While these are workflow artifacts, incomplete File Lists make future audit difficult. Both  pnpm-lock.yaml and the sprint tracker should appear under "Modified files."
> 

### Screenshots

_No response_

### Which module is this for?

### BMad Version

6.0.3

Claude Code

### Operating System

Linux

### Relevant log output

```shell

```

### Confirm

A386official:
## Summary

The dev-story workflow builds its File List from agent memory, which misses infrastructure side-effects like `pnpm-lock.yaml`, the story file itself, and `sprint-status.yaml`. This causes a false-positive MEDIUM finding on every code-review run.

The fix replaces agent-driven File List updates (Steps 8 and 9) with `git diff --name-only` / `git status --porcelain` - the same approach the code-review workflow already uses for change discovery.

Also adds a hint comment to the story template's File List section so the pattern is visible to agents from the start.

Closes #1789

## Test plan

- Run the dev-story workflow on a story that installs a package or modifies sprint-status.yaml
- Verify the File List in the completed story includes lock files, story file, and sprint-status.yaml
- Run code-review afterward and confirm no false-positive "files changed but not in File List" finding

### DIFF
--- a/src/bmm/workflows/4-implementation/create-story/template.md
+++ b/src/bmm/workflows/4-implementation/create-story/template.md
@@ -48,5 +48,13 @@ so that {{benefit}}.
 
 ### File List
 
-<!-- Use `git diff --name-only` and `git status --porcelain` to capture ALL changed files,
-     including infrastructure side-effects: lock files, this story file, sprint-status.yaml -->
+<!-- 
+  Populate using git commands after each task:
+    git diff --name-only
+    git diff --cached --name-only
+    git status --porcelain
+  Include infrastructure side-effects (lock files, sprint-status.yaml, story file).
+  Exclude: _bmad/, _bmad-output/ directories.
+-->
+
+- (list files here, one per line)
--- a/src/bmm/workflows/4-implementation/dev-story/instructions.xml
+++ b/src/bmm/workflows/4-implementation/dev-story/instructions.xml
@@ -296,10 +296,12 @@
     <!-- ONLY MARK COMPLETE IF ALL VALIDATION PASS -->
     <check if="ALL validation gates pass AND tests ACTUALLY exist and pass">
       <action>ONLY THEN mark the task (and subtasks) checkbox with [x]</action>
-      <action>Run `git diff --name-only` and `git status --porcelain` to discover ALL changed files (including infrastructure
-        side-effects like lock files, the story file itself, and sprint-status.yaml)</action>
-      <action>Update File List section from git output with ALL new, modified, or deleted files (paths relative to repo root)</action>
       <action>Add completion notes to Dev Agent Record summarizing what was ACTUALLY implemented and tested</action>
+      <action>Save the story file NOW so its own changes appear in the working tree</action>
+      <action>Run `git diff --name-only`, `git diff --cached --name-only`, and `git status --porcelain` to discover ALL changed files
+        (including infrastructure side-effects like lock files, the story file itself, and sprint-status.yaml)</action>
+      <action>Filter out paths under `_bmad/` and `_bmad-output/` directories from the git output</action>
+      <action>REPLACE the File List section entirely with the filtered results: ALL new, modified, or deleted files (paths relative to repo root, one per line)</action>
     </check>
 
     <check if="ANY validation fails">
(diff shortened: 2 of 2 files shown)


A386official:
Addressed in the latest push — thanks for the thorough review. Changes made:

1. **Save-ordering fix (main bug):** Moved `Save the story file` to *before* the git inspection commands in Step 8, so the story file's own changes are visible in the working tree when `git diff` / `git status` run. Also added the same save-before-inspect pattern in Step 9.

2. **Added `git diff --cached --name-only`:** All three git inspection sites (Step 8, Step 9, DoD checklist) now use the full three-command set: `git diff --name-only`, `git diff --cached --name-only`, and `git status --porcelain`.

3. **`_bmad/` and `_bmad-output/` exclusion:** Added explicit filtering instructions to exclude these directories from the File List in both `instructions.xml` and `template.md`.

4. **Replace semantics:** Clarified that the File List update is a full replace (not append), avoiding duplicates.

5. **Template improvements:** Restructured the HTML comment in `template.md` to list all three git commands clearly, added exclusion guidance, and added a placeholder list structure (`- (list files here, one per line)`).

PR #1800, opened 2026-03-01
feat: copilot installer frontmatter improvements

### LINKED ISSUE #1794: Update GitHub Copilot Installer to remove Tool declaration, add Agent names, and Agent assignments for Prompts
**Describe your idea**
For both Agents and Prompts, it's not necessary to specify Tools, this can be removed.
For Agents, adding a name lets you keep file names descriptive, but activations clean
For Prompts, the same applies as Agents for names
For Prompts, specifying an Agent will automatically select that agent in the chat for followup message consistency

**Why is this needed?**
Users may want to give agents their own tooling, including MCP tools, etc, and also, the built in tools change. Removing the tools section allows users to specify what tools they want agents to use without having to individually modify every single agent and prompt file to include them. This is tedious, and breaks on update. Most users give all agents the same tooling anyway, and usually not the small limited set that the installer currently specifies. GHCopilot is also the only installer to even do this.

Adding names to Agents lets users select them when asking an agent to execute a task inside GitHub web interface directly, it also allows for shorter, cleaner names. Current names that include the full bmad + module are quite "dirty" in use and hard to read. The PM agent should simply be called PM. The same applies for Prompts to activate through `/` commands.

Prompts can specify a specific agent by name so that when it's run, it will switch to that agent in the selector. This means if you send followup questions, the agent's persona and context are already loaded and consistent. Setting the agent to "agent" sets it back to default every time, deselecting any selected persona. For workflows that have more than one primary agent (e.g. Code Review is used by dev and quick flow solo dev), it may be best to set it back to agent, or not include an "agent" line in the frontmatter, leaving whatever agent was previously or manually selected as active. I will go with this approach.

**How should it work?**
Modification to the github installer should be straight forward in most cases, but the last one - skipping adding the agent to frontmatter for workflows with multiple activating agents - will be a little more complicated. Not impossible though.

**PR**
If you'd like to contribute, please indicate you're working on this or link to your PR. Please review [CONTRIBUTING.md](../../CONTRIBUTING.md) — contributions are always welcome!

I plan on forking and implementing this myself soon.

**Additional context**
Add any other context, screenshots, or links that help explain your idea.

skittlz444:
What
Updated the GitHub Copilot installer to remove tool declarations from frontmatter, add clean agent names, resolve prompt filename collisions, and set correct agent assignments in prompt files.

Why
Tools in frontmatter forced users to manually edit every agent/prompt file to customize tooling, and broke on update. Agent names used persona display names (e.g., "Mary") instead of clean identifiers (e.g., "dev"), making @mentions and /commands
  hard to read. Create Story and Validate Story shared the same command slug, causing one to silently overwrite the other. Fixes #1794

How
- Removed tools from all agent and prompt frontmatter, along with the now-unused getToolsForFile() and collectExistingToolPermissions() methods
- Changed agent .agent.md name field and all prompt agent: references to use the raw agent name (e.g., "dev", "pm") instead of persona display names
- Added duplicate command detection in generatePromptFiles() — when multiple CSV entries share a command, unique filenames are derived from the entry name slug, and workflow options (Create Mode, Validate Mode) are passed in the prompt body

Testing
Ran the full test suite (test:schemas, test:refs, test:install, validate:schemas) — all 71 tests pass. Verified ESLint and Prettier both report clean on the modified file.

fixes #1794

### DIFF
--- a/tools/cli/installers/lib/ide/github-copilot.js
+++ b/tools/cli/installers/lib/ide/github-copilot.js
@@ -143,6 +143,9 @@ class GitHubCopilotSetup extends BaseIdeSetup {
    * @returns {string} Agent file content
    */
   createAgentContent(artifact, manifestEntry) {
+    if (!artifact?.name) {
+      throw new Error('Agent artifact must have a name');
+    }
     // Build enriched description from manifest metadata
     let description;
     // Use the raw agent name (e.g., "dev", "pm") for clean @mention selection
@@ -202,6 +205,7 @@ You must fully embody this agent's persona and follow all activation instruction
         if (!entry.command || !entry['workflow-file']) continue;
         commandCounts.set(entry.command, (commandCounts.get(entry.command) || 0) + 1);
       }
+      const seenSlugs = new Set();
 
       for (const entry of helpEntries) {
         const command = entry.command;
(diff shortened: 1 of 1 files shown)


alexeyv:
Thank you! We are switching everything to skills, including Copilot. I'd love to hear from an actual Copilot user that the new setup works well enough, but this PR I have to therefore close.

Don't hesitate to ping Brian, myself or anybody else from the core team on the Discord before submitting another PR to minimize the chances of it suffering the same fate.

PR #1829, opened 2026-03-06
mv instructions to SKILL.md and add license: MIT

steftux:
## What
moved contents of instructions.md to SKILL.md

## Why
see discussion https://github.com/bmad-code-org/BMAD-METHOD/discussions/1799 

## How
copy paste

## Testing
none needed


bmadcode:
good change here - I am about to move these maintainer tools to a separate repo to make them available as a plugin to all repos

steftux:
> Thanks, that's a good idea. Before I can approve this PR, can you please remove the `license: MIT` from the front matter?
> 
> No reason to pollute every session's context with this irrelevant information.

Hi @alexeyv thanks for your comment. If you worry about redundancy you might instead specify a text like "LICENSE.txt has complete terms" however that text is even longer than just stating "MIT", so I would prefer the shorter version. Regarding your point about the relevance: I do not think it is irrelevant information as the user might be interested in the terms of use. And it is a valid field as specified in https://agentskills.io/specification#license-field , so if it is actually MIT license I would rather keep it than leaving the user in doubt. Could I convince you or not yet? Thanks for a reply, Stefan

bmadcode:
it does seem to be a standard I see across all anthropic skills. Some other skills I see instead keep a LICENSE.md file at the root of the skill folder.

But again since these are about to be moved to  a separate repo - I am going to close this repo. But with all the other files becoming skills soon, I will decide on a standard for them soon and publish it in the project standards guide for skill standards.

Thanks for raising the topic @steftux

PR #1890, opened 2026-03-11
feat(codex): install BMAD as Codex custom prompts

LINGREN201:
## How

-


alexeyv:
We've been there, wasted a lot of time maintaining the installer. From now on it's all skills all the way.

PR #1897, opened 2026-03-11
feat: add platform capability flags for conditional workflow execution

### LINKED ISSUE #1898: feat: platform capability flags for conditional workflow behavior
### What

Add a capabilities system to `platform-codes.yaml` so BMAD workflows can adapt behavior based on what the IDE/CLI platform supports — specifically sub-agent delegation, task tracking, and structured user prompting.

### Why it would benefit the BMad community

BMAD supports 20+ IDE platforms, but implementation workflows (quick-dev, dev-story) currently have no way to detect and leverage advanced platform features. This means:

- Platforms with rich tool support (like Claude Code's Agent tool, TaskCreate/TaskUpdate) can't be utilized by workflows
- Adding platform-specific instructions to shared workflows would break other platforms
- Users on capable platforms miss out on orchestrator-pattern delegation and structured progress tracking

### How it strengthens human-AI collaboration

- **Capability-aware workflows** automatically use the best available tools — agent delegation when sub-agents exist, direct implementation when they don't
- **Task tracking integration** gives users structured visibility into implementation progress on platforms that support it
- **Extensible by design** — as other IDE platforms add capabilities, the community can simply flip flags in `platform-codes.yaml` without changing workflow logic

### Implementation

PR: #1897

Dynokostya:
## What

Add a platform capabilities system that allows BMAD workflows to adapt behavior based on what the IDE/CLI tool supports — specifically sub-agent delegation, task tracking, and structured user prompting.

## Why

BMAD supports 20+ IDE platforms, but implementation workflows (quick-dev, dev-story) previously had no way to leverage advanced platform features like sub-agent delegation or task tracking. This meant workflows either used platform-specific features (breaking other platforms) or avoided them entirely (leaving capabilities on the table).

This was targeted mainly for Claude Code which currently has the richest tool support. Other coding tools should explore adding `capabilities` support as their platforms evolve — the infrastructure is ready, just flip the flags in `platform-codes.yaml`.

Fixes #1898

## How

- Added `capabilities` schema to `platform-codes.yaml` with three flags per platform: `sub_agents`, `task_tracking`, `structured_ask`
- Currently only Claude Code has all three enabled; all others default to `false` (conservative — expand as verified)
- Installer persists capabilities to `_config/ides/{platform}.yaml` at install time via `ide-config-manager.js`
- Updated `quick-dev/step-03-execute.md`: loads capabilities, conditionally registers tasks, uses orchestrator pattern (agent delegation) when `sub_agents` available, falls back to direct implementation
- Updated `dev-story/workflow.md`: step 4 conditionally registers tasks, step 5 conditionally delegates to agents or implements directly with red-green-refactor

## Testing

- All 215 installation component tests pass
- Schema validation: 10/10 agents pass
- File reference validation: 498 references, 0 broken
- Lint + format: clean


Dynokostya:
Worth noting: most BMAD users run Claude Code since it has the strongest coding models, so this feature will be immediately beneficial for the majority of the user base. The capability flags infrastructure is ready for other platforms to opt in as they add similar tooling — just a one-line change in `platform-codes.yaml` per capability.

Dynokostya:
**Re: agent discovery path hardcoded to `.claude/agents/`** — Fixed in 5b2003de. Both workflow files now use platform-agnostic phrasing with `.claude/agents/` only as an example.

**Re: `loadAllIdeConfigs` doesn't return capabilities** — Valid observation, but changing the return shape would be a breaking change. All 4 callers in `installer.js` use `savedIdeConfigs[ide]` directly as the raw configuration object (e.g., `ideConfigurations[ide] = savedIdeConfigs[ide]`). Wrapping it in `{ configuration, capabilities }` would require updating all call sites.

In practice there's no gap: workflows read the YAML file directly at runtime and get the full object including `capabilities`. The `loadAllIdeConfigs` path is only used by the installer for re-applying saved configs during updates. If a future feature needs programmatic access to all capabilities, that refactor can be scoped separately.

bmadcode:
@Dynokostya I like this idea, not sure we want to maintain it though - I have a skill coming soon called bmad-init or something similar that every workflow will call at the start, and pulling in tool capabilities also at that time is a great idea, then as we are converting to skills there will be common language or phrasing that can exist and progressive disclosure to different prompts based on capabilities.

bmadcode:
Hi @Dynokostya, thank you for this PR and the careful design behind it. The capability-flags concept is real, but the maintainer direction (per the discussion on this PR) is to handle this through a bmad-init skill that pulls platform capabilities at workflow start, rather than per-platform YAML capability files. Closing this PR as superseded by the planned approach. Issue #1898 remains open for the underlying need. Really appreciate the effort you put in here.

PR #1907, opened 2026-03-12
fix: ensure --yes flag assigns defaults for all config fields

### LINKED ISSUE #1803: CLI install --yes flag does not suppress new config prompts
## Description

When running bmad-cli install --yes (or --action quick-update --yes), newly added configuration prompts still present interactive TUI prompts (via clack), blocking headless/CI execution.

## Steps to Reproduce

1. Have an existing BMAD installation with bmm module
2. Run the installer with --yes after a version that introduces new config questions (e.g., Subagents supported? and Agent Teams supported?):
   ```
   node tools/cli/bmad-cli.js install --directory /path/to/project --modules bmm --tools claude-code --yes
   ```
3. The CLI enters quick-update mode as expected, but then presents interactive prompts for the new config options

## Expected Behavior

--yes should auto-accept defaults for all prompts, including newly added config questions. The flag enables fully non-interactive operation.

## Actual Behavior

The CLI blocks on interactive prompts. In non-TTY environments (piped stdin, CI runners, Claude Code subagents), the process hangs indefinitely because the TUI library cannot read input.

## Environment

- BMAD v6.0.4
- macOS / Node.js
- Non-interactive terminal (stdin piped or not a TTY)

nikolasdehor:
## Summary

Fixes #1803

When running `bmad install --yes`, the config collector skips interactive prompts but only assigns values for fields that have explicit defaults. Fields without defaults are silently skipped, leaving `undefined` values that can cause downstream template rendering failures (e.g., `{{name}}` placeholders left unresolved).

This change ensures that in `skipPrompts` mode, fields without defaults are assigned an empty string fallback, preventing undefined values from propagating through the installer pipeline.

## Changes

- `tools/cli/installers/lib/core/config-collector.js`: Updated `skipPrompts` branch to assign empty string for fields without explicit defaults
- Early-continue for function-type defaults (computed defaults that require prior answers)

## Test plan

- [ ] Run `npx bmad-method install --yes` and verify no `undefined` values appear in generated files
- [ ] Run `npx bmad-method install --yes --directory ./test-output` and inspect output config
- [ ] Verify interactive mode (`npx bmad-method install`) still works as before

### DIFF
--- a/tools/cli/installers/lib/core/config-collector.js
+++ b/tools/cli/installers/lib/core/config-collector.js
@@ -735,12 +735,13 @@ class ConfigCollector {
       // Skip prompts mode: use all defaults without asking
       if (this.skipPrompts) {
         await prompts.log.info(`Using default configuration for ${moduleDisplayName}`);
-        // Use defaults for all questions
+        // Use defaults for all questions; use empty string for fields without defaults
         for (const question of questions) {
-          const hasDefault = question.default !== undefined && question.default !== null && question.default !== '';
-          if (hasDefault && typeof question.default !== 'function') {
-            allAnswers[question.name] = question.default;
+          if (typeof question.default === 'function') {
+            continue;
           }
+          const hasDefault = question.default !== undefined && question.default !== null && question.default !== '';
+          allAnswers[question.name] = hasDefault ? question.default : '';
         }
       } else {
         if (!this._silentConfig) await prompts.log.step(`Configuring ${moduleDisplayName}`);


bmadcode:
Hi @nikolasdehor, thanks for the fix. The target file (config-collector.js) no longer exists, the installer was rewritten and --yes handling is now substantially more comprehensive. Closing as stale. If you still see config-default issues with current main, please open a fresh issue. Appreciate the contribution.

PR #1914, opened 2026-03-12
fix: add brainstorming reconciliation to prevent silent idea loss in PRD

### LINKED ISSUE #1674: PRD workflow silently drops brainstorming ideas — needs reconciliation step
## Problem

When feeding a brainstorming session into the PRD workflow (along with research documents), the PRD silently drops a significant number of brainstorming ideas. The pattern is consistent and non-random: concrete/structural ideas survive, but "what should this feel like" ideas get lost — tone, coaching philosophy, interaction design, personality, UX feel.

Examples of ideas that were silently dropped:
- "The agent speaks in physics by default" (tone)
- "Assumed ignorance, no shame" (personality)
- "Science serves diagnostics, not education" (design philosophy)
- "Glanceable screen layout — instrument panel, not chat log" (interaction design)

The resulting PRD reads as complete and coherent, so you'd never know things were missing without manually diffing it against the brainstorming output. This is a silent failure mode.

## Possible causes

- **Lost in the middle** — brainstorming docs are long, and soft/qualitative ideas may not survive the context pressure alongside structured research documents
- The **doc import step** in the PRD workflow isn't specific enough about what to look for regarding relevance — it may be implicitly biased toward concrete, implementable requirements
- Soft ideas don't map cleanly to the PRD template's structure (functional requirements, non-functional requirements, etc.), so they fall through the cracks

## Possible mitigations

1. **Explicit reconciliation step at the end of the PRD workflow** — cross-check brainstorming output against finished PRD, flag ideas that didn't land anywhere
2. **Brainstorming → Brief → PRD pipeline** — routing through a brief first may force more brainstorming ideas to be captured before the PRD's structural template takes over
3. **Improve doc import relevance guidance** — make the PRD workflow more specific about what to extract from brainstorming docs, explicitly calling out tone/philosophy/interaction design as first-class concerns alongside functional requirements

nikolasdehor:
## Summary

Fixes #1674

When feeding brainstorming sessions into the PRD workflow, soft/qualitative ideas (tone, personality, interaction design, coaching philosophy) get silently dropped because they don't map cleanly to the PRD template's structured format (FR/NFR). The resulting PRD reads as complete and coherent, so the loss is invisible without manually diffing against the brainstorming output.

## Changes

- `src/bmm/workflows/2-plan-workflows/create-prd/steps-c/step-11-polish.md`: Added section "2b. Brainstorming Reconciliation" that:
  1. Checks if a brainstorming document exists in `inputDocuments` frontmatter
  2. Cross-references all brainstorming ideas against the PRD
  3. Identifies dropped ideas (especially tone, philosophy, interaction design)
  4. Presents findings to user for decision on incorporation
  5. Skips gracefully when no brainstorming input exists

## Design Rationale

- Placed in step-11 (polish) rather than a new step to avoid renumbering the 12-step workflow
- The polish step is the natural place for document quality review
- The check is conditional — only runs when brainstorming was an input document
- No breaking changes to existing flow

## Test plan

- [ ] Run PRD workflow with brainstorming document as input — verify reconciliation identifies dropped ideas
- [ ] Run PRD workflow without brainstorming input — verify step is skipped gracefully
- [ ] Verify soft/qualitative ideas (tone, philosophy) are specifically flagged

### DIFF
--- a/src/bmm/workflows/2-plan-workflows/create-prd/steps-c/step-11-polish.md
+++ b/src/bmm/workflows/2-plan-workflows/create-prd/steps-c/step-11-polish.md
@@ -99,6 +99,22 @@ Review the entire document with PRD purpose principles in mind:
 - Are technical terms used appropriately?
 - Would stakeholders find this easy to understand?
 
+### 2b. Brainstorming Reconciliation (if brainstorming input exists)
+
+**Check the PRD frontmatter `inputDocuments` for any brainstorming document** (e.g., `brainstorming-session*.md`, `brainstorming-report.md`). If a brainstorming document was used as input:
+
+1. **Load the brainstorming document** and extract all distinct ideas, themes, and recommendations
+2. **Cross-reference against the PRD** — for each brainstorming idea, check if it landed in any PRD section (requirements, success criteria, user journeys, scope, etc.)
+3. **Identify dropped ideas** — ideas from brainstorming that do not appear anywhere in the PRD. Pay special attention to:
+   - Tone, personality, and interaction design ideas (these are most commonly lost)
+   - Design philosophy and coaching approach ideas
+   - "What should this feel like" ideas (UX feel, not just UX function)
+   - Qualitative/soft ideas that don't map cleanly to functional requirements
+4. **Present findings to user**: "These brainstorming ideas did not make it into the PRD: [list]. Should any be incorporated?"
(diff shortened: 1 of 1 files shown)


bmadcode (approved):

bmadcode:
@nikolasdehor Nice idea here! This whole workflow is going to drastically change soon, but I am merging this in so the idea is not lost! Thank you, nice idea.

PR #2041, opened 2026-03-17
fix(cli): support parent directories in --custom-content path

### LINKED ISSUE #2040: Custom content path silently ignores parent directories without module.yaml
## Summary

When using `--custom-content` with a **parent directory** containing multiple module subdirectories, the installer silently skips the path and reports modules as "updated" even though no new files are actually copied. This is misleading and reduces the practicality of maintaining a multi-module workspace.

## Steps to Reproduce

**Workspace structure:**
```
~/bmad-modules/dev-productivity/
├── bmad-doc-architect/
│   ├── module.yaml
│   └── workflows/...
├── another-module/
│   ├── module.yaml
│   └── ...
```

**Command that fails silently:**
```bash
npx bmad-method install --custom-content ~/bmad-modules/dev-productivity
```

**Output says:**
```
✓ Module: bmad-doc-architect (updated)
```

**But:** No files are actually copied from source. The installer rebuilds from the cached version in `_bmad/_config/custom/`, so new/changed files in the source are never picked up.

**Command that works correctly:**
```bash
npx bmad-method install --custom-content ~/bmad-modules/dev-productivity/bmad-doc-architect
```

## Root Cause

`validateCustomContentPathSync()` in `tools/cli/lib/ui.js` checks for `module.yaml` **only at the root** of the provided path. When the path is a parent directory (no `module.yaml` at root), validation fails and the path is silently discarded. The installer then falls back to the cached version, producing a false "updated" message.

This affects **all three custom content flows**: fresh install, update, and quick-update. Additionally, the interactive prompt flow (`promptCustomContentSource()`) shares the same limitation.

## Expected Behavior

When `--custom-content` points to a directory without `module.yaml` at root:

1. Scan immediate subdirectories (1 level) for `module.yaml` files
2. Register each discovered subdirectory as a custom module source
3. Emit a clear warning when no modules are found at root or in subdirectories

## Impact

- Users maintaining a multi-module workspace (`~/bmad-modules/{domain}/{module}/`) must list every module path explicitly with commas
- The silent failure + false "updated" message leads to wasted debugging time
- Defeats the ergonomics of an organized multi-module workspace

## Fix

PR: https://github.com/bmad-code-org/BMAD-METHOD/pull/2041

**Changes:**
- Added `resolveCustomContentPaths()` method — resolves a path to an array of module directories (direct path → `[path]`, parent dir → scan subdirs)
- Updated all 3 CLI custom content flows (quick-update, update, fresh install) to use it
- Updated `validateCustomContentPathSync()` to accept parent directories
- Fixed `promptCustomContentSource()` (interactive flow) which would have crashed on parent dirs after validation was loosened
- 20 unit tests covering direct paths, parent dirs, edge cases (missing `code` field, malformed YAML, mixed paths)

## Environment

- bmad-method v6.2.0
- Node.js on Linux (WSL2)

henryavila:
## Summary

- When `--custom-content` points to a parent directory (no `module.yaml` at root), scan immediate subdirectories (1 level) for `module.yaml` files and register each as a custom module source
- Show a clear warning when no modules are found at root or in subdirectories
- Log info message when multiple modules are discovered from a single parent path

Previously the path was silently rejected and the installer fell back to cached modules, reporting "updated" without actually copying new files.

Closes #2040

## Changes

- **`tools/cli/lib/ui.js`**: Added `resolveCustomContentPaths()` method that returns `[path]` for direct module dirs or scans 1-level subdirs. Updated all 3 custom content processing flows (quick-update, update, fresh install) and `validateCustomContentPathSync()` to use it.
- **`test/test-custom-content-parent-dir.js`**: 15 tests covering direct module paths, parent dirs with multiple modules, empty dirs, and non-existent paths.
- **`package.json`**: Added `test:custom-content` script.

## Test plan

- [x] New test suite passes: `npm run test:custom-content` (15/15)
- [x] Existing tests unaffected: `npm run test:install` (232/232)
- [x] Full quality pipeline passes: lint, format, markdown lint, schema validation
- [ ] Manual test: `npx bmad-method install --custom-content ~/some-parent-dir` with multiple module subdirs

### DIFF
--- a/test/test-custom-content-parent-dir.js
+++ b/test/test-custom-content-parent-dir.js
@@ -67,6 +67,10 @@ async function createParentDirFixture() {
   await fs.ensureDir(path.join(parentDir, 'not-a-module'));
   await fs.writeFile(path.join(parentDir, 'not-a-module', 'readme.md'), '# Not a module\n');
 
+  // Module with missing code field (invalid)
+  await fs.ensureDir(path.join(parentDir, 'bad-module'));
+  await fs.writeFile(path.join(parentDir, 'bad-module', 'module.yaml'), 'name: Bad Module\n');
+
   return parentDir;
 }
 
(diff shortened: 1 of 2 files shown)


alexeyv:
Thanks for the PR but config/customization is about to go into a MAJOR surgery as part of the transition to shipping skills. So if we merged it now, it wouldn't live for very long. I'll close this PR, let's get back to this when we are out of the woods with the bigger change.

PR #2056, opened 2026-03-18
fix(opencode): use native agents/ and commands/ layout instead of skills/

arnoe:
## Summary

- OpenCode does not have a native "skills" concept — it uses `.opencode/agents/` for agent persona definitions and `.opencode/commands/` for slash commands
- The installer was incorrectly placing everything into `.opencode/skills/` as `SKILL.md` directories, which duplicated content from `_bmad/` and prevented workflows from appearing as native opencode commands
- Switch opencode to a **multi-target layout** using the existing `targets` mechanism in `_config-driven.js`:
  - `.opencode/agents/` — flat agent launcher files (using `opencode-agent` template with `mode: all` frontmatter)
  - `.opencode/commands/` — flat workflow/task/tool command files (thin loaders pointing to `_bmad/` canonical sources)

### `platform-codes.yaml`
- Replaced single `target_dir: .opencode/skills` + `skill_format: true` with a `targets` array containing two entries (agents + commands)
- Moved `.opencode/skills` into `legacy_targets` so existing skill-based installations are cleaned up on upgrade
- Removed `.opencode/agents` and `.opencode/commands` from `legacy_targets` since they are now active output targets

### `_config-driven.js`
- **`detect()`**: Added multi-target detection — checks all `targets[].target_dir` directories for bmad-prefixed entries (previously only worked for single `target_dir` with `skill_format`)
- **`findAncestorConflict()`**: Now collects target directories from both `target_dir` and `targets[]` for ancestor conflict checking, so platforms with multiple output directories are fully covered

### `test-installation-components.js`
- Updated OpenCode test suite (Suite 8) to verify multi-target layout: flat agent files in `.opencode/agents/`, commands directory created, opencode-specific `mode: all` frontmatter
- Updated ancestor conflict test (Suite 15) to check `.opencode/agents/` instead of `.opencode/skills/`
- All 234 tests pass, 0 failures

## Motivation

This was discovered by comparing a v6.0.4 installation (`poc-create-brief`) — which correctly used `.opencode/agents/` + `.opencode/commands/` — against a v6.2.0 installation which placed everything in `.opencode/skills/`. The v6.0.4 output was the correct structure for opencode's native concepts, and this PR restores that behavior using the config-driven architecture.

### DIFF
--- a/test/test-installation-components.js
+++ b/test/test-installation-components.js
@@ -410,34 +410,47 @@ async function runTests() {
   console.log('');
 
   // ============================================================
-  // Test 8: OpenCode Native Skills Install
+  // Test 8: OpenCode Multi-Target Install (agents + commands)
   // ============================================================
-  console.log(`${colors.yellow}Test Suite 8: OpenCode Native Skills${colors.reset}\n`);
+  console.log(`${colors.yellow}Test Suite 8: OpenCode Multi-Target Install${colors.reset}\n`);
 
   try {
     clearCache();
     const platformCodes = await loadPlatformCodes();
     const opencodeInstaller = platformCodes.platforms.opencode?.installer;
 
-    assert(opencodeInstaller?.target_dir === '.opencode/skills', 'OpenCode target_dir uses native skills path');
+    assert(
+      Array.isArray(opencodeInstaller?.targets) && opencodeInstaller.targets.length === 2,
+      'OpenCode installer uses multi-target layout with 2 targets',
+    );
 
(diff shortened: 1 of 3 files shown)


alexeyv:
Sorry, not doing that anymore. BMAD (and just about everybody else in this space) has switched from 20 custom, constantly changing installers to shipping skills. If you want some sort of wrapper around them for your own platform, it usually takes about 15 minutes to make one. And if you decide to publish that, please advertise it on BMAD's Discord!

PR #2118, opened 2026-03-24
feat(junie): add support for Junie

preigile:
## What
Implemented Junie support

## Why
User requests [JUNIE-1125](https://youtrack.jetbrains.com/issue/JUNIE-1125) Support specification toolkits (spec-kit, openspec, agent-os, etc) for spec-driven development

## How
- Using Junie CLI
- Based on previous commits

## Testing
- Run `npm test`
- Manual testing with `npx bmad-method install`


alexeyv:
Thanks! Following massive installer refactoring, this was used as a basis for #2142, which is now merged.

PR #2123, opened 2026-03-25
fix: add consumer propagation check to bmad-quick-dev workflow

### LINKED ISSUE #2122: fix: bmad-quick-dev workflow misses consumer propagation for new type variants
### Description

When the bmad-quick-dev workflow introduces new type variants, error codes, or DB values, the spec's Code Map only covers the primary code path. Existing consumers of the parent type/column are not enumerated during planning (step-02), leading to regressions in untouched code that the review step (step-04) misclassifies as "pre-existing" (defer) instead of "caused by this story."

### Steps to Reproduce

1. Use bmad-quick-dev to implement a bugfix that adds a new error variant to a union type
2. Step-02 planning populates Code Map with only the files being directly modified
3. Other consumers of the union type (shadow copies, string-literal switches, untyped error maps) are not listed
4. Step-04 review correctly finds the gaps via edge-case hunter, but classifier marks them as "defer" because the Code Map has no consumer inventory to cross-reference

### Expected Behavior

Step-02 should enumerate all consumers of new type variants/DB values in the Code Map. Step-03 should HALT if consumer tracing was missed.

### Proposed Fix

Two minimal additions (~388 chars total): consumer tracing instruction in step-02-plan.md, precondition check in step-03-implement.md.

gabadi:
## What

Adds consumer propagation tracing to the bmad-quick-dev workflow. Two additions (~388 chars) across step-02 and step-03.

## Why

When introducing new type variants, error codes, or DB values, the workflow's Code Map only covers files being directly modified. Consumers of the parent type/column in other modules are missed during planning, leading to regressions that the review step misclassifies as pre-existing issues. Discovered during a real bugfix where a new error variant (`FETCH_NOT_PRINTABLE`) broke an untraced consumer's error mapping.

## How

- **step-02-plan.md**: Consumer tracing instruction — grep all consumers of new types/columns during investigation, annotate in Code Map as `consumer — update required` or `consumer — excluded (reason)`.
- **step-03-implement.md**: Precondition gate — HALT if spec introduces new values but Code Map has no consumer annotations (catches missed step-02 tracing).

## Testing

Validated against the incident that exposed the gap. With these changes, step-02 would have enumerated 4 missing consumers, and step-03 would have HALTed if step-02 was skipped.

Fixes #2122

### DIFF
--- a/src/bmm-skills/4-implementation/bmad-quick-dev/step-02-plan.md
+++ b/src/bmm-skills/4-implementation/bmad-quick-dev/step-02-plan.md
@@ -12,7 +12,7 @@ deferred_work_file: '{implementation_artifacts}/deferred-work.md'
 
 ## INSTRUCTIONS
 
-1. Investigate codebase. _Isolate deep exploration in sub-agents/tasks where available. To prevent context snowballing, instruct subagents to give you distilled summaries only._
+1. Investigate codebase. _Isolate deep exploration in sub-agents/tasks where available. To prevent context snowballing, instruct subagents to give you distilled summaries only._ If the approach introduces new type variants, error codes, or DB values: grep all consumers of the parent type/column and annotate each in the Code Map as `consumer — update required` or `consumer — excluded (reason)`.
 2. Read `./spec-template.md` fully. Fill it out based on the intent and investigation, and write the result to `{wipFile}`.
 3. Self-review against READY FOR DEVELOPMENT standard.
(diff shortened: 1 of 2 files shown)


don-petry (commented):
The step-02 consumer propagation check looks solid. The annotation format is clear and the zero-consumers edge case is handled well with `consumer -- none found`.

One note: the PR description still references a step-03-implement.md HALT gate that's no longer in the diff. Consider updating the description to reflect that only step-02 is modified, to avoid confusion for future readers.

Otherwise this is clean and ready to merge.

PR #2128, opened 2026-03-26
refactor: rewrite bmad-party-mode as platform-adaptive multi-agent roundtable

pbean:
## Summary
- Rewrites the `bmad-party-mode` skill from a simple workflow delegation into an outcome-based skill that spawns independent sub-agents per BMAD agent
- Adds platform adapter system with support for Claude Code, Codex, and Gemini CLI
- Restructures steps into initialize, orchestrate, and exit phases with a new agent prompt template reference

## Test plan
- [x] Activate party mode in Claude Code and verify agents spawn as independent sub-agents
- [x] Confirm initial responses are displayed before cross-talk
- [x] Test fallback behavior on platforms without sub-agent support
- [x] Verify adapter detection works correctly per platform

### DIFF
--- a/src/core-skills/bmad-party-mode/SKILL.md
+++ b/src/core-skills/bmad-party-mode/SKILL.md
@@ -22,7 +22,9 @@ Party mode accepts optional arguments when invoked:
 
 1. **Parse arguments** — check for `--model` and `--solo` flags from the user's invocation.
 
-2. **Load config via bmad-init skill** — store `{user_name}`, `{communication_language}`, and other config vars.
+2. Load config from `{project-root}/_bmad/bmm/config.yaml` and resolve:
+  - Use `{user_name}` for greeting
+  - Use `{communication_language}` for all communications
 
 3. **Read the agent manifest** at `{project-root}/_bmad/_config/agent-manifest.csv`. Build an internal roster of available agents with their displayName, title, icon, role, identity, communicationStyle, and principles.
 


src/core-skills/bmad-party-mode/steps/step-02-orchestrate.md
```diff
@@ -0,0 +1,171 @@
...
+React briefly — agree, challenge, or build on one specific point. 2-3 sentences max. Don't repeat yourself.
+```
+
+Spawn 1-2 agents max for cross-talk. Prefer agents whose domains were referenced by others.
+
+### 8. Assess Round Quality
+
+Before presenting, quick sanity check:
+
+- **Redundancy:** If 2+ agents said essentially the same thing, present the richer version and briefly summarize the agreement rather than showing redundant full responses
```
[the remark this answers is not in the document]
pbean: all addressed in latest commit

src/core-skills/bmad-party-mode/SKILL.md
```diff
@@ -1,6 +1,110 @@
...
+- Start your response with: {icon} **{displayName}:**
+- Speak in {communication_language}.
+- Scale your response to the substance — don't pad. If you have a brief point, make it briefly.
+- Disagree with other agents when your expertise tells you to. Don't hedge or be polite about it.
+- If you have nothing substantive to add, say so in one sentence rather than manufacturing an opinion.
+- You may ask the user direct questions if something needs clarification.
+- Do NOT use tools. Just respond with your perspective.
+```
+
+**Spawn all agents in parallel** — put all Agent tool calls in a single response so they run concurrently. Use `model: "sonnet"` for the subagents to keep things fast.
```
bmadcode: we cannot do this it needs to work across tools and models. we could offer a subagent model commandline arg though. but then only use it if available.

Another option is in the future, we could let the user let us know what models they have registered or available in some config and we could allow the user to tell us what to use as a fast agent, worker agent and thinking agent for example. but beyond the scope of this.

pbean: alright made changes per our discussion and removed bmad-init dependency

PR #2139, opened 2026-03-27
fix: extend bad_spec to cover data-loss findings within story domain

### LINKED ISSUE #2138: fix: quick-dev review classification misses data-loss findings as defer
## Description

The `bad_spec` classification in step-04-review.md does not cover findings where data is silently dropped or never reaches its destination within the story's domain. This causes such findings to be misclassified as `defer` (pre-existing) even when they fall within the initiative's scope.

## Problem

In a real-world scenario, an edge case hunter correctly identified that a CreditCheck builder never mapped a field (`learningPopulation`) that its parallel sibling module already mapped. The finding was classified as `defer (pre-existing)` because the diff didn't touch that line — but the missing mapping was within the initiative's goals and was a blocking data-loss bug.

The current `bad_spec` definition only covers "caused by the change, including direct deviations from spec." It doesn't account for findings where data is silently dropped within the story's domain, even if the code predates the diff.

## Proposed Fix

Extend the `bad_spec` definition in `step-04-review.md` to include data-loss findings within the story's domain, preventing them from being misclassified as `defer`.

## Context

Discovered through a postmortem after a quick-dev workflow shipped a spec that missed a field mapping. The edge case hunter found it, but classification killed the finding.

gabadi:
## What

Extends the `bad_spec` classification in quick-dev step-04-review to cover findings where data is silently dropped or never reaches its destination within the story's domain — even if the code predates the diff.

## Why

In a real-world quick-dev run, an edge case hunter correctly identified a missing field mapping (data silently dropped at a domain boundary). The finding was misclassified as `defer` because the diff didn't touch that line. The bug shipped and was only caught by human review.

Fixes #2138

## How

- Added one sentence to the `bad_spec` definition in `step-04-review.md`: data-loss findings within the story's domain qualify as `bad_spec`, not `defer`
- No changes to `defer` definition — the gate lives where the reviewer looks when deciding what qualifies as `bad_spec`

## Testing

Validated through adversarial review and edge case analysis of the proposed change. The rule correctly reclassifies the original finding (missing `learningPopulation` on CreditCheck builder) as `bad_spec` while leaving genuinely out-of-scope pre-existing issues as `defer`.


alexeyv:
@adambiggs Thanks for the PR. Way I see it, you discovered a fundamental failure mode, but the PR is a patch for one of the symptoms thereof. It solves your situation that exposed the failure mode, but not the entire failure mode.

So I've tried to generalize the rule to anything that the intent implies but spec doesn't cover, and here is a version I came up with. Can you check that it classifies your data loss finding as bad spec? 

```
Classify each finding by checking categories in strict order. Assign the first that applies.

  This story's problem — within the story's domain:

  - intent_gap — Spec intent is incomplete; finding cannot be resolved from the spec alone. Do not
  infer intent unless there is exactly one possible reading.
  - bad_spec — Spec should have prevented this: either it directly contradicts the spec, or the spec
   omitted what was needed and a patch would be inconsistent without spec guidance. Prefer bad_spec
  over patch when in doubt.
  - patch — Pure implementation detail, trivially fixable in the diff with no spec change needed.

  Not this story's problem:

  - defer — Real issue clearly outside this story's domain, surfaced incidentally. Prefer reject
  when unsure.
  - reject — Noise or not a real issue. Drop silently.
  ```

don-petry (commented):
Good intent — catching data-loss findings as `bad_spec` rather than `defer` when they're within the story's domain is the right call.

One issue to address: the `bad_spec` definition preamble says "caused or exposed by the current change," but the specific `bad_spec` bullet line starts with "caused by the change" before the new clause adds "even if the code predates the diff." This creates an internal contradiction within the definition.

**Suggestion:** Update the `bad_spec` bullet to say "caused **or exposed** by the change" to be consistent with the preamble. This makes the data-loss addition flow naturally without contradiction.

Also worth defining "within the story's domain" more precisely — e.g., "files or modules listed in the Code Map" — to reduce ambiguity for the agent.

Small fixes, but important for consistent classification behavior.

PR #2140, opened 2026-03-27
docs(fr): translate output folder path resolution section

eatse21:
Syncs French translation with commit 1040c3c (fix: correctly resolve output_folder paths outside project root #2132).


alexeyv:
Merci! :)

PR #2168, opened 2026-03-29
fix: quick-dev self-review verifies AC coverage against parent requirements

### LINKED ISSUE #2167: fix: quick-dev drops parent-artifact requirements during AC decomposition
### Description

When `quick-dev` writes a spec for a story that derives from a parent artifact (epic plan, PRD), it can silently drop requirements assigned to the story's scope during AC decomposition. The self-review step does not verify that every upstream requirement has a corresponding acceptance criterion.

### Steps to Reproduce

1. Create an epic with an FR Coverage Map assigning specific FRs to a story
2. Run `quick-dev` to plan and implement that story
3. Observe that the generated spec may omit ACs for some assigned FRs — particularly when the epic annotation describes the *architectural mechanism* (e.g., "AccountingProvider abstraction validates this") rather than explicitly spelling out the test

### Expected Behavior

Every requirement assigned to the story's scope in the parent artifact should have a corresponding AC in the quick-dev spec.

### Actual Behavior

Requirements can be dropped when the spec author interprets architectural annotations as exemptions from testing. The self-review and adversarial review steps do not catch this because they validate the spec's internal consistency, not its completeness against upstream requirements.

### Environment

- BMad Method latest (main)
- Discovered during production use with epic-level planning

### Proposed Fix

Two minimal changes to `bmad-quick-dev`:
- **step-02-plan.md**: Extend the self-review to verify AC coverage against parent artifact requirements
- **spec-template.md**: Add a comment hint in the AC section reminding that parent-assigned requirements need ACs

Both changes are general-purpose (no project-specific logic) and preserve quick-dev's speed philosophy.

gabadi:
## What

Adds a general-purpose safeguard to `quick-dev` so that specs derived from parent artifacts (epics, PRDs) don't silently drop assigned requirements during AC decomposition.

## Why

When a story derives from an epic with requirement mappings, `quick-dev` can omit ACs for assigned requirements — particularly when annotations describe architectural mechanisms rather than explicit test expectations. The self-review and adversarial review steps validate internal consistency but not completeness against upstream requirements.

Fixes #2167

## How

- **step-02-plan.md**: Extend self-review instruction to verify every upstream-assigned requirement has a corresponding AC
- **spec-template.md**: Add comment hint in AC section reminding that parent-assigned requirements need ACs

Both changes are one line each — no project-specific logic, preserves quick-dev's speed philosophy.

## Testing

Validated against a real postmortem where FR18/FR19 were correctly assigned at the epic level but dropped during spec writing, causing a false PASS on traceability.

### DIFF
--- a/src/bmm-skills/4-implementation/bmad-quick-dev/spec-template.md
+++ b/src/bmm-skills/4-implementation/bmad-quick-dev/spec-template.md
@@ -54,7 +54,8 @@ context: [] # optional: max 3 project-wide standards/docs. NO source code files.
 
 <!-- Tasks: backtick-quoted file path -- action -- rationale. Prefer one task per file; group tightly-coupled changes when splitting would be artificial. -->
 <!-- If an I/O Matrix is present, include a task to unit-test its edge cases. -->
-<!-- AC covers system-level behaviors not captured by the I/O Matrix. Do not duplicate I/O scenarios here. -->
+<!-- AC covers system-level behaviors not captured by the I/O Matrix. Do not duplicate I/O scenarios here.
+     If requirements were assigned from a parent artifact, every assigned requirement must have an AC here. -->
 
 **Execution:**
 - [ ] `FILE` -- ACTION -- RATIONALE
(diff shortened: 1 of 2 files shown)


alexeyv (changes requested):
Augment is right. An AC from an upstream document can end either in the ACs, or the I/O matrix section of the spec file. Not in both.

don-petry (commented):
Agreeing with @alexeyv's feedback — the current wording says "every assigned requirement must have a corresponding AC" but requirements can legitimately be covered by the I/O Matrix instead.

**Suggested fix:** Change the wording to:

> Every upstream-assigned requirement must be represented in either the Acceptance Criteria or the I/O Matrix — not both, and none omitted.

This preserves the intent (no requirement slips through) while respecting that I/O Matrix is a valid coverage mechanism. The template comment hint should get the same treatment.

Once that's addressed, this looks good to merge.

Issue #2162, opened 2026-03-29
[DOCS] bmad & pencil


d3vr3n:
### What type of documentation issue is this?

Addition

### Documentation location

https://github.com/bmad-code-org/BMAD-METHOD/tree/main/docs/reference  ... 3rd Party Tools

### What's the issue?

Some users have expressed frustration / confusion using pencil within BMAD workflows.

 If the model is not fully aware of the existence of Pencil it will just see the mcp as a nice to have, not as fundamental to the project.  Before the PRD, UX and Architecture docs are drafted the agent will confirm with you what files will provide input to the creation of the doc  ... the .pen file MUST be included in this list , otherwise the user can expect several (avoidable) course corrections during sprints.

### Suggested improvement

 After some trouble shooting and course correction. I found that  if you specify pencil in your tooling before `_bmad-output/planning-artifacts/architecture.md`  and enforce in `_bmad-output/project-context.md `... and  use pencil in your interactions/planning/designing  , it will pick up the pencil mcp  naturally... the sooner you bring pencil into  the process the better ... if you bring it in after architecture (and obviously ux)  and dont put in the guard rails ... it (claude-bmad-combo) will drive you nuts.

It may be an idea to mention this in  the docs

abrahamgreyson:
+1

Kavleri:
Hi @d3vr3n ! I am new to open source and would love to make this my very first contribution. Could I take this issue? I can add a warning/recommendation about bringing Pencil into the process early in the 3rd Party Tools documentation. Let me know if that sounds good!

bmadcode:
There are many external tools that could be used with bmad - cli or mcp - I do not think we can start documenting every one of them.

But the work around for this is through the agent and workflow skill customization process (being reenabled soon), or adding clear instructions to CLAUDE.md, AGENTS.md or similar.

But best path forward will be the customization route for this. IE the user will be able to inject custom prompts or scripts or configurations into most workflows and agents - so this nudge to use pencil can be integrated - or any other cli, mcp or future tool

PR #2171, opened 2026-03-30
docs: add pencil usage guidelines to core-tools #2162

Kavleri:
This PR addresses issue #2162. Added recommendations for using Pencil in the early planning stages.


bmadcode:
Thank you @Kavleri but I want to avoid adding specific call out of specific extra tools as they come and go so frequently.

But with the new customization feature - this guidance can be baked in directly to agent and workflow customization configs!

PR #2181, opened 2026-04-01
feat: AgentVibes TTS integration — install prompt + core skill

paulpreibisch:
## Summary

Adds optional AgentVibes TTS integration to the BMAD install flow, giving users an easy path to voice-enable their agents (especially party mode).

### Changes

**`src/core-skills/bmad-install-agentvibes/`** (new skill)
- Guides users through `npx agentvibes install` post-BMAD-setup
- Documents per-agent voice config, platform support, and troubleshooting
- Registered in `module-help.csv` with menu code `AV`

**`tools/installer/core/installer.js`**
- After the install summary, shows a 3-option select prompt:
  - **Yes** — runs `npx agentvibes install` inline
  - **No** — skip, reminder shown in Next Steps
  - **Tell me more** — displays description + GitHub link without leaving the installer
- Non-fatal: cancelled prompt or install failure doesn't affect the BMAD install result
- Default is **No** so it's fully opt-in and non-intrusive

## Why

Party mode agents are currently silent unless users independently discover and install AgentVibes. This surfaces the integration at exactly the right moment — right after BMAD is set up.

## Testing

\`\`\`bash
npm link   # from BMAD-METHOD root
mkdir ~/test-project && cd ~/test-project
bmad install
# AgentVibes prompt appears after install summary
\`\`\`

### DIFF
--- a/src/core-skills/bmad-install-agentvibes/SKILL.md
+++ b/src/core-skills/bmad-install-agentvibes/SKILL.md
@@ -1,0 +1,65 @@
+---
+name: bmad-install-agentvibes
+description: 'Install AgentVibes TTS voice system for BMAD agents. Gives each agent a unique voice, personality, and audio effects. Use when user wants to add voice/TTS to their BMAD setup, or when bmad-party-mode is active and agents are silent.'
+---
+
+# Install AgentVibes
+
+AgentVibes is an open-source Text-to-Speech system that gives BMAD agents unique voices, personalities, and audio effects. Once installed, party mode agents speak aloud — each with their own voice.
+
+## What AgentVibes Provides
+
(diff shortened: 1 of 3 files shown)


bmadcode:
Hi @paulpreibisch, thank you for the detailed integration work and for the genuine effort that went into AgentVibes. Promoting a third-party tool from the core installer is something we want to avoid, both to keep core install lean and to stay neutral on TTS providers. The right path for this is a community module that users can install separately. The plugin marketplace pattern (bmad-plugins-marketplace) is the intended home for things like this. Closing as out of scope for core. Genuinely appreciate the contribution and would happily link to a standalone module.

PR #2189, opened 2026-04-02
fix(dev-agent): fix review findings subsection and step mismatch in workflow

cantecim:
## What

Agent allowed to write to Review Findings section in dev story skill

## Why

In a strict mode, after a successfull dev story run, the agent can not mark corresponding findings completed.

## How

In step 8, it expects "Review Follow-ups (AI)" list, but according to the review skill, the reviewer agent writes the findings under a general "Review Findings" section.

This patch fixes review and dev-story skills


don-petry (commented):
The intent is right — dev-story should be able to mark review findings as resolved. However there's a design concern to address:

**Write-ownership conflict:** `bmad-code-review` (step-04-present.md) creates and writes to "Review Findings." Granting dev-story full write access to the same section creates a conflict with no coordination protocol. Two workflows claiming write access to the same section is a recipe for overwrites.

**Missing implementation:** Step 8 of the dev-story workflow doesn't contain logic to actually mark Review Findings checkboxes as completed. The permission grant alone is inert without corresponding workflow instructions.

**Suggestion:** Consider one of these approaches:
1. **Narrow the permission** to checkbox-toggling only (not content modification) and add Step 8 logic for marking findings done
2. **Add a "Review Findings Status" sub-section** that dev-story owns, separate from the findings themselves that code-review owns
3. **Use a different mechanism** like appending resolution notes rather than modifying the original findings

This would be a more robust solution than the quick permission grant. Happy to help work through the design if useful.

cantecim:
> The intent is right — dev-story should be able to mark review findings as resolved. However there's a design concern to address:
> 
> **Write-ownership conflict:** `bmad-code-review` (step-04-present.md) creates and writes to "Review Findings." Granting dev-story full write access to the same section creates a conflict with no coordination protocol. Two workflows claiming write access to the same section is a recipe for overwrites.
> 
> **Missing implementation:** Step 8 of the dev-story workflow doesn't contain logic to actually mark Review Findings checkboxes as completed. The permission grant alone is inert without corresponding workflow instructions.
> 
> **Suggestion:** Consider one of these approaches:
> 
> 1. **Narrow the permission** to checkbox-toggling only (not content modification) and add Step 8 logic for marking findings done
> 2. **Add a "Review Findings Status" sub-section** that dev-story owns, separate from the findings themselves that code-review owns
> 3. **Use a different mechanism** like appending resolution notes rather than modifying the original findings
> 
> This would be a more robust solution than the quick permission grant. Happy to help work through the design if useful.

You are right on the approaches.

This is very interesting, Claude models and GPT were already only marking items under Review Findings completed in dev-story skill run. I'm not 100% sure how this behavior is inferred from the workflow, but I see, "Mark task checkbox [x] in "Tasks/Subtasks → Review Follow-ups (AI)" section" instruction must lead to it.

In this current approach, marking review findings completed is actually complete and done in dev-story skill, however, code-review skill uses the Review Findings section to write those to the story artifact.

I have thought the same as your second suggestion, as briefly explained in the "How".

To achieve this:

1. code-review skill should get an update to use "Review Follow-ups (AI)" section strictly
2. dev-story skill should get an update to be privileged to write same section only for marking them completed

In this case, dev-story can not change any decision-needed finding to a concrete action (in reality, it can because of the same model's very own mental model)

What do you think about this exact approach? Any objections/blind spots?

I'll look into it when I have a better time

cantecim:
@coderabbitai help

cantecim:
@coderabbitai review

cantecim:
@coderabbitai review

cantecim:
This looks ready for a maintainer review @bmadcode

muratkeremozcan (dismissed):

bmadcode:
@cantecim sorry for the late review here, i missed the ping in email. there are some conflicts, can you update this? just fly, the bmad-dev-story is about to be drastically rewritten and story tracking will change, but since it might still be a few weeks before ready if this is updated we can get this merged.

cantecim:
I fixed the conflicts @bmadcode. Could you guys give a quick look again @muratkeremozcan

bmadcode:
Thank you for this fix, and for your patience — it predates two major restructures. Closing because both patched files are gone: dev-story is now a legacy shim under `v6-shims/` and code-review was rewritten under `ship/bmad-code-review`, so the diff can no longer apply. Worth noting: a similar section-name mismatch still exists between `ship/bmad-code-review/steps/step-04-present.md` and the v6 shim — if it still bites you, a fresh issue against the new paths would be appreciated.

—

Issue #2275, opened 2026-04-16
bmad-dev-story: ATDD gate checks file existence only — no Gherkin, no E2E enforcement for UI flows


pablontiv:
## What's broken

`bmad-dev-story` has three enforcement gaps that allow stories to complete without BDD or E2E coverage:

1. **No ATDD gate before implementation** — when the TEA module is installed, there is no check for an ATDD checklist before dev starts. The red/green cycle begins immediately regardless of whether acceptance tests exist.

2. **Gherkin not verified** — even when an ATDD checklist file exists, its content is never scanned for Gherkin scenarios (`Given/When/Then`). A checklist file with only prose passes silently.

3. **E2E tests are opt-in** — Step 6 reads "Include end-to-end tests for critical user flows *when story requirements demand them*" and the Step 9 DoD mirrors this. User-facing stories complete without E2E tests unless the story author explicitly opted in.

## Impact

Projects with the TEA module installed and user-facing flows complete multiple epics with zero E2E tests and no Gherkin-described acceptance criteria, despite the tooling being present and configured.

## Expected behavior

- If TEA is installed and no ATDD checklist exists → warn and offer to halt before implementation starts
- If checklist exists but has no Gherkin scenarios → warn that BDD coverage is missing
- E2E tests required (not optional) for user-facing flows; unit tests required for business logic and API routes

## Environment

- BMad version: 6.3.0 (BMM module)
- Observed with TEA module installed, `tea_use_playwright_utils: true`

bmadcode:
Closing — TEA now lives in its own repo at https://github.com/bmad-code-org/bmad-method-test-architecture-enterprise. ATDD enforcement scope and Gherkin/E2E checks belong over there.

PR #2272, opened 2026-04-15
fix: add TEA ATDD pre-check gate to dev-story initialization

pablontiv:
## What
Adds an explicit TEA ATDD pre-check gate in Step 1 of `dev-story` initialization. When TEA is installed and no ATDD checklist exists for the story, the workflow warns and asks for confirmation before proceeding.

## Why
When TEA module is installed, `dev-story` previously had zero enforcement of ATDD. The only TEA reference was a Step 10 "Optional" suggestion placed after the story was already complete. Entire epics could be implemented with 0% new test coverage and no acceptance scaffolds, with no warning.
Fixes #2271. Related to #2243 and PR #2260 — this targets the missing *entry gate*, while #2260 addresses *artifact consumption* inside the workflow (both fixes are complementary).

## How
- Added TEA detection + ATDD checklist lookup at `<anchor id="task_check" />` in Step 1
- If TEA installed and no checklist found: explicit warning + ask to confirm or halt
- Changed Step 10 "Optional: ..." wording to remove the "Optional" label for TEA automate suggestion

## Testing
Verified locally with TEA 1.7.2 + BMM 6.3.0: dev-story now prompts when `testarch-atdd` output is missing. Existing behavior unchanged when TEA is not installed or checklist already exists.

PR #2276, opened 2026-04-16
fix: enforce BDD/E2E in dev-story — Gherkin gate + E2E required for UI flows

pablontiv:
## What

Four targeted changes to `src/bmm-skills/4-implementation/bmad-dev-story/workflow.md` that enforce BDD/E2E discipline when the TEA module is present.

## Why

Without these changes, stories with user-facing flows complete without E2E tests and without Gherkin-described acceptance criteria, even when the TEA module is installed and configured. See #2275.

Note: this extends the intent of #2272 (ATDD gate) with Gherkin content verification and E2E opt-out (rather than opt-in).

## How

- **Step 1**: adds BDD/SDD/E2E pre-check gate — warns when TEA is installed but no ATDD checklist exists; warns when checklist has no Gherkin scenarios (`Given/When/Then`)
- **Step 5**: red phase now identifies test type (unit vs E2E) before writing the failing test, making the E2E red phase explicit for UI tasks
- **Step 6**: E2E tests required for all user-facing flows (removes the `when story requirements demand them` opt-in conditional)
- **Step 9 DoD**: same language fix — E2E mandatory for UI flows, unit mandatory for logic/API

## Testing

Verified locally against a project with TEA module installed: `bmad-dev-story` now warns at Step 1 when no ATDD checklist is found, and warns when the checklist contains no Gherkin scenarios.

### DIFF
--- a/src/bmm-skills/4-implementation/bmad-dev-story/workflow.md
+++ b/src/bmm-skills/4-implementation/bmad-dev-story/workflow.md
@@ -157,6 +157,31 @@ Load config from `{project-root}/_bmad/bmm/config.yaml` and resolve:
 
     <anchor id="task_check" />
 
+    <!-- BDD/SDD/E2E pre-check gate -->
+    <check if="TEA module is installed (bmad-testarch-atdd skill exists in project _bmad directory or skills folder)">
+      <action>Search {test_artifacts} for an ATDD checklist matching pattern: `atdd-checklist-{{story_key}}*.md` or `*{{story_key}}*atdd*.md`</action>
+      <check if="no ATDD checklist found for {{story_key}}">
+        <output>⚠️ **No ATDD checklist found for {{story_key}}**
+
+Missing before implementation:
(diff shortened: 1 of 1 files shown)


src/bmm-skills/4-implementation/bmad-dev-story/workflow.md
```diff
@@ -157,6 +157,31 @@ Load config from `{project-root}/_bmad/bmm/config.yaml` and resolve:
 
     <anchor id="task_check" />
 
+    <!-- BDD/SDD/E2E pre-check gate -->
```
bmadcode, line 160: the nodules need to remain decoupled.

another plan I have for this is to allow modules apply configuration values when installed to other modules, hence the rationale partially behind the recent addition to main for customization of agents and workflows.

bmadcode:
@pablontiv this is a good idea, but we cannot hard couple like this independent modules. the way we can handle this is through modules on install applying customization to other modules somehow through the newly introduced module config - not exactly sure on a final plan of how this will work though.

PR #2307, opened 2026-04-24
fix: installer live version for external modules

muratkeremozcan:
Fixes the BMad installer so official external modules show the latest npm version in the module picker instead of stale version metadata from the local external-module cache.

 What Changed

  - Updated the installer UI in BMAD-METHOD to prefer live npm version lookup for official external modules
  - Kept the existing local metadata resolution as a fallback when npm is unreachable
  - This prevents cached module clones under ~/.bmad/cache/external-modules/* from causing stale version labels in the installer

Why

The installer was showing Test Architect (v1.13.1) even after bmad-method-test-architecture-enterprise@1.14.0 had been published, because the picker label came from cached local metadata rather than the npm registry.

 Validation

  - npx prettier --check tools/installer/ui.js
  - npx eslint tools/installer/ui.js --ext .js,.cjs,.mjs,.yaml --max-warnings=0
  - git diff --check
  - verified live npm lookup returns 1.14.0 for bmad-method-test-architecture-enterprise

### DIFF
--- a/test/test-installation-components.js
+++ b/test/test-installation-components.js
@@ -2622,6 +2622,229 @@ async function runTests() {
     }
   }
 
+  // --- Official module picker uses git tags for external module labels ---
+  {
+    const { UI } = require('../tools/installer/ui');
+    const prompts = require('../tools/installer/prompts');
+    const channelResolver = require('../tools/installer/modules/channel-resolver');
+    const { ExternalModuleManager } = require('../tools/installer/modules/external-manager');
+
+    const ui = new UI();
+    const originalOfficialListAvailable39 = OfficialModules.prototype.listAvailable;
+    const originalExternalListAvailable39 = ExternalModuleManager.prototype.listAvailable;
+    const originalAutocomplete39 = prompts.autocompleteMultiselect;
+    const originalSpinner39 = prompts.spinner;
+    const originalWarn39 = prompts.log.warn;
+    const originalMessage39 = prompts.log.message;
+    const originalResolveChannel39 = channelResolver.resolveChannel;
+
+    const seenLabels39 = [];
+    const spinnerStarts39 = [];
+    const spinnerStops39 = [];
+    const warnings39 = [];
+
+    OfficialModules.prototype.listAvailable = async function () {
+      return {
+        modules: [
+          {
+            id: 'core',
+            name: 'BMad Core Module',
+            description: 'always installed',
+            defaultSelected: true,
+          },
+        ],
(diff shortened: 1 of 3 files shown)


bmadcode:
Hey Murat, thanks for jumping on this. I want to flag a concern before this lands: npm is not the source of truth for external modules, and this will regress some labels.

Quick audit against the current registry:

| Module | `npmPackage` in registry | npm latest | Actual latest | Result with this PR |
|---|---|---|---|---|
| TEA (`bmad-method-test-architecture-enterprise`) | present | 1.15.0 | 1.15.0 | correct today |
| BMB (`bmad-builder`) | present | **1.1.0** | **1.7.0** | **regresses to 1.1.0** |
| CIS (`bmad-creative-intelligence-suite`) | present | 404 | 0.2.0 | falls back to stale cache |
| GDS (`bmad-game-dev-studio`) | present | 404 | 0.4.0 | falls back to stale cache |

BMB is the clearest signal: npm has it frozen at 1.1.0 while the real module is 1.7.0. After this PR the picker would show 1.1.0, which is worse than the current cached-version behavior.

Bigger picture: we're moving away from publishing external modules to npm. Going forward, only the installer itself (delivered via \`npx\`) plus \`core\` and \`bmm\` (coupled to the installer package) will be published to npm. Everything else (TEA, CIS, GDS, BMB, future modules) is versioned via git tags on its own GitHub repo and installed by \`git clone\`. TEA will stop publishing to npm too once the git-tag-driven release flow is fully in place.

So the source of truth for the picker label needs to match the source of truth for the install itself: the module's GitHub tags. That's exactly what #2305 just built with \`channel-resolver.js\` — \`resolveChannel(moduleInfo, { channel: 'stable' })\` hits the GitHub \`/tags\` endpoint and returns the highest stable semver tag, no clone required. It already handles caching, fallbacks, and the stable/next/pinned concept. Reusing it here keeps the picker and the installer reading from the same well.

Two other asks while you're in here:

1. **Show an upgrade arrow when a newer version is available.** If the module is already installed locally at vX and latest is vY (X != Y), render the label as \`Name (vX → vY)\` instead of just \`(vY)\`. For fresh installs or same-version, \`(vY)\` alone is fine.
2. **Add a single spinner** (\`Checking latest module versions...\`) wrapping the parallel lookups, and a one-line warn if all lookups fail so the user knows they're seeing cached versions.

Happy to pair on the \`resolveChannel\` integration if useful. Thanks for picking this up.

PR #2316, opened 2026-04-26
fix(installer): support local custom-source modules in resolveInstalledModuleYaml and TOML key

cidemaxio:
## What
Two installer fixes for local custom-source modules (installed via `--custom-content` pointing to a local directory rather than a registry/git source).

## Why
When using `--custom-content` with a local path, the module is never written to the external cache (`~/.bmad/cache/external-modules/`), so `resolveInstalledModuleYaml` returned `null` for these modules. This caused manifest writing to silently skip the module's config schema, and the TOML config file ended up with the full display name as the section key (e.g. `[modules.MDO: Maxio DevOps Operations]`) instead of the short code (`[modules.mdo]`).

## How
- `resolveInstalledModuleYaml` (`project-root.js`): refactored candidate-path search into a shared `searchRoot()` helper; added fallback to `CustomModuleManager._resolutionCache` for local custom-source modules whose cache path does not exist; added `*-setup/assets/module.yaml` as a BMB standard candidate path
- `manifest-generator.js`: read the `code` field from each module's `module.yaml` and use it as the TOML section key (`[modules.<code>]`) instead of the installer's internal module name

## Testing
Installed a local custom-source module (`--custom-content /path/to/module`) and confirmed:
- `resolveInstalledModuleYaml` correctly resolves via the resolution cache fallback
- Generated `config.toml` uses `[modules.mdo]` (code) not `[modules.MDO: Maxio DevOps Operations]` (display name)
- Re-install / quick-update round-trip succeeds without errors

### DIFF
--- a/tools/installer/core/manifest-generator.js
+++ b/tools/installer/core/manifest-generator.js
@@ -435,6 +435,9 @@ class ManifestGenerator {
     // this means user-scoped keys (e.g. user_name) could mis-file into the
     // team config, so the operator should notice.
     const scopeByModuleKey = {};
+    // Maps installer moduleName (may be full display name) → module code field
+    // from module.yaml, so TOML sections use [modules.<code>] not [modules.<name>].
+    const codeByModuleName = {};
     for (const moduleName of this.updatedModules) {
       const moduleYamlPath = await resolveInstalledModuleYaml(moduleName);
       if (!moduleYamlPath) {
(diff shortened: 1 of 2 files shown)


bmadcode (approved):
@cidemaxio Thank you for the well-scoped fix and clear write-up! Both the local custom-source resolution and the TOML section-key bug are real issues, and the changes look clean. Merging.

Issue #2333, opened 2026-04-27
Feature Request: HTTP URLs in --custom-source for custom module installation


LanyGuan:
**Describe your idea**
When installing custom modules via the --custom-source flag, the tool should support HTTP-based Git repository URLs and preserve the original protocol instead of forcibly rewriting it to HTTPS.

**Why is this needed?**
In enterprise or internal network environments, custom modules are often not published publicly. Their source code is hosted on internal Git repositories that are only accessible within the company's private network. These internal Git servers typically do not enable HTTPS, as they rely on the isolated network for security. Currently, when an HTTP Git URL is provided, the tool automatically replaces it with an HTTPS prefix, causing the clone operation to fail. This makes it impossible to install custom modules from internal HTTP-only Git repositories.

**How should it work?**
The tool should respect the protocol explicitly specified by the user in the --custom-source URL. If the user provides an http:// address, it should be used as-is without being rewritten to https://. Alternatively, the tool could skip protocol enforcement for custom source URLs, or provide a flag/configuration option to allow insecure HTTP connections for internal repositories.

**PR**
I am not currently working on a PR for this

**Additional context**
This is a common scenario in corporate/enterprise environments where internal GitLab or Gitea instances are hosted on private networks without TLS termination. The current behavior blocks adoption of the tool in such environments.

bmadcode:
Closing as duplicate of #2264, which tracks the same root cause (parseSource hardcodes HTTPS scheme, dropping http:// from --custom-source URLs). The fix there will preserve the scheme captured by the regex. Please follow #2264 for progress, and feel free to reopen if you have additional repro details not covered there.

PR #2335, opened 2026-04-27
fix(installer): preserve HTTP protocol in custom module clone URLs

LanyGuan:
Previously, parseSource() hardcoded 'https://' when building cloneUrl, forcing http:// Git URLs (e.g., internal LAN hosts) to upgrade to https. This broke cloning for self-hosted Git servers that only serve over HTTP.

- Capture the protocol from the regex match instead of discarding it
- Update JSDoc and inline comments to document HTTP support
- Update install-custom-modules docs (EN, ZH, VN) to list HTTP URL type

Fixes the --custom-source flag for [#2333](https://github.com/bmad-code-org/BMAD-METHOD/issues/2333)

### DIFF
--- a/docs/how-to/install-custom-modules.md
+++ b/docs/how-to/install-custom-modules.md
@@ -68,6 +68,7 @@ Select **Yes**, then provide a source:
 | Input Type            | Example                                           |
 | --------------------- | ------------------------------------------------- |
 | HTTPS URL (any host)  | `https://github.com/org/repo`                     |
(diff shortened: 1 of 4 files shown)


LanyGuan:
@coderabbitai review

LanyGuan:
@augmentcode review

LanyGuan (commented):
review

LanyGuan:
@coderabbitai review

PR #2339, opened 2026-04-27
feat(config): per-user global config across worktrees

### LINKED ISSUE #2338: Per-user global config not shared across git worktrees
## Problem

Users working across multiple git worktrees or cloned repos must re-enter personal settings (`user_name`, `communication_language`, `user_skill_level`) in every project's `_bmad/config.user.toml`. The installer writes defaults (e.g. OS username) into this file, but these are machine-generated — the user's actual preferences are lost across projects.

## Solution

A global user layer at `~/.bmad/config/` that **overrides installer defaults** but is still overridden by hand-authored project customizations.

### Priority order (lowest → highest)

**`resolve_config.py` (5 layers):**
1. `_bmad/config.toml` (installer team)
2. `_bmad/config.user.toml` (installer user — machine defaults)
3. `~/.bmad/config/config.user.toml` (global user preferences)
4. `_bmad/custom/config.toml` (custom team)
5. `_bmad/custom/config.user.toml` (custom user)

**`resolve_customization.py` (4 layers):**
1. `{skill}/customize.toml` (skill defaults)
2. `~/.bmad/config/{skill}.user.toml` (global user preferences)
3. `_bmad/custom/{skill}.toml` (project team)
4. `_bmad/custom/{skill}.user.toml` (project user)

### Why this order

- Installer-generated files (`_bmad/config.user.toml`) contain machine defaults (OS username, "English", "intermediate") — not the user's actual preferences
- Global user config (`~/.bmad/config/`) is hand-authored — it should override those machine defaults
- Custom project config (`_bmad/custom/`) is also hand-authored but project-specific — it should override everything

### Files changed

| File | Change |
|------|--------|
| `src/scripts/resolve_config.py` | Add global user layer between installer layers and custom layers |
| `src/scripts/resolve_customization.py` | Add global user layer between skill defaults and project layers |
| `test/test-config-resolution.js` | New — 14 unit tests for both scripts |

No installer changes needed — the scripts are copied to `_bmad/scripts/` via `_installSharedScripts`.

## Follow-up: Setup UX

No guided setup exists yet for `~/.bmad/config/`. Users must create it manually. Two options considered:

1. **`--global-user-config` flag on the installer** — reuses existing prompt logic to write answers into `~/.bmad/config/config.user.toml` instead of the project-local file
2. **Dedicated skill (e.g. `bmad-setup-global-config`)** — interactive guided setup, could also handle global skill customizations (`{skill}.user.toml`)

Leaving this for a future iteration — input welcome on which approach (or both) to pursue.

jrevillard:
## Summary

- Add a global user config layer at `~/.bmad/config/` that applies as lowest-priority fallback across all projects and worktrees
- `resolve_config.py` now merges 5 layers (was 4): global user → base team → base user → custom team → custom user
- `resolve_customization.py` now merges 4 layers (was 3): global user → skill defaults → project team → project user
- 14 new tests covering priority ordering, backward compat, `--key` flag, and cross-layer value preservation

## Problem

Users working across multiple git worktrees or cloned repos must re-enter personal settings (`user_name`, `communication_language`, `user_skill_level`) in every project's `_bmad/custom/config.user.toml`. This is tedious and error-prone.

## Solution

A global user layer at `~/.bmad/config/config.user.toml` (and `~/.bmad/config/{skill}.user.toml` for skill customization) is merged first as the lowest-priority fallback. Any project-level override always wins. Missing global dir is fully backward-compatible — existing installs work unchanged.

## Test plan

- [x] All 313 existing tests pass
- [x] 14 new tests for global layer merge behavior
- [x] Backward compat: missing `~/.bmad/config/` doesn't break anything
- [x] Priority chain verified: global < project layers
- [ ] Manual test: create `~/.bmad/config/config.user.toml`, verify it's resolved
- [ ] Manual test: project override wins over global value

Closes #2338


bmadcode:
all configs have not yet converted to toml, which they need to - and the config script which is currently unused in the workflows (it should be) also needs to convert soon with the change to toml for the next version roll out

can you carve out the change to the config loader and lean this PR just to the customization change to support the ~/.bmad/custom - and a change is coming soon to the other script and I will include a similar ~/.bmad/configs

Thanks @jrevillard

jrevillard:
Thanks for the feedback! Two things I'd like to clarify before splitting:

**1. Directory paths**

My implementation uses `~/.bmad/config/` for both scripts (e.g. `~/.bmad/config/config.user.toml` and `~/.bmad/config/{skill}.user.toml`). You mentioned `~/.bmad/custom` and `~/.bmad/configs` — should we use different paths?

**2. Skills still read `config.yaml` directly**

The reason I updated the skill files to call `resolve_config.py` is that skills currently read `_bmad/bmm/config.yaml` directly, completely bypassing the TOML merge mechanism. Without that change, the global user layer is never resolved by agents. 

If I strip the skill updates from this PR, the global customization layer in `resolve_customization.py` will work, but the global config layer in `resolve_config.py` won't be consumed until skills are updated separately.

Should I:
- **(A)** Keep only `resolve_customization.py` changes in this PR (the global config script change + skill updates come later)?
- **(B)** Keep both resolver changes but defer the skill updates to a follow-up PR?

bmadcode:
@jrevillard this is probably worth revisiting now that all configs are in place - but there are some conflicts. worth updating still?

bmadcode:
actually - have another plan here @jrevillard - going to close this one for now.

PR #2348, opened 2026-04-28
fix(config): promote project_name to core (closes #2279)

### LINKED ISSUE #2279: [BUG] bmad-brainstorming loads project_name from core/config.yaml, but project_name is owned by bmm
## Summary

`src/core-skills/bmad-brainstorming/workflow.md` instructs the LLM to load `project_name` from `_bmad/core/config.yaml`, but `project_name` is defined by the `bmm` module — it lives in `_bmad/bmm/config.yaml`. On a fresh install, the key simply isn't present where brainstorming looks for it.

## Expected

`bmad-brainstorming` loads config values from the correct per-module location(s) so `project_name` resolves on a standard install.

## Actual

`src/core-skills/bmad-brainstorming/workflow.md` lines 35–37:

```markdown
Load config from `{project-root}/_bmad/core/config.yaml` and resolve:

- `project_name`, `output_folder`, `user_name`
- `communication_language`, `document_output_language`, `user_skill_level`
- `date` as system-generated current datetime
```

On a 6.3.0 install, `_bmad/core/config.yaml` contains only:

```yaml
user_name
communication_language
document_output_language
output_folder
```

`project_name` lives in `_bmad/bmm/config.yaml`, which is consistent with `src/bmm-skills/module.yaml` defining `project_name` as a bmm module variable (prompt: "What is your project called?"). `src/core-skills/module.yaml` does not define `project_name`.

The installer itself already knows this. In `tools/installer/core/installer.js` (around lines 1414–1415), its own config resolver looks up `project_name` in `bmm/config.yaml`. Only `bmad-brainstorming`'s SKILL doc points at the wrong file.

## Blast radius

Swept `src/core-skills/*` for skills that reference `project_name`. Only `bmad-brainstorming` has this mismatch. All bmm skills that need `project_name` (e.g., `bmad-product-brief`, `bmad-agent-dev`) correctly load from `bmm/config.yaml`.

## Relevant paths

- Skill doc (wrong config path): `src/core-skills/bmad-brainstorming/workflow.md:35`
- Canonical home of `project_name`: `src/bmm-skills/module.yaml` + `_bmad/bmm/config.yaml` on install
- Installer's own resolver (correct behavior reference): `tools/installer/core/installer.js` lines ~1414–1415

## Suggested fix

Change line 35 from `{project-root}/_bmad/core/config.yaml` to `{project-root}/_bmad/bmm/config.yaml`. A follow-up could split the load instruction so `user_name`/`communication_language`/`output_folder` come from core and `project_name` from bmm separately, since they genuinely live in different modules — but the one-line path fix is the immediate correctness fix.

PR incoming.

bmadcode:
## Summary

Closes #2279.

`project_name` was a bmm-specific prompt despite being a universal project-level concept used by every module. Core skills like `bmad-brainstorming` already load from `_bmad/core/config.yaml` and try to resolve `project_name` from there — but the value was being written under `[modules.bmm]` instead, so brainstorming was silently broken. Worse, users who install only bmb/cis (without bmm) couldn't run brainstorming at all because `project_name` had no source.

This PR moves `project_name` into `core/module.yaml` where it belongs alongside `user_name`, `communication_language`, etc., and adds migration so existing installs don't lose their prior `project_name` value on upgrade.

## Architectural fit

Comment block in `manifest-generator.js:468`: *"Core keys are always known... they must be stripped from every non-core module bucket because legacy `_bmad/{mod}/config.yaml` files spread core values into each module. Core belongs in `[core]` only — workflows that need `user_name`/language/etc. read `[core]` directly."*

Today `project_name` was the lone outlier — a project-level key that lived under `[modules.bmm]`. After this change, it slots into `[core]` like every other shared key.

## Files

- `src/core-skills/module.yaml` — declare `project_name` after `user_name` (prompt, default `{directory_name}`).
- `src/bmm-skills/module.yaml` — remove the bmm definition; add `## project_name` to the inherited-keys header comment.
- `tools/installer/modules/official-modules.js` — generic hoist migration in `loadExistingConfig`. After reading per-module `config.yaml` files, any key now declared in core but appearing under a non-core module gets moved to core's bucket (preserving any pre-existing core value as the canonical one). Without this, `writeCentralConfig`'s partition logic silently drops the prior answer on the next `quick-update`.
- `tools/installer/ui.js` — add `project_name: path.basename(directory)` to the `--yes` core seed so non-interactive fresh installs populate it. Matches what `buildQuestion` resolves for `{directory_name}` in the interactive flow.
- `test/test-installation-components.js` — Suite 43 (9 assertions): schema move, hoist migration, precedence rule. Suite 35 fixture updated: `project_name` moved to core bucket; stale bmm copy left in place to verify strip.

Brainstorming workflow.md is untouched — it was already correctly loading from `core/config.yaml`. It just couldn't find `project_name` there because the schema didn't declare it. After this PR, it works.

## Behavior

| Scenario | Before | After |
|---|---|---|
| Fresh interactive install | Prompted for project_name under bmm | Prompted under core |
| Fresh `-y` install | project_name lands in `[modules.bmm]` | project_name lands in `[core]` |
| `-y` install without bmm (e.g. only bmb/cis) | project_name absent entirely; brainstorming broken | project_name in `[core]` from core's prompt |
| `quick-update` on existing pre-fix install | project_name stripped from bmm and lost | project_name hoisted to core, value preserved |
| Existing install where user manually set project_name in `[core]` | n/a | Hoist preserves the core value, strips bmm duplicate |

## Test plan

- [x] Fresh install `-y --modules bmm --tools claude-code`: `[core] project_name = "<dirname>"` in config.toml
- [x] Existing install with `bmm/config.yaml::project_name = "my-legacy-project"` and core lacking it: `quick-update` migrates to `[core]`, strips from `[modules.bmm]`
- [x] Existing install with project_name in BOTH bmm and core: hoist does not overwrite the core value
- [x] Test Suite 43: 9 assertions for schema + migration + precedence
- [x] Suite 35 (writeCentralConfig): updated to assert project_name now belongs in `[core]`
- [x] Full suite: 310/310 passing

### DIFF
--- a/src/bmm-skills/module.yaml
+++ b/src/bmm-skills/module.yaml
@@ -5,14 +5,10 @@ default_selected: true # This module will be selected by default for new install
 
 # Variables from Core Config inserted:
 ## user_name
+## project_name
 ## communication_language
 ## document_output_language
 ## output_folder
-
-project_name:
-  prompt: "What is your project called?"
-  default: "{directory_name}"
-  result: "{value}"
 
 user_skill_level:
   prompt:
--- a/src/core-skills/module.yaml
+++ b/src/core-skills/module.yaml
@@ -11,6 +11,11 @@ user_name:
   default: "BMad"
   result: "{value}"
 
+project_name:
+  prompt: "What is your project called?"
+  default: "{directory_name}"
+  result: "{value}"
+
 communication_language:
   prompt: "What language should agents use when chatting with you?"
   scope: user
--- a/test/test-installation-components.js
+++ b/test/test-installation-components.js
@@ -2943,8 +2943,38 @@ async function runTests() {
       'hoist still strips the duplicate from bmm so writeCentralConfig partition stays clean',
     );
 
+    // Malformed config.yaml (parses to a scalar) must not crash loadExistingConfig
+    // or the hoist pass — they should treat it as "no config for that module"
+    // and continue. Regression for augment review on PR #2348.
+    const fixtureRoot43c = await fs.mkdtemp(path.join(os.tmpdir(), 'bmad-fixture-43c-'));
+    const bmadDir43c = path.join(fixtureRoot43c, '_bmad');
+    await fs.ensureDir(path.join(bmadDir43c, '_config'));
+    await fs.writeFile(path.join(bmadDir43c, '_config', 'manifest.yaml'), 'modules: []\n', 'utf8');
+    await fs.ensureDir(path.join(bmadDir43c, 'core'));
+    await fs.ensureDir(path.join(bmadDir43c, 'bmm'));
+    // Scalar YAML — yaml.parse returns the literal 42 (truthy non-object).
+    // Pre-fix this crashed _hoistCoreKeysFromLegacyModuleConfigs with
+    // "Cannot use 'in' operator to search for 'project_name' in 42".
+    await fs.writeFile(path.join(bmadDir43c, 'core', 'config.yaml'), '42\n', 'utf8');
+    await fs.writeFile(path.join(bmadDir43c, 'bmm', 'config.yaml'), 'project_name: rescued\n', 'utf8');
+
+    const officialModules43c = new OfficialModules();
+    let crashErr;
+    try {
+      await officialModules43c.loadExistingConfig(fixtureRoot43c);
+    } catch (error) {
+      crashErr = error;
+    }
+    assert(!crashErr, 'loadExistingConfig does not crash on a scalar core/config.yaml', crashErr?.stack);
+
+    assert(
+      officialModules43c.existingConfig.core?.project_name === 'rescued',
+      'scalar core gets replaced with {} and bmm.project_name still hoists in',
+    );
+
     await fs.remove(fixtureRoot43).catch(() => {});
     await fs.remove(fixtureRoot43b).catch(() => {});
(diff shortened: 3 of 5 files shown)


bmadcode:
Addressed augment review (medium) in 93f65475:

**Issue**: \`_hoistCoreKeysFromLegacyModuleConfigs\` could hit a TypeError ("Cannot use 'in' operator to search for ... in 42") if \`config.yaml\` parsed to a truthy non-object — and the bug was upstream in \`loadExistingConfig\`, which stored any truthy parse result.

**Fix (defense in two places)**:
- \`loadExistingConfig\`: only keep parses that are plain objects (not scalars or arrays). A corrupt \`config.yaml\` is now treated the same as a parse error — skipped, not crashed-on.
- \`_hoistCoreKeysFromLegacyModuleConfigs\`: belt-and-suspenders type guards on \`_existingConfig.core\` and on each module \`cfg\` in the loop, so anyone populating \`_existingConfig\` directly in a future code path can't crash this helper.

**Test**: 2 new assertions in Suite 43 — a scalar \`core/config.yaml\` (\`42\\n\`) must not crash \`loadExistingConfig\`, and a legacy \`bmm.project_name\` must still hoist into a clean core bucket.

Full suite: 312/312 passing.

PR #2364, opened 2026-05-01
feat(bmm): add bmad-investigate skill

anderewrey:
## What
Adds `bmad-investigate`, a forensic case-investigation skill under Amelia's menu (`IN`) that produces an evidence-graded case file engineers can hand off cleanly.

## Why
See https://github.com/bmad-code-org/BMAD-METHOD/discussions/2265
BMM has no skill for pre-fix forensic investigation. This carves it out as its own discipline with evidence grading, hypothesis preservation, and a structured case file that survives session
interruptions and hands off cleanly.

## How
- New `SKILL.md` plus a flat `case-file-template.md`.
- One calibrated procedure: the same five-outcome walk leans into hypothesis tracking and timeline reconstruction for symptom-driven cases, and into I/O mapping, control-flow filtering, and
boundary tracing for exploration cases. The case-file template covers the union.
- Integration: registered in `bmad-agent-dev`'s menu as code `IN`, picked up by `bmad-create-prd` step-01 discovery so case files surface as PRD input, listed on the workflow-map Phase 4 row,
explained at `docs/explanation/forensic-investigation.md`. EN + FR for all public docs.

## Testing
Used with real investigation made against brownfield source code.


anderewrey:
@bmadcode I reworked completely the original PR https://github.com/bmad-code-org/BMAD-METHOD/pull/2340

src/bmm-skills/4-implementation/bmad-investigate/SKILL.md
```diff
@@ -0,0 +1,204 @@
+---
+name: bmad-investigate
+description: Forensic case investigation with evidence-graded findings, calibrated to the input. Use when the user asks to investigate a bug, trace what caused an incident, walk through unfamiliar code, or build a mental model of a code area before working on it. Accepts a ticket ID, log file path, diagnostic archive, error message, code area name, problem description, or a path to an existing case file.
```
bmadcode: Please reduce the description 'short human readable phrase. Use when user says/does X or Y or Z.

Please let me know if you ran thing through the bmad builders workflow skill analysis mode and it did not catch this.

src/bmm-skills/4-implementation/bmad-investigate/SKILL.md
```diff
@@ -0,0 +1,204 @@
...
+- A symptom-driven case (bug, incident, error) leans into hypothesis tracking, timeline reconstruction, and a fix
+  direction.
+- A no-symptom case (understanding an unfamiliar area) leans into I/O mapping, control-flow filtering, and a mental
+  model.
+- An existing case file is read in full, identified open hypotheses and backlog items are surfaced, and new findings
+  land under a new dated `## Follow-up: {date}` block.
+
+After every outcome, update the case file, present what was learned, and stop for human input before continuing.
+
+### Outcome 1: Scope and stronghold are established
```
bmadcode: these investigations are going to be context heavy - need to make better usage of subagents in parallel that understand the initial issue, can fan out and investigate and just write or return to the parent only relevant findings. - potentially each outcome here could be an agent if it makes more sense to do so. but minimally would put in some sub agent directing language

could benefit from having a customize.toml to specify common input or research sources (ie registered mcp or cli tools or other services) to avoid having to specify below.

src/bmm-skills/4-implementation/bmad-investigate/SKILL.md
```diff
@@ -0,0 +1,204 @@
...
+If the user arrives with a hypothesis, register it as Hypothesis #1 and target evidence collection at confirming or
+refuting it, while still scanning broadly for the unexpected.
+
+Find a stronghold: a confirmed piece of evidence (an error message, a function name, an HTTP route, a configuration
+parameter, a test case). Anchor the case here.
+
+Initialize `{case_file}` from `case-file-template.md`. Fill in: Case Info, Problem Statement, initial Evidence
+Inventory.
+
+Present scope, stronghold, file path, proposed approach. Halt.
```
bmadcode: why is there a halt instruction at the end of each outcome? are all 5 outcomes supposed to be run?

bmadcode (commented):
This is a great idea for a skill, have some ideas to take it up a notch . Investigations need good non corruptible logs, as the template markdown grows it will be more likely the llm will non deterministically screw it up, throw stuff out, change things it should not etc...

I would suggest instead some python deterministic scripts that append to jsonl log - a format we will start using in other places. scripts are simple and calls to it can be like "call the 'caselog append {findings}' or something along those lines - then nondeterministically the history of findings and decisions can be maintained reliably.

would also be really cool to have a visual representation similar to a findings report html file that pops up upon completion or demand - this could be something that scans the log or case file and renders a nice output.

these is probably a bit of scope creep to initially get this in, but I think some deterministic scripting of events to a append log of some sort in the case of investigations could be useful - and will make its way also into other workflows.

src/bmm-skills/4-implementation/bmad-investigate/SKILL.md
anderewrey: Description updated.

Ran the workflow-builder quality scan three times during this rework; none of the runs flagged the description as too long.

anderewrey:
> I would suggest instead some python deterministic scripts that append to jsonl log - a format we will start using in other places. scripts are simple and calls to it can be like "call the 'caselog append {findings}' or something along those lines - then nondeterministically the history of findings and decisions can be maintained reliably.


I'm not experienced with the JSONL / deterministic-scripts approach you're describing. The generated .md case file is what I've mainly used in real investigations and it's been well maintained by the skill (no drift in practice). Leaving this open for someone with that specific need (and the experience) to tackle.

src/bmm-skills/4-implementation/bmad-investigate/SKILL.md
anderewrey: Added a Delegation discipline Principle (>10K tokens or 5+ files → subagent returning JSON) and an explicit parallel-tool-calls Principle. Outcome 2 fans out per evidence category, and Outcome 4 batches grep/glob/git-log.

Didn't go as far as outcome-per-subagent (feels like overkill for now).

src/bmm-skills/4-implementation/bmad-investigate/SKILL.md
anderewrey: Wording changed to "Pause for user with the recap above; wait for direction.". That allows the AI to get feedback from the user. This is part of the continuous calibration method.

bmadcode:
fixed the decsription that got split across multiple rows - merged. thanks @anderewrey

PR #2379, opened 2026-05-13
fix: support nested group paths in SSH Git URLs

hanhnt2-hblab:
## What
SSH URL parser now accepts nested group paths (e.g. `git@host:group/subgroup/repo.git`)

## Why
GitLab and self-hosted Git servers commonly use nested groups/subgroups in repository paths. The existing regex only matched two-segment `owner/repo` paths, causing `--custom-source` to reject valid SSH URLs with "Not a valid Git URL or local path".

## How
- Changed SSH owner capture group from `([^/]+)` to `(.+?)` in `parseSource()`
- Repo segment `([^/.]+?)` still captures only the final path component
- `cacheKey` and `displayName` naturally extend to include the full owner path

## Testing
Verified nested-group SSH URLs (e.g. `git@gitlab.example.com:org/team/project.git`) now parse correctly. Standard two-segment URLs (`git@github.com:owner/repo.git`) continue to work unchanged.


bmadcode:
Thanks @hanhnt2-hblab

PR #2390, opened 2026-05-18
fix(installer): restore automator install compatibility

dickymoore:
## Summary

Restore BMad Automator install/update compatibility in the BMAD installer.

## Problem

Current installer behavior has two linked problems:

1. legacy installs can still have Automator recorded as `baut`, while the bundled BMAD module registry currently advertises `automator`
2. the bundled Automator registry entry points at `skills/module.yaml`, but `bmad-automator` is a pure-skill payload repo and does not ship that module definition path

That combination makes `quick-update` / reinstall runs treat an installed Automator as having no source, producing:

- `Skipping 1 module(s) - no source available: baut`

and then removing the existing Automator install as if it were deselected.

## Scope

This PR is intentionally limited to the installer compatibility path for that failure:

- restore the bundled Automator module code to `baut` for compatibility with existing installs
- treat `automator` and `bmad-automator` as CLI aliases for `baut`
- force `--yes` existing-install runs onto the full update path when explicit module/channel/tool flags are present, so commands like `--modules automator --next automator` are not silently downgraded to quick-update
- add `source-root` support to external official module resolution
- point the bundled Automator registry entry at `payload/.claude/skills`
- copy the Automator Python runtime (`source/scripts`, `source/src`) into the installed `bmad-story-automator` skill during install

## Out of Scope

- broader external-module/source-root redesign
- marketplace registry changes
- unrelated Automator runtime behavior

## Validation

- commit hook also ran the full BMAD test stack successfully:
  - `npm run test:refs`
  - `npm run test:install`
  - `npm run test:urls`
  - `npm run test:channels`
  - `npm run lint`
  - `npm run lint:md`
  - `npm run format:check`

### DIFF
--- a/bmad-modules.yaml
+++ b/bmad-modules.yaml
@@ -33,10 +33,10 @@ modules:
 
   bmad-automator:
     url: https://github.com/bmad-code-org/bmad-automator
-    module-definition: skills/module.yaml
-    code: automator
-    name: "BMad Automator Epic Builder Experimental"
-    description: "EXPERIMENTAL: only supports claude and codex currently"
+    source-root: payload/.claude/skills
+    code: baut
+    name: "BMad Automator (Experimental)"
+    description: "BMAD story automation skills for create/dev/QA/review/retro orchestration"
     defaultSelected: false
     type: experimental
     npmPackage: bmad-story-automator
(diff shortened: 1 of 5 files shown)


dickymoore:
Superseded by #2391. This narrower draft only fixes the stale installed-module update failure ( retained in existing manifests) and leaves broader Automator installability work out of scope.

PR #2391, opened 2026-05-18
fix(installer): preserve stale installed modules during update

dickymoore:
## Summary
Fixes the update-time compatibility path for already-installed stale modules such as legacy Automator `baut` entries.

## Changes
- preserve installed modules whose source is no longer resolvable
- preserve their config during modify/update and quick-update flows instead of dropping it or mis-scoping it

## Why
This PR is about safe handling of existing installs. It prevents update flows from breaking or silently losing config when a previously installed module can no longer be resolved from current sources.

## Scope limits
- does not make Automator freshly installable again
- does not add new module source resolution behavior
- does not change registry metadata

## Relationship To Other PRs
This PR is independent.

It solves the stale-install compatibility problem only. The separate “make Automator installable again” work is handled by:
- `BMAD-METHOD#2392` installer support for `source_root`
- `bmad-plugins-marketplace#26` registry metadata for Automator

## Validation
- full pre-commit suite passed during commit

### DIFF
--- a/test/test-installation-components.js
+++ b/test/test-installation-components.js
@@ -165,6 +165,39 @@ async function runTests() {
   console.log('');
 
   // ============================================================
+  // Test 4b: Preserve installed modules with no source
+  // ============================================================
+  console.log(`${colors.yellow}Test Suite 4b: Preserve Installed Modules Without Source${colors.reset}\n`);
+
+  let staleInstallRoot;
+  try {
+    staleInstallRoot = await fs.mkdtemp(path.join(os.tmpdir(), 'bmad-stale-module-'));
+    const staleBmadDir = path.join(staleInstallRoot, '_bmad');
+    const staleModuleDir = path.join(staleBmadDir, 'baut');
+    await fs.ensureDir(staleModuleDir);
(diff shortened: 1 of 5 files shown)


dickymoore:
Addressed on the latest branch tip in `1541840ec`.

The retained-module config fix now preserves prior `[modules.<code>]` blocks verbatim when a module is being kept installed but its schema source can no longer be resolved. That avoids the remaining quick-update drift where user-scoped values could be rewritten from `_bmad/config.user.toml` into `_bmad/config.toml`.

Scope is still intentionally narrow:
- fresh unknown-schema modules keep the existing fallback behavior and still get a normal `[modules.*]` section
- only retained modules with an existing prior block take the verbatim-preservation path
- no install-layout or source-resolution expansion was added

Validation: `node test/test-installation-components.js` and the full pre-commit test/lint/format suite passed on commit.

dickymoore:
Cross-linking for review context:

- This PR is the stale-install compatibility fix only.
- It is independent of the fresh-installability work.
- The separate “make Automator installable again” pair is:
  - `BMAD-METHOD#2392`
  - `bmad-plugins-marketplace#26`

Recommended merge order:
- merge this PR independently
- merge `#26` and `#2392` together, or `#26` first and `#2392` immediately after

PR #2392, opened 2026-05-18
fix(installer): support Automator source-root installs

dickymoore:
## Summary
Restores Automator installability on the installer side with a minimal upstream-compatible change set.

## Changes
- teach external module resolution to honor `source_root`
- add targeted installer-component coverage for source-root modules and path-escape rejection

## Why
Automator's installable module content now lives under a module root directory (`skills/`) rather than being discovered through a `module.yaml` entrypoint path. The installer therefore needs to support official modules declared via `source_root`.

## Scope limits
- does not change stale-module preservation behavior
- does not add CLI aliases for `automator`
- does not redesign registry loading
- does not change Automator runtime behavior
- does not reintroduce deleted fallback-registry machinery

## Relationship To Other PRs
This PR is part of the pair that restores fresh Automator installability.

It should be merged together with, or immediately after:
- `bmad-plugins-marketplace#26` which updates the live Automator registry entry to `source_root: skills`

Related but separate:
- `BMAD-METHOD#2391` is the stale-install compatibility fix for existing `baut` installs

## Validation
- full pre-commit suite passed during commit (`npm test`, lint, markdown lint, format check)`

### DIFF
--- a/test/test-installation-components.js
+++ b/test/test-installation-components.js
@@ -85,6 +85,23 @@ async function createTestBmadFixture() {
   return fixtureDir;
 }
 
+async function createSourceRootModuleFixture() {
+  const repoRoot = await fs.mkdtemp(path.join(os.tmpdir(), 'bmad-source-root-module-'));
+  const sourceRoot = path.join(repoRoot, 'skills');
+  const skillNames = ['bmad-story-automator', 'bmad-story-automator-review'];
+
+  for (const skillName of skillNames) {
+    const skillDir = path.join(sourceRoot, skillName);
+    await fs.ensureDir(skillDir);
+    await fs.writeFile(
+      path.join(skillDir, 'SKILL.md'),
+      ['---', `name: ${skillName}`, `description: ${skillName} description`, '---', '', `${skillName} body`].join('\n'),
+    );
+  }
+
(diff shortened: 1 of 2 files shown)


dickymoore:
Cross-linking for review context:

This PR is one half of the fresh-installability fix for Automator.

Paired PR:
- `bmad-plugins-marketplace#26`

Why the pair is needed:
- this PR teaches the installer to consume official modules declared via `source_root`
- `#26` publishes Automator in the live registry with `source_root: skills`

Related but separate:
- `BMAD-METHOD#2391` is only the stale-install compatibility fix for existing `baut` installs

Recommended merge order:
- merge `#26` and this PR together, or `#26` first and this PR immediately after

dickymoore:
Closing this draft as unnecessary after rechecking current upstream state.

Automator already ships the standard BMAD module metadata on `main` at `skills/module.yaml` and `skills/module-help.csv`, and current upstream BMAD installs `baut` successfully through the normal `module_definition: skills/module.yaml` path. That means the `source_root` installer extension is no longer the right fix.

The remaining structural cleanup is now tracked in:
- `bmad-code-org/bmad-automator#14`

That Automator PR keeps BMAD conventional by aligning the module metadata with the existing registry/installer code instead of teaching BMAD a special module shape.

PR #2403, opened 2026-05-21
fix: capture dev story baseline commits

seven-steven:
## What
Capture `baseline_commit` when `bmad-dev-story` starts a fresh ready-for-dev story.

## Why
Repro: implement stories through earlier epics with `bmad-dev-story`, then later run `/bmad-checkpoint-preview` to inspect epic 1-3 while already developing epic 4. The story files have no `baseline_commit`, so BMAD cannot identify the original pre-implementation HEAD and falls back to inferred ranges like merge-base or latest commit, which can produce inaccurate previews.

## How
- Allow `baseline_commit` in story YAML frontmatter.
- Capture `git rev-parse HEAD` only before a fresh ready-for-dev story moves forward.
- Preserve existing baselines and support projects without sprint status.

## Testing
Ran `git diff --check`, `npm run validate:refs`, and `npm run validate:skills`. Commit hooks also ran `npm run test` successfully.


bmadcode:
thanks @seven-steven

PR #2409, opened 2026-05-22
feat(docs): Add sidebar order validator for doc frontmatter

eatse21:
## Summary

Introduces a new CI-time validator (`tools/validate-sidebar-order.js`) that checks `sidebar.order` values in markdown frontmatter across the `docs/` tree.

## Why

Multiple English doc files shared the same `sidebar.order` value (e.g. `explanation/` had duplicates across multiple files), causing unpredictable sidebar rendering. Translations also accumulate structural issues over time with no automated check.

**What it checks:**
- **English docs (strict):** duplicate `sidebar.order` values within a directory, gaps in the ordering sequence (ex. 1, 2, 4), and missing `order:` fields when a `sidebar:` block is present
- **Translations:** same structural rules as English (errors) + order drift from the English counterpart (non-blocking warnings)

**Changes:**
- New `tools/validate-sidebar-order.js` — plain Node.js, no dependencies, modeled after the existing `validate-doc-links.js` pattern
- New `npm run docs:validate-sidebar` script in `package.json`
- Integrated into the `quality` script so it runs in CI
- Added to the Husky pre-commit hook — runs when docs/ files are staged, alongside docs:validate-links

**Usage:** `node tools/validate-sidebar-order.js` — exits 0 if all orders are valid, exits 1 on any structural error. Warnings for translation drift are printed but don't affect the exit code, to allow English docs to move forward without blocking errors from missing translations.

Covers the acceptance criteria from the spec: duplicate detection, gap detection, missing-order detection, cross-language drift warnings, and graceful skipping of files without sidebar frontmatter.

:warning: In current state of english/translations, it will make the CI fail since ordering needs to be fixed overall first

---

Implemented by `quick-dev` and its review cycles


eatse21:
augment review

bmadcode:
Hi @eatse21.

I just pushed two commits to your branch:

1. **Merge `main`** — your branch was a bit behind; this is a clean fast-forward merge with no conflicts.
2. **Fix sidebar.order duplicates and gaps** across English + all four translations so the validator passes.

### What changed

**English (`docs/{explanation,how-to,reference}/`):** resolved all duplicates and renumbered for the new pages added since orders were last set. The `explanation/` section now follows a deliberate reading order (foundations → analysis → solutioning → implementation → quality → collaboration → FAQ).

**Translations (cs, fr, vi-vn, zh-cn):** mirrored the English structural ordering where files exist, then compacted to 1..N within each directory to eliminate gaps caused by translations missing some English files.

### Result

- ✅ **0 errors**, validator passes
- ⚠️ ~42 non-blocking drift warnings remain where translation directories have fewer files than English — these are expected per your validator's design

### Note on translation order

I made structural (numeric) changes only — no content was modified. If translation contributors prefer a different reading order in their language, they can adjust freely; the validator only requires unique, gap-free 1..N within each directory.

Thanks again for the contribution! Let me know if anything looks off.

PR #2422, opened 2026-05-25
fix(bmm-skills): define project_context in dev-story, sprint-planning, sprint-status

bdsoha:
## What
Declare `project_context` in `bmad-dev-story`, `bmad-sprint-planning`, and `bmad-sprint-status`.

## Why
All three call `Load {project_context}` but never define the variable — so `project-context.md` is silently skipped. `bmad-code-review` and `bmad-quick-dev` declare it correctly; same shape, just missing here.

## How
One line added to each skill's Step 4 Load Config, mirroring `bmad-code-review/SKILL.md:49`.

## Testing
Grep — every skill referencing `{project_context}` now also declares it.


bmadcode (approved):

bmadcode:
thanks @bdsoha

PR #2447, opened 2026-06-02
fix(bmad-dev-story): unify review findings handoff between code-review and dev-story

oneby-wang:
## What
Unifies the code review rework loop around the `Review Findings` section so `bmad-dev-story` consumes the same findings that `bmad-code-review` writes. Code review now sends unresolved patch action items back to `ready-for-dev` instead of `in-progress`.

## Why
`Senior Developer Review (AI)` and `Review Follow-ups (AI)` overlapped with `Review Findings`, creating a fragile CR -> DS handoff. This makes the flow simpler and keeps decision-needed findings in the existing code-review decision path.

## How
- Updated `bmad-code-review` Step 4 to return stories with unresolved review work to `ready-for-dev` and clarify the next-step prompt.
- Updated `bmad-dev-story` to detect `Review Findings`, halt on unresolved `[Review][Decision]`, and prioritize unchecked `[Review][Patch]` findings.
- Updated the dev-story checklist and allowed story-edit wording to cover `Review Findings` checkboxes.

## Testing
Ran `npm run validate:skills` and `npm run lint:md`. Also verified old `Senior Developer Review`, `Review Follow-ups`, and `[AI-Review]` markers no longer remain in `bmad-dev-story`.


bmadcode (approved):

bmadcode:
why move it back to ready instead of in progress @oneby-wang ?

oneby-wang:
Hi @bmadcode, thanks for reviewing. I think this is a good topic to discuss.

I moved the story back to ready-for-dev because I believe we're currently missing a dedicated state for stories that require changes after review.

In this case, the review findings require fairly significant updates, so I wanted to make it clear that additional development work is needed before the story is ready for another review cycle.

Keeping the story in in-progress introduces some ambiguity, as there is no clear distinction between:

- a story that is actively being developed, and
- a story that has been reviewed, received change requests, and is awaiting rework.

We could keep such stories in in-progress, but that would likely require additional logic (for example, "in-progress + unresolved review findings") to identify them correctly. Even then, it would still be difficult to distinguish between stories actively being worked on and stories that failed review but have not yet been picked up for rework.

Also, even without introducing a new state, moving the story back to ready-for-dev while retaining the review findings provides a fairly clear signal that the review did not pass and that further development work is required. In my view, that is still more explicit than leaving the story in in-progress.

If we conclude that a dedicated workflow state is needed for this scenario, I think it would be better to address that separately in a dedicated PR.

bmadcode:
Thank you, but this has been resolved with other changes

PR #2448, opened 2026-06-03
fix(bmad-retrospective): renumber retrospective workflow steps

oneby-wang:
## What
Renumbered the `bmad-retrospective` workflow so document discovery is Step 2 and the remaining workflow steps run sequentially through Step 13.

## Why
The previous `n="0.5"` step appeared after Step 1 but could be misordered by tools or readers that sort step numbers numerically, even though it depends on the epic discovered in Step 1.

## How
- Changed the project document discovery step from `0.5` to `2`.
- Shifted the former Steps 2-12 to Steps 3-13.
- Updated internal step references and the `customize.toml` completion hook comment to match the new numbering.

## Testing
Ran the retrospective skill validator and the full required pre-push quality command on Node 22: `npm ci && npm run quality`.


bmadcode:
thanks @oneby-wang

PR #2461, opened 2026-06-10
fix: remove empty skill-group dirs left in _bmad after install

bdsoha:
## Problem

After installing, empty grouping folders are left behind under `_bmad/<module>` — e.g. `_bmad/bmm/1-analysis`, `2-plan-workflows`, `3-solutioning`, `4-implementation`. These aren't output/artifact dirs and aren't meant to stick around; they're just leftovers from the install.

## Cause

`_cleanupSkillDirs` (`tools/installer/core/installer.js`) removes each skill's own directory after skills are copied to the IDE target, but it only deletes the skill dir itself — not the grouping folder above it. Once every skill under a group is removed, the empty group folder remains.

## Fix

After removing a skill dir, walk up and drop any parent directories that are now empty, stopping at (and never removing) the bmad root. Module-level files like `config.yaml` and `_config/` keep their parents alive, so only genuinely-empty grouping dirs get pruned.

## Tests

Appended a suite to `test/test-installation-components.js` covering:
- empty skill-group dir is pruned
- empty nested group dir (e.g. `1-analysis/research`) is pruned
- module-level files are preserved
- the bmad root is never removed

Full suite passes.


bmadcode (approved):

bmadcode:
Thank you @bdsoha

PR #2464, opened 2026-06-11
fix(installer): gitignore default _bmad/config.user.toml

### LINKED ISSUE #2456: [BUG] Default _bmad/config.user.toml file is not ignored during installation
### Description

The default `_bmad/config.user.toml` file is not added to `.gitignore` during installation, unlike `_bmad/custom/config.user.toml`.

This creates inconsistent behavior between the default config location and the custom config location. User-specific configuration files should be gitignored.

### Steps to reproduce

1. Run 'npx bmad-method install'

### Expected behavior

_bmad/config.user.toml should be added to .gitignore during installation/update in the same way as _bmad/custom/config.user.toml.

User-specific configuration files should be gitignored.

### Actual behavior

_bmad/config.user.toml is not gitignored during installation, while _bmad/custom/config.user.toml is gitignored.

### Screenshots

_No response_

### Which module is this for?

### BMad Version

6.8.0

Claude Code

### Operating System

Windows

### Relevant log output

```shell

```

### Confirm

bdsoha:
## What
Seed `_bmad/.gitignore` with `config.user.toml` during install so the default user config is gitignored, matching the existing `_bmad/custom/.gitignore` rule for `custom/*.user.toml`.

## Why
`_bmad/config.user.toml` holds personal, machine-specific install answers — the same kind of content already ignored under `custom/`. It was being left tracked, so personal config could be committed by accident.

Fixes #2456

## How
- Extend `_installSharedScripts` to call a new `_ensureUserConfigGitignored` helper after seeding the custom `.gitignore`.
- The helper creates `_bmad/.gitignore` when missing, or appends the entry to an existing file that doesn't list it — so the rule reaches both fresh installs and updates of older projects without duplicating the line.

## Testing
Added Test Suite 46 to `test/test-installation-components.js` covering the fresh-seed, idempotent re-install, and append-to-existing cases. Full `npm test` suite passes.


bmadcode:
Thanks @bdsoha, and sorry this sat as long as it did — the fix was correct and the reasoning was right: `_bmad/config.user.toml` holds the same kind of personal, machine-specific content already ignored under `custom/`, so leaving it tracked was inconsistent.

**BMAD is changing how it is distributed.** The project is moving to a flat `skills/` tree installed through `npx skills` and plugin marketplaces, and `tools/installer/` is removed entirely as part of that change. Both files this PR touches — `tools/installer/core/installer.js` and `test/test-installation-components.js` — go away with it, so there's nothing left to merge this into.

The underlying concern still stands and is now a requirement on the new install path: personal config must not land in a state where it's committed by accident. If you want to check that the replacement gets it right once it's up, that would be welcome.

---

PR #2473, opened 2026-06-15
fix: adjust nav height for dual announcement banners

slur16105:
## What

Adjusted the navigation height to account for two top announcement banners displayed on the documentation site.

## Why

The site now displays two announcement banners, but the navigation height was still calculated for a single banner. This caused layout overlap issues on both desktop and mobile views.

## How

* Updated `--sl-nav-height` from `6.25rem` to `9rem`
* Adjusted the height calculation comment to reflect two announcement banners
* Verified the navigation and page content alignment with both banners enabled

## Testing

Tested locally in desktop and mobile responsive views. Confirmed that the header and page content no longer overlap and spacing is rendered correctly.

### DIFF
--- a/website/src/styles/custom.css
+++ b/website/src/styles/custom.css
@@ -15,7 +15,7 @@
    ============================================ */
 :root {
   --ai-banner-height: 2.75rem;
-  --sl-nav-height: 6.25rem; /* Base nav height (~3.5rem) + banner height (2.75rem) */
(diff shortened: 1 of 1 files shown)


bmadcode:
thanks @slur16105

slur16105:
Thank you! BMAD has been incredibly helpful for me while coding with AI. I’m glad I could contribute even a little.

PR #2474, opened 2026-06-16
feat(bmm): add improve architecture skill

kylewvl-ship-it:
What
Adds bmad-improve-architecture as a built-in BMM skill in the framework source so it is installed with BMAD instead of requiring separate module registration. It also registers the skill in the BMM help catalog so it appears as a first-party solutioning workflow.
Why
This is needed to make the architecture-improvement capability part of the normal BMAD install experience, with the skill already baked into BMM. It removes the extra install/setup step and makes the workflow available immediately in projects that install BMAD.
How
Added a new built-in skill at src/bmm-skills/3-solutioning/bmad-improve-architecture/ with SKILL.md, LANGUAGE.md, DEEPENING.md, INTERFACE-DESIGN.md, and customize.toml.
Adapted the skill to the framework’s built-in BMM workflow model by reading standard BMM config and writing artifacts under {planning_artifacts}/architecture-review.
Registered the new skill in src/bmm-skills/module-help.csv so the installer-generated help/catalog includes it at install time.
Testing
Ran npm ci successfully and ran npm run test:install, which passed. I also attempted the full repo validation path, but npm run quality and npm run validate:skills are currently blocked by existing repo-baseline issues unrelated to this change.


bmadcode:
@kylewvl-ship-it there are some really good ideas in here, i dont think this needs to be a separate skill though, this could be another mode progressively disclosed if chosen or if an existing architecture is input. could you rework this as a feature of the new bmad-architecture that routes to if user wants to do this.

im not sure of the terminology deepen, have never heard that in regards to architecture - maybe there is some more common verbiage. 

one of the goals with the evolving bmad-architecture and upcoming v7 approach is altitude appropriate decisioning at each layer for the architecture, producing an architectural spine and optional other artifact or deep dive into certain areas as needed. but generally that can be done further down the tree in individual epics, keeping the highest level architecture focused on the invariants and decisions that must be made so lower levels do not make conflicting decisions.

kylewvl-ship-it:
Thank you for the feedback.Im still quite new to this so it helps massively.

bmadcode:
Thank you for this — the ideas here were well received at the time. Closing because the restructure removed `src/bmm-skills/3-solutioning/` entirely and the architecture skill was rewritten at `plan/bmad-architecture`, so nothing remains for this diff to apply to. If the improvements still resonate, re-proposing them against the new skill would be very welcome.

—

Issue #2452, opened 2026-06-05
bmad-investigate: method-internal jargon ("stronghold") leaks into user-facing communication


alexeyv:
## What happened

In a live session, the `bmad-investigate` skill (BMad v6.8.0, BMM module) had the agent tell the user:

> "Let me verify that path and anchor the stronghold in code."

The user had no idea what this meant and had to stop the investigation to ask. Their feedback, lightly censored: "I hate this bird's language. Use plain English, please."

## Why it happens

The skill prompt defines an internal method vocabulary and then explicitly instructs the agent to use it in user-facing output. From the skill's instructions:

- Principles section: "**Stronghold first.** Anchor in one Confirmed piece of evidence and expand outward."
- Outcome 1: "Find a stronghold: a Confirmed piece of evidence (error message, function name, HTTP route, config parameter, test case). Anchor here."
- Outcome 1 recap instruction: "Otherwise present scope, **stronghold**, file path, proposed approach."

That last line tells the agent to present "the stronghold" to the user verbatim. The term is meaningful inside the method but is opaque to a user who just asked to investigate a bug. The agent dutifully follows the prompt and the result reads like RPG terminology instead of an engineering status update.

"Stronghold" is the clearest offender, but the general pattern is worth a pass: terms that structure the agent's reasoning (fine) vs. terms that reach the user (should be plain language).

## Suggested fix

Keep the internal discipline and vocabulary — it's good prompt engineering — but separate it from user-facing language. Options, roughly in order of preference:

1. Add one line to the skill's Communication principles, e.g.: "User-facing messages use plain language; never expose method-internal vocabulary (stronghold, outcome numbers, etc.) — describe what was found and what happens next instead."
2. Reword the recap instructions to plain equivalents, e.g. "present scope, the confirmed starting evidence, file path, proposed approach."

Either is a small diff to the skill markdown; happy to PR if maintainers agree on the direction.

## Environment

- BMad v6.8.0 (BMM module), installed via the BMAD installer
- Skill: `bmad-investigate`
- Observed with Claude Code as the agent runtime

bmadcode:
i think skills like this that can leak into common use should really be not auto firing unless very explicitly called. feel free to change it or improve it in any way.

PR #2478, opened 2026-06-17
fix(bmad-investigate): keep method-internal vocabulary out of user-facing output

mc856:
## What

Two related fixes in `bmad-investigate`, both addressing #2452:

1. **Stop method-internal vocabulary from leaking into user-facing output** — the Outcome 1 recap instruction told the agent to "present scope, stronghold, file path" verbatim, which in real sessions produced lines like "let me anchor the stronghold in code".
2. **Tighten the trigger description** so the skill only fires on explicit invocation, per the maintainer's note in the issue thread that "skills like this … should really be not auto firing unless very explicitly called. feel free to change it or improve it in any way."

## How

Three edits to `src/bmm-skills/4-implementation/bmad-investigate/SKILL.md`:

- **Frontmatter `description`** rewritten to match the house style used by `bmad-code-review`, `bmad-create-story`, etc. — quoted explicit invocation phrases (`"bmad investigate"`, `"open an investigation"`). Drops broad-paraphrase triggers like `"walk through unfamiliar code"` / `"build a mental model"`.
- **Communication principle** extended with a guardrail: user-facing chat uses plain English; skill-internal terms ("stronghold", "Outcome 1", "Hypothesis #3") stay in the case file and the agent's reasoning, not in chat.
- **Outcome 1 recap instruction** changed from `present scope, stronghold, file path` to `present scope, the confirmed starting evidence, file path` — the exact phrasing the issue author suggested.

## Notes

- **Internal vocabulary is intentionally preserved.** The `Stronghold first.` principle (line 32), the Outcome 1 heading, and the "find the stronghold" instructions in Outcome 1's body are unchanged. Per the issue author's preference, this PR fixes the chat-output leak via guardrail rather than renaming the method's internal language. Whether to rename `Stronghold first.` itself is a separate question, fit for a follow-up PR if you want it.
- **Trigger tightening is beyond the issue body's Suggested fix** — included here in direct response to the maintainer comment quoted above. Happy to split it into a follow-up PR if preferred.

## Verification

- `node tools/validate-skills.js --json src/bmm-skills/4-implementation/bmad-investigate` → `[]`
- `npm ci && npm run quality` → `Errors: 0` (23 unrelated `vi-vn`/`zh-cn` sidebar-order warnings)

Opening as draft to get a quick direction-check on the new trigger phrasing before polish.

### DIFF
--- a/src/bmm-skills/4-implementation/bmad-investigate/SKILL.md
+++ b/src/bmm-skills/4-implementation/bmad-investigate/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: bmad-investigate
-description: Forensic case investigation with evidence-graded findings, calibrated to the input. Use when the user asks to investigate a bug, trace what caused an incident, walk through unfamiliar code, or build a mental model of a code area before working on it.
+description: 'Forensic case investigation with evidence-graded findings, producing a structured case file an engineer can pick up cold. Use when the user says "bmad investigate" or "open an investigation".'
 ---
 
 # Investigate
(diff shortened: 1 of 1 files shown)


alexeyv:
The skill is gone, so this is now moot.

PR #2480, opened 2026-06-18
fix(installer): support ssh custom-source ports

### LINKED ISSUE #2476: Support SSH `--custom-source` URLs with custom ports for installer
**Describe your idea**

Add support for SSH protocol Git URLs with custom ports in `--custom-source`, for example:

```text
ssh://git@host:port/owner/repo.git
```

Currently, `--custom-source` supports scp-like SSH syntax such as:

```text
git@github.com:org/repo.git
```

but does not support the explicit `ssh://` URL form with a custom port.

**Why is this needed?**

Some self-hosted Git servers expose SSH on a non-standard port. In those environments, users need to provide the port directly in the Git URL when installing custom modules.

Without this support, users must rely on SSH config aliases as a workaround, which is less convenient for documentation, automation, CI, and shared installation commands.

**How should it work?**

`--custom-source` should accept valid SSH protocol Git URLs with optional custom ports, preserving the original clone URL so Git can clone it correctly:

```shell
npx bmad-method install --custom-source ssh://git@host:port/owner/repo.git
```

Expected parsing behavior:

```json
{
  "type": "url",
  "cloneUrl": "ssh://git@host:port/owner/repo.git",
  "cacheKey": "host:port/owner/repo",
  "displayName": "owner/repo",
  "isValid": true
}
```

This should be covered by a regression test in `test/test-parse-source-urls.js`.

**PR**

I can contribute an update. The change would update `CustomModuleManager.parseSource()` to handle `ssh://user@host[:port]/path/repo.git` URLs.

**Additional context**

The current parser handles:

```text
git@host:owner/repo.git
http://...
https://...
```

but `ssh://...` URLs currently fall through as invalid.

loicduong:
## What
Added support for `ssh://` Git URLs with optional custom ports in installer `--custom-source` parsing.

## Why
Self-hosted Git servers may expose SSH on non-standard ports, and users need to pass URLs like `ssh://git@host:2222/path/repo.git` directly without relying on SSH config aliases.
Fixes #2476.

## How
- Added explicit parsing for `ssh://user@host[:port]/path/repo.git` URLs while preserving the original `cloneUrl`.
- Derived `cacheKey` and `displayName` from the SSH URL path, including the custom port in `cacheKey`.
- Added a filesystem-safe cache path helper so custom-port cache keys work on Windows.

## Testing
Ran `npm run test:urls` to verify URL parsing regressions, including the new SSH custom-port cases. Also ran ESLint and Prettier checks on the changed files.


loicduong:
augment review

loicduong:
augment review

loicduong:
augment review

bmadcode:
Thanks @loicduong — supporting `ssh://` URLs with non-standard ports was a reasonable gap to close, and handling the `cacheKey`/`displayName` derivation rather than just the clone URL was the right level of care.

**BMAD is changing how it is distributed.** The project is moving to a flat `skills/` tree installed through `npx skills` and plugin marketplaces, and `tools/installer/` is removed entirely as part of that change. `tools/installer/modules/custom-module-manager.js` and `test/test-parse-source-urls.js` are both removed, so this can't land as written.

Custom module sources on non-standard SSH ports are still a legitimate use case. If the new install path doesn't handle them, please open a fresh issue against it and reference this PR — the parsing work here would be a good starting point.

---

PR #2488, opened 2026-06-20
fix(validate-refs): flag single-backslash Windows path leaks

### LINKED ISSUE #2487: [BUG] Absolute-path-leak check misses single-backslash Windows paths (C:\Users\...)
### Problem

`checkAbsolutePathLeaks` in `tools/validate-file-refs.js` is meant to catch absolute path
leaks (`/Users/...`, `/home/...`, `C:\...`) in skill and doc content, but it misses the most
common Windows form: a normal single-backslash path like `C:\Users\me\notes.md` is never
flagged.

### Cause

The pattern is `/(?:\/Users\/|\/home\/|[A-Z]:\\\\)/`. In a regex literal `\\` already matches
one backslash, so `[A-Z]:\\\\` requires **two** backslashes — it only matches the escaped form
`C:\\Users`. A real pasted Windows path has a single backslash and slips through. (`C:/Users`
gets caught, but only incidentally via the `/Users/` branch.)

### Repro

```js
const ABS_PATH_LEAK = /(?:\/Users\/|\/home\/|[A-Z]:\\\\)/;
ABS_PATH_LEAK.test('C:\\Users\\me\\notes.md'); // false  <- should be true
```

### Suggested fix

Match either separator: `[A-Z]:[\\/]`. I have a small branch with the fix plus a unit test
ready and can open a PR.

Environment: `main` (v6.8.0), `tools/validate-file-refs.js`.

zied-jlassi:
## Problem

The absolute-path-leak check misses normal Windows paths. A pasted `C:\Users\me\notes.md`
isn't flagged, while `/Users/...` and `/home/...` are.

## Cause

`ABS_PATH_LEAK` uses `[A-Z]:\\\\`. In a regex literal that's two backslashes, so it only
matches the escaped `C:\\Users` form — single-backslash Windows paths slip through.

## Fix

Widen the Windows branch to `[A-Z]:[\\/]` (drive letter + either separator). Also exported
`checkAbsolutePathLeaks` so it can be unit-tested, and added `test/test-abs-path-leak.js`
(wired into `npm test`) covering the single-backslash case plus the existing Unix /
forward-slash paths and code-block stripping.
Fixes #2487

## Testing

- `node test/test-abs-path-leak.js` -> 6/6 passed
- `node tools/validate-file-refs.js --strict` -> `Absolute path leaks: 0`, exit 0 (the wider
  pattern adds no new false positives on the repo's own content)


zied-jlassi:
Good catch on the lowercase drive letters — fixed in 3df821d.

I went with `\b[A-Za-z]:[\\/]` rather than a bare `[A-Za-z]:[\\/]`: without the word boundary the forward-slash branch also matches URL schemes (`https://` contains `s:/`), which flagged ~190 doc lines in a quick scan. The `\b` keeps real drive paths (`c:\`, `c:/`) while skipping `https://`. Added lowercase + a URL-not-flagged case to the test.

zied-jlassi:
Rebased onto the latest `main` to clear a merge conflict.

- **Conflict** was in `package.json`: `main` added `test:skills` to the `test` chain while this branch added `test:leaks` — resolved by keeping **both** (`… && test:skills && test:leaks && …`).
- Also adapted `test/test-abs-path-leak.js` to the `unicorn/prefer-string-raw` rule that `main` now enforces (the Windows path literals use `String.raw\`…\``; behaviour unchanged).

Full local run is green: `npm test` passes (refs/install/urls/channels/skills/leaks + eslint + markdownlint + prettier). CI is green here too. Ready for review whenever you have a moment 🙏

alexeyv:
Thanks @zied-jlassi — good catch, and the analysis in the description was spot on. Since this had drifted into conflict again, I landed the one-line regex change on its own in #2794 rather than ask for another rebase. Closing this one in favour of that; appreciate you finding and diagnosing it.

PR #2500, opened 2026-06-22
Add bmad-dev-auto unattended workflow

alexeyv:
## What
Adds `bmad-dev-auto`, a separate BMAD implementation skill for unattended development runs.

## Why
Quick Dev's core mechanics are useful for automation, but its human checkpoints and one-shot shortcuts do not fit an invocation-only workflow.

## How
- Adds a step-file workflow that plans, implements, reviews, repairs, and finalizes through result artifacts.
- Uses spec frontmatter and `## Auto Run Result` as the return contract.
- Blocks cleanly for missing intent, missing subagents, intent gaps, and unrecoverable review loops.

## Testing
Validated by the commit hook: refs, install tests, URL/channel tests, ESLint, markdownlint, and Prettier check.

## Release Notes

* **New Features**
  * Added a new automated development workflow that transforms small development intents into production-ready code with no human interaction.
  * Workflow automates planning, implementation, and review phases with integrated validation.

* **Documentation**
  * Added workflow guides and specification templates.
  * Updated command and workflow reference documentation.


pbean (dismissed):
Went through the changes, other than what i attach below when I did a review as well with claude, the only question i had are part of the dual review anyways. Is the third reviewer being excluded intentional? Other than that im good with it ive got a plan to work with it as its constructed.

dual review below:

Approving this. It's a clean, well-scoped primitive and the decoupling is exactly what we agreed on — no sprint-status, no env-var coupling, no result.json assumptions. The spec contract is genuinely strong: the
  three-tier Boundaries (Always / Block If / Never), the I/O & Edge-Case Matrix, <frozen-after-approval>, and the append-only Spec Change Log with KEEP instructions for re-derivation are all the right primitives.
  Routing by reasoning instead of regex (epic-story vs freeform) is what makes it actually general-purpose. I can drive it from the automator as-is, so none of the below blocks the merge — but a couple are worth
  fixing for any external driver, not just mine.

  Worth fixing regardless of us:

  1. on_complete only fires on the success path. Steps 02 and 03 "terminate cleanly" on a blocked outcome without ever reaching step-04's ## On Complete. So the one sanctioned "tell the outside world I'm done" seam
  is silent on every failure path. For a primitive meant to be orchestrated, I'd expect the completion hook to fire on all terminal paths. This is the highest-value one for me — it's the difference between a
  one-line completion handshake and having to also watch the spec on disk.
  2. The specLoopIteration cap looks broken. It's declared in step-04-review.md's frontmatter (specLoopIteration: 1), but the bad_spec loopback does step-04 → step-03 → "read fully and follow step-04-review.md" —
  re-reading the step file fresh each time, which re-initializes the counter to 1. So "if it exceeds 5 → blocked" may never trigger; the only thing bounding a review loop is context exhaustion. Storing the counter
  in the spec frontmatter (which persists across re-reads) fixes it.

  Worth considering:

  3. A small machine-readable result block. ## Auto Run Result being prose-for-an-LLM is fine, but any deterministic consumer has to regex free text. A tiny fenced block alongside the prose (status, blocked_reason,
  verification results) would let any program parse the outcome without weakening the human-readable narrative. Purely additive.
  4. The blocked signal lives in a couple of places — frontmatter status: plus prose when a spec exists, but a separate bmad-dev-auto-result-{slug-or-timestamp}.md when one doesn't. One canonical location would
  simplify consumers.
  5. Two hunters, no acceptance pass. Blind + Edge is a reasonable minimal review, but nothing verifies the diff against the spec's own acceptance criteria — a little surprising for a skill whose input is a spec
  full of ACs. Deliberate? (I'm covering it on my side, so not blocking — just flagging.)

  Minor: no empty-diff guard (a no-op session can still reach done); the name bmad-dev-auto inverts the ecosystem's bmad-auto-dev ordering, which might confuse; and project-context.md looks loaded twice
  (persistent_facts default + activation step 4).

  Net: ship it. Number 1 and 2 we'd probably genuinely want fixed for other automation as well; the rest are takes, not requirements.

PR #2509, opened 2026-06-25
chore(bmm): retire bmad-investigate skill and its documentation

### LINKED ISSUE #2452: bmad-investigate: method-internal jargon ("stronghold") leaks into user-facing communication

alexeyv:
## What

Retires the `bmad-investigate` skill in full. Removes the skill (`SKILL.md`, `customize.toml`, `case-file-template.md`), both EN/FR `forensic-investigation` explainer docs, and every integration point:

- Amelia's `IN` menu entry (`bmad-agent-dev/customize.toml`)
- the plugin marketplace manifest (`.claude-plugin/marketplace.json`)
- `module-help.csv`
- the workflow-map reference tables and HTML diagrams (EN + FR)
- the agent-trigger tables (`agents.md`, `named-agents.md`, EN + FR)

EN/FR explanation sidebar `order` values are renumbered to close the gap left by the removed doc. `npm run quality` → 0 errors.

## Why

I tested it. Four real "why" questions about a C++ codebase, each answered twice by identical agents — one using bmad-investigate, one just investigating normally.

Every pair landed on the same answer. The skill never got it more right. On one question the plain agent actually found more. The skill cost about 1.5× as much and took twice as long every time. One time out of a hundred where I need an investigation file, I can just ask the LLM to create it. 

In my experience, this gets invoked unexpectedly in response to random questions, burns some tokens and produces output that speaks in birds language. In the few acts that I knew of its existence, I never actually asked for it or needed it per se.

Github issues pointing the same way:
- **#2452** flagged the vocabulary leak, and a maintainer noted in that thread that skills like this *"should really be not auto firing unless very explicitly called."*
- **#2478** was an in-flight patch for the leak that deliberately left the internal vocabulary in place, deferring the deeper question.

## Supersedes / resolves

- Added in #2364.
- Supersedes the cleanup in #2478.
- Resolves #2452 by removal.

## Testing

- `npm ci && npm run quality` → **Errors: 0** (17 pre-existing `cs`/`vi-vn`/`zh-cn` sidebar-order warnings, unrelated).


hau1412:
Which skill should I use as a replacement? Thanks

anderewrey:
I'm really surprised by your conclusion. For sure the LLM model used has a huge impact on that.

romanek-adam-b2c2:
I'm very surprised this skill was retired. It's extremely valuable especially on very complex problems. Its case file template helps organise the process. I really doubt a hand-written prompt would achieve similar results for very complex problems. Such a shame :(

PR #2517, opened 2026-06-27
feat(elicitation): halt with clear error on missing/malformed registry

### LINKED ISSUE #2516: feat: fall back to built-in methods when the elicitation registry is missing or malformed
**Describe your idea**

**Why is this needed?**

**How should it work?**

**Additional context**

devilliersdutoit:
## What

Add a registry load guard to `bmad-advanced-elicitation` Step 1: HALT with a clear error if `methods.csv` is missing, empty, or malformed.

## Why

Step 1 loaded `methods.csv` with no failure path, so a missing/corrupted registry made the agent silently invent methods from memory. This fails fast with an explicit message instead.
Fixes #2516

## How

- Insert a natural-language HALT after the load step: unreadable / empty / wrong header / no rows → stop and report.
- Forbid inventing methods or proceeding from memory.
- Natural-language only; no code, no new control surface.

## Testing

- `npm run validate:refs`, `npm run validate:skills`, and `markdownlint` on the file: no new findings vs `main` (verified by diffing counts with/without the change).
- Confirmed the guard text matches the actual CSV header.

### DIFF
--- a/src/core-skills/bmad-advanced-elicitation/SKILL.md
+++ b/src/core-skills/bmad-advanced-elicitation/SKILL.md
@@ -35,7 +35,11 @@ When invoked from another prompt or process:
 
 ### Step 1: Method Registry Loading
 
-**Action:** Load `./methods.csv` for elicitation methods. If party-mode may participate, resolve the agent roster via:
+**Action:** Load `./methods.csv` for elicitation methods.
+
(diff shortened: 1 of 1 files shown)


bmadcode:
it might be better for the agent to invent and still perform the AE instead of failing

PR #2534, opened 2026-06-29
feat(elicitation): fall back to built-in methods on missing registry

### LINKED ISSUE #2516: feat: fall back to built-in methods when the elicitation registry is missing or malformed

devilliersdutoit:
## What

Replaces #2517 (closed). That PR halted Advanced Elicitation with an error when `methods.csv` was missing or malformed. Per maintainer feedback on #2517 — _"it might be better for the agent to invent and still perform the AE instead of failing"_ — this takes the opposite, graceful approach.

When `./methods.csv` cannot be read, is empty, or is malformed, the skill no longer halts. Instead the agent:

1. Falls back to a built-in default set of well-known elicitation techniques drawn from its own knowledge (e.g. Critique and Refine, First Principles, Pre-Mortem, Red Team vs Blue Team, Stakeholder Round Table, Tree of Thoughts).
2. Briefly notes to the user that the registry was unavailable and built-in defaults are being used.
3. Continues the normal flow.

## Why

A missing or corrupt registry should degrade gracefully, not block the user from running Advanced Elicitation at all. The agent already has deep knowledge of these techniques, so it can keep working.

Fixes #2516

## Scope

One paragraph added to Step 1 of `src/core-skills/bmad-advanced-elicitation/SKILL.md`. No code, no behavior change on the happy path.


devilliersdutoit:
Superseded by #2605.

#2603 rewrote `bmad-advanced-elicitation/SKILL.md`, so every section this PR touched (`FLOW`, `Step 1: Method Registry Loading`, `Step 3`) no longer exists and the branch is unmergeable. Rebuilt the same idea against the new `Serving the Catalog` structure in #2605 rather than untangling the conflict here.

Closing in favor of that one.

PR #2535, opened 2026-06-29
feat(review): add verification-gap reviewer as a third review layer

alexeyv:
## What

Adds a third code-review layer, **`bmad-review-verification-gap`**, alongside the existing adversarial bug-hunter (`bmad-review-adversarial-general`) and edge-case hunter (`bmad-review-edge-case-hunter`).

The two existing layers ask *"is this code wrong?"*. This one asks the orthogonal question: *"if the behavior this change introduces or alters stopped holding where it's actually used, would any test fail?"* It hunts for **verification gaps** — changed behavior that nothing would catch regressing, and sites that should adopt a new rule but don't — never for bugs.

## How it works

- **Cheap-by-default triage:** non-behavioral changes exit at near-zero cost; equivalence for a candidate-neutral change is established locally (the changed hunk plus any test already pinning the surface) without tracing consumers.
- **Mutation-style reasoning, no mutator:** name the one concrete regression a consumer would observe, then read the real test and ask whether any assertion would fail.
- **The named regression gates consumer qualification** — an untested downstream consumer the change doesn't actually reach is a place to look, not a finding.
- **No severity or confidence** on findings — the downstream triage owns scoring. Emits `No verification gaps found.` on a clean run so it isn't mistaken for a failed layer.

## Wiring

- Registered in `.claude-plugin/marketplace.json` (bmad-pro-skills).
- Added to the review step of **bmad-code-review**, **bmad-quick-dev**, and **bmad-dev-auto**, launched in parallel without prior context like the other layers.
- **bmad-code-review** triage normalizes its output, adds the `vgap` source, and treats the clean sentinel as a clean run rather than a failed layer.

## Not included (intentional)

- **Documentation** (`docs/**/core-tools.md`) is deliberately not updated yet — to follow.
- Empirical validation against real diffs is still pending; hence draft.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -29,7 +29,8 @@
         "./src/core-skills/bmad-editorial-review-structure",
         "./src/core-skills/bmad-index-docs",
         "./src/core-skills/bmad-review-adversarial-general",
-        "./src/core-skills/bmad-review-edge-case-hunter"
+        "./src/core-skills/bmad-review-edge-case-hunter",
+        "./src/core-skills/bmad-review-verification-gap"
       ]
     },
     {
--- a/src/bmm-skills/4-implementation/bmad-code-review/steps/step-02-review.md
+++ b/src/bmm-skills/4-implementation/bmad-code-review/steps/step-02-review.md
@@ -13,7 +13,7 @@ failed_layers: '' # set at runtime: comma-separated list of layers that failed o
 
 1. If `{review_mode}` = `"no-spec"`, note to the user: "Acceptance Auditor skipped — no spec file provided."
 
-2. Launch Blind Hunter and Edge Case Hunter in parallel without prior conversation context. If `{review_mode}` = `"full"`, include the Acceptance Auditor in the same parallel launch. If subagents are not available, generate prompt files in `{implementation_artifacts}` for each applicable reviewer role and HALT. Ask the user to run each in a separate session (ideally a different LLM) and paste back the findings. When findings are pasted, resume from this point and proceed to step 3.
+2. Launch Blind Hunter, Edge Case Hunter, and Verification Gap Reviewer in parallel without prior conversation context. If `{review_mode}` = `"full"`, include the Acceptance Auditor in the same parallel launch. If subagents are not available, generate prompt files in `{implementation_artifacts}` for each applicable reviewer role and HALT. Ask the user to run each in a separate session (ideally a different LLM) and paste back the findings. When findings are pasted, resume from this point and proceed to step 3.
 
    - **Blind Hunter** — prompt:
      > Invoke the `bmad-review-adversarial-general` skill on this diff:
@@ -22,6 +22,11 @@ failed_layers: '' # set at runtime: comma-separated list of layers that failed o
 
    - **Edge Case Hunter** — prompt:
      > Invoke the `bmad-review-edge-case-hunter` skill on this diff:
+     >
+     > {diff_output}
+
+   - **Verification Gap Reviewer** — prompt:
+     > Invoke the `bmad-review-verification-gap` skill on this diff:
      >
      > {diff_output}
 
--- a/src/bmm-skills/4-implementation/bmad-code-review/steps/step-03-triage.md
+++ b/src/bmm-skills/4-implementation/bmad-code-review/steps/step-03-triage.md
@@ -9,16 +9,9 @@
 
 ## INSTRUCTIONS
 
-1. **Normalize** findings into a common format. Expected input formats:
-   - Adversarial (Blind Hunter): markdown list of descriptions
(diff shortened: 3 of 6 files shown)


anthonyonazure:
Ran an empirical validation of this skill against real diffs, since the PR notes that was the missing piece. Sharing results in case they help take it out of draft.

## Setup

- **Corpus:** 7 real commits from a production HIPAA mental-health assessment platform (private repo; React 18 + TypeScript frontend, Azure Functions + Postgres API, ~60 clinical instruments). The repo's history gives known ground truth: several commits introduced regressions that demonstrably shipped undetected and were only fixed weeks later, one commit is the test-adding "pinning" commit, one is docs-only.
- **Protocol:** ground truth was written down before any run. Each run was blind: the SKILL.md verbatim plus the full `git show <sha>` as content, executed headless (Claude Code CLI, Sonnet, read-only tools) in a detached worktree at that exact commit, so the reviewer saw the repo as it was at review time, with no knowledge of later fixes. Every finding was then hand-verified against the historical code.

## Results

| Run | Commit type (ground truth) | Outcome |
|-----|---------------------------|---------|
| 1 | Introduced a date-comparison bug that shipped and was fixed 3 months later | Flagged the changed surface; also found 2 additional real, unpinned regressions in the same commit that we had not put in the rubric (both verified line-for-line) |
| 2 | Behavioral scoring fix across 4 call sites, landed with zero tests | 2/2 findings true, in the exact "drop the filter and nothing fails" shape; one included correct interaction reasoning between a new DB CHECK constraint and an untouched offline-sync consumer (regression would permanently strand patient data client-side) |
| 3 | Large security-hardening batch (authz guards across ~16 files), almost entirely untested | 12 findings; 6 spot-verified, all exact. Includes correctly detecting that an existing test file covered a *mirror* of the real guard function rather than the guard itself (the test file admits this in a comment; the reviewer read it and used it) |
| 4 | The "pinning" commit: extracted scoring logic + added tests | Ideal behavior: credited the new tests instead of flagging the pinned surface, and raised 2 missing-adoption findings for sibling call sites not routed through the new helper. One of them named a subset-scoring behavior the new helper lacked. That is exactly the refactor this repo actually made weeks later |
| 5 | Bug fix WITH good unit tests for the helper, but not for its call sites | The hardest discrimination in the panel, and it got it right: did not flag the well-tested helper, flagged the untested API-side change, and raised caller-path gaps of the form "revert the call site and every helper test still passes," explicitly citing the tests it read |
| 6 | Docs/tooling-only commit (control) | `No verification gaps found.` via the cheap triage exit, no manufactured findings |
| 7 | Contained a change that was already broken in production (queried a DB column that does not exist; the endpoint 500'd on every evaluation and was only caught by a manual re-hunt the next day) | Named the exact function and the exact missing verification ("no test exercises this evaluation path at all, and nothing verifies the column is populated"). Its chosen Demonstration was a weaker neighbor (NULL value rather than absent column, which a repo-only reviewer cannot know), but a test built from the finding would have caught the real failure |

## Cross-cutting observations

**Fabrication: none observed.** We hand-checked about a dozen "Existing test evidence" claims. Every "no test exists for X" was true; every cited assertion existed as described. The skill's fabrication taboo appears to hold under a weaker model than the skill was presumably authored with.

**False positives: none under our rubric.** Every verified finding was a genuine unpinned behavior change. The volume scales with reality (12 findings on the sweeping security batch, 0 on the docs commit), not with diff size.

**Weak-check classification works.** The corpus has Playwright smoke tests of the `getByText(/assessment|survey/i).toBeVisible()` kind. Runs consistently found them, read them, and correctly ruled them unable to catch the named regressions instead of counting "a test exists." One run also correctly detected that a test file covered a locally-defined mirror of the real guard function rather than the guard itself.

**Production corroborated one run in the strongest way possible.** Two of run 7's findings targeted a scheduled function as having zero verification on its changed paths. Unknown to the reviewer, that function had been throwing on every 15-minute invocation since inception (its target tables never existed) and nobody had noticed for months. The gap it reported was not hypothetical; reality had already run the experiment.

Two smaller notes, both possibly worth a line in the skill text:

1. **One Demonstration per surface.** Twice, a surface the reviewer had already flagged contained a second, historically real regression it did not name (it demonstrated a different one). Once a surface has one good Demonstration the enumeration stops. The suggested test would often catch the neighbor anyway, but if you want exhaustiveness per surface, Step 4 may need an explicit nudge.
2. **Prose outside the blocks.** 5 of 6 completed runs emitted a lead-in sentence ("I have all the information I need...") before the first finding block, and the clean run appended a justification paragraph after the sentinel line. If bmad-code-review's triage parses output mechanically, the OUTPUT FORMAT section may want a "your final message must contain only the blocks / only the sentinel line" sentence; headless runs seem prone to narrating.

## Bottom line

On this panel the layer does what the PR claims: it found real, later-confirmed verification gaps (including predicting an actual future refactor of this codebase), stayed silent where behavior was pinned, exited cheap on the non-behavioral control, and did not fabricate test evidence. It reads as ready to come out of draft from an empirical standpoint.

PR #2541, opened 2026-07-01
fix(bmad-help): resolve central config

### LINKED ISSUE #2455: [BUG] bmad-help does not read config.user.toml
### Description

The `bmad-help` command does not read to load or respect settings from `config.user.toml`.

When running `bmad-help`, the command seems to use default behavior instead of reading configuration values defined in the local  `config.user.toml` file.

### Steps to reproduce

1. Create or update a `config.user.toml` file with custom configuration values.
2. Run:

   ```bash
   bmad-help
3. Check the generated help output.
4. Observe that the output does not reflect the values from `config.user.toml`.

### Expected behavior

`bmad-help` should detect and read the relevant `config.user.toml` files, then apply the configured values when generating help output.

### Actual behavior

bmad-help appears to ignore `config.user.toml` and behaves as if no configuration file exists.

### Screenshots

_No response_

### Which module is this for?

### BMad Version

6.8.0

Claude Code

### Operating System

Windows

### Relevant log output

```shell

```

### Confirm

mc856:
## What
bmad-help now reads its config through the shared four-layer resolver instead of the legacy `config.yaml` / `user-config.yaml` files.

## Why
Fixes #2455. The Data Sources section still pointed at `config.yaml` / `user-config.yaml`, which current installs no longer produce — config now lives in the four TOML layers behind `_bmad/scripts/resolve_config.py`. So `communication_language` and `project_knowledge` set in `config.user.toml` never reached bmad-help.

## How
- Point the Config data source at `resolve_config.py`, the same shared resolver other skills use, and read the merged JSON
- Read `core.communication_language` and `modules.bmm.project_knowledge` (four layers, highest priority last)
- Invoke via `uv run`, consistent with the uv standardization in #2495

## Testing
Docs-only, single-line change to `bmad-help/SKILL.md` — no runtime code touched. `npm run quality` passes on the rebased branch: `test:install` (379), `test:urls` (68), lint / refs / skill checks clean.

## Summary by CodeRabbit

* **Documentation**
  * Updated configuration guidance to use the latest workflow for resolving merged settings.
  * Clarified the order of multiple configuration files and how to use the resolved output for key runtime values.


bmadcode (approved):

bmadcode:
good catch here @mc856 thank you

Issue #2533, opened 2026-06-29
[Feature Request] installer: user_name should be set only in config.user.toml and referenced in module configs


icedimych:
## Problem

During installation, `user_name` entered by the user is hardcoded into every module config file (e.g. `_bmad/bmm/config.yaml`, and potentially other `_bmad/*/config.yaml` files). This creates two problems:

1. **Team collaboration breakage**: When `_bmad/` is committed to a shared repository (which is the intended workflow), every teammate who installs bmad will have their own `user_name` baked into committed config files, causing constant conflicts.

2. **No single source of truth**: `_bmad/config.user.toml` already exists as the designated file for personal configuration (and is correctly gitignored). However, `user_name` is duplicated into other config files instead of being referenced from there.

## Expected behavior

The installer should:
- Write `user_name` **only** to `_bmad/config.user.toml`
- Replace hardcoded `user_name` values in `_bmad/*/config.yaml` with a reference to `config.user.toml` (e.g. `user_name: ${user_name}` or whatever templating syntax bmad supports)

This way:
- `_bmad/*/config.yaml` files are stable and safe to commit
- Each developer's personal identity lives only in their local `config.user.toml` (which is gitignored)

## Current behavior

After installation, `_bmad/bmm/config.yaml` contains:
```yaml
user_name: <value entered during install>
```

This value should instead be resolved from `_bmad/config.user.toml` at runtime.

## Related issues

Related to #2455 (bmad-help does not read config.user.toml) and #2456 (config.user.toml not gitignored on install) — all three point to `config.user.toml` not being treated as the authoritative source for user-specific settings.

## Environment

- BMad Version: 6.9.0
- AI IDE: Claude Code

jheyworth:
Put team name in for user name when the bead projects be used by multiple team members

icedimych:
> Put team name in for user name when the bead projects be used by multiple team members

Thanks for the suggestion! However, using a team name instead of a personal user name doesn't solve the core use case.

In a team workflow, it's important to track **who** did what over time — which developer worked on a specific feature, whose ideas shaped a particular decision, who created a story or task. This historical context is valuable for:
- Code reviews and retrospectives
- Understanding the reasoning behind architectural decisions
- Onboarding new team members who need to know who to ask about specific areas

[...]

PR #2542, opened 2026-07-01
fix(installer): ignore generated BMAD configuration

LakunleD:
## What
Added an installer-managed _bmad/.gitignore block that excludes generated configuration and personal override files while keeping team configuration commit-able. #2533 

## Why
Generated configuration contains personal values and changing timestamps that create unnecessary diffs and merge conflicts in shared repositories.

## How

- Added root-anchored ignore rules for generated config files.
- Preserved user-defined .gitignore content during installs and upgrades.
- Added tests for creation, upgrades, idempotency, and pattern safety.

## Testing
Ran node test/test-installation-components.js; all 388 tests passed.

### DIFF
--- a/test/test-installation-components.js
+++ b/test/test-installation-components.js
@@ -1901,6 +1901,52 @@ async function runTests() {
   console.log('');
 
   // ============================================================
+  // Test Suite 36b: Root .gitignore keeps generated config uncommitted
+  // ============================================================
+  console.log(`${colors.yellow}Test Suite 36b: Root .gitignore for installer-generated config${colors.reset}\n`);
+
+  {
+    const tempBmadDir36b = await fs.mkdtemp(path.join(os.tmpdir(), 'bmad-root-gitignore-'));
+
+    try {
+      const generator36b = new ManifestGenerator();
+      const gitignorePath = path.join(tempBmadDir36b, '.gitignore');
+
+      // First install: _bmad/.gitignore is created with root-anchored patterns
+      await generator36b.ensureRootGitignore(tempBmadDir36b);
+      assert(await fs.pathExists(gitignorePath), 'ensureRootGitignore creates _bmad/.gitignore');
+
+      const gitignore = await fs.readFile(gitignorePath, 'utf8');
+      assert(/^\/config\.user\.toml$/m.test(gitignore), '.gitignore excludes /config.user.toml (holds personal user_name)');
+      assert(/^\/config\.toml$/m.test(gitignore), '.gitignore excludes /config.toml');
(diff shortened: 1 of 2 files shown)


danielqb:
This is a solid fix for the committed-churn/personal-data symptom, and the root-anchoring to protect `_bmad/custom/config.toml` is exactly right.

Separately (not a blocker here, opened as its own issue so it doesn't scope-creep this PR): the per-module `_bmad/*/config.yaml` files this ignores aren't just personal/churny, they're never read back once `config.toml` exists — `loadExistingConfig` (`official-modules.js:880`) only falls back to them when neither toml file is present, and that's commented as the pre-v6 migration path. So the installer still generates N dead files every run under this PR; #2691 tracks stopping that generation as a follow-up.

One thing this PR's gitignore block does that I want to flag rather than assume is intentional: it ignores `/config.toml` itself, not just `/config.user.toml`. #2285 (which introduced the four-file split) documents `config.toml` as the *team* install-answers surface, distinct from the user-scoped file — so treating it as never-commit alongside the per-module files is a bigger scope change than "stop the personal-data leak," and worth a maintainer call on whether that's the intended direction vs. keeping `config.toml` committable and only ignoring `config.user.toml` + the per-module files.

LakunleD:
> This is a solid fix for the committed-churn/personal-data symptom, and the root-anchoring to protect `_bmad/custom/config.toml` is exactly right.
> 
> Separately (not a blocker here, opened as its own issue so it doesn't scope-creep this PR): the per-module `_bmad/*/config.yaml` files this ignores aren't just personal/churny, they're never read back once `config.toml` exists — `loadExistingConfig` (`official-modules.js:880`) only falls back to them when neither toml file is present, and that's commented as the pre-v6 migration path. So the installer still generates N dead files every run under this PR; #2691 tracks stopping that generation as a follow-up.
> 
> One thing this PR's gitignore block does that I want to flag rather than assume is intentional: it ignores `/config.toml` itself, not just `/config.user.toml`. #2285 (which introduced the four-file split) documents `config.toml` as the _team_ install-answers surface, distinct from the user-scoped file — so treating it as never-commit alongside the per-module files is a bigger scope change than "stop the personal-data leak," and worth a maintainer call on whether that's the intended direction vs. keeping `config.toml` committable and only ignoring `config.user.toml` + the per-module files.

Thanks. Honestly I'd forgotten this PR existed until you replied.

Ignoring `/config.toml` and not just `/config.user.toml` was intentional. It's installer owned and regenerated on every install.

bmadcode:
Thank you for this fix. Closing as out of sync with direction: as written it also gitignores the team-level `/config.toml`, which is a shared install-answers surface rather than a per-user artifact, and the v7 changes in progress will reshape how install config is handled. Issue #2533 stays open for the underlying need.

—

PR #2543, opened 2026-07-01
fix(dev-auto): require synchronous subagent invocation

pbean:
## Summary

The `bmad-dev-auto` workflow runs **unattended** — there is no event loop to resume a yielded turn. If a subagent is launched detached/backgrounded (e.g. `run_in_background: true`), or the turn ends to await a completion notification, control is never handed back and the run stalls.

This PR makes the subagent-invocation contract explicit:

- **`SKILL.md`** — adds a Subagents rule requiring every subagent to be invoked **synchronously** (launch, await within the same turn, then continue). Clarifies that running reviewers "in parallel" means several **blocking** calls awaited together in one turn — not detached execution. The only sanctioned way to end a turn remains the HALT protocol with an explicit terminal `status`.
- **`step-03-implement.md`** — updates the implementation hand-off to state the subagent must be invoked synchronously and awaited in the same turn (no `run_in_background`, no yielding to await a notification), resuming at "Tasks & Acceptance Verification" only after it returns.

## Test plan

- Docs-only change to workflow skill instructions; no runtime code touched.
- Pre-commit hook ran the full test suite + markdownlint — all passed.

### DIFF
--- a/src/bmm-skills/4-implementation/bmad-dev-auto/SKILL.md
+++ b/src/bmm-skills/4-implementation/bmad-dev-auto/SKILL.md
@@ -33,6 +33,8 @@ To HALT with a final status and optional blocking condition:
 
 Using subagents when instructed is mandatory. If you cannot, HALT with status `blocked` and blocking condition `no subagents`.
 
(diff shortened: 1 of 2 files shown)


alexeyv:
Maybe I'll find a better way to do this, but for now - merged

PR #2546, opened 2026-07-04
feat(bmad-module): add skill bmad detail design

PhanAnh1001:
## What

  Adds the new bmad-detail-design skill to BMAD-METHOD and wires it into the implementation workflow, including story creation, sprint flow, developer routing, and
  workflow documentation.

  ## Why

  This change makes detailed design a first-class BMAD step before development so stories consistently carry the design context needed for implementation. It also
  keeps the documented workflow, agent menus, and diagrams aligned with the actual skill behavior.

  ## How

  - Added the bmad-detail-design skill under src/bmm-skills/4-implementation/ and exposed it through the developer workflow surface.
  - Updated bmad-dev-story, story creation, sprint planning, and sprint status flows so detail design is required and referenced before development proceeds.
  - Refreshed workflow docs, agent references, and static workflow diagrams across the supported locales to reflect the new step.

  ## Testing

  Tested with npm ci and npm run quality on Node v22.23.1 with ASTRO_TELEMETRY_DISABLED=1. This includes the repo’s deterministic skill validation, docs validation,
  linting, install test, and full docs/site build.


alexeyv:
Not doing that. It's not generally a story-level concern, the whole Create Story/Dev Story/Code Review workflow is going away soon, and it should either be an optional bmad-spec capability, or something highly project-specific that a generic framework should not prescribe. Not entirely sure which, this would be a couple weeks' worth of work to do properly.

Issue #1472, opened 2026-01-30
Antigravity/OpenCode Integration: Agent Switching & Missing Workflows


cccczl:
## Overview

Multiple issues affecting Antigravity and OpenCode IDE integration.

### 1. Agent Switching Broken (#1472)
**Problem**: Unable to switch agents via tab key in OpenCode

**Steps to Reproduce**:
1. Open project in OpenCode
2. Try to switch agents using tab key
3. Switching doesn't work properly

### 2. Missing bmad-bmm-validate-brief Workflow (#1445)
**Problem**: `/bmad-bmm-validate-brief` command not found in Antigravity or OpenCode

**Error**: Workflow is not available/recognized

**Expected**: All BMM workflows should be available in all supported IDEs

## Environment
- IDEs: Antigravity, OpenCode
- BMad Version: Latest v6

## Root Cause
These issues suggest incomplete IDE integration for Antigravity/OpenCode:
1. Agent switching mechanism not fully implemented
2. Workflow discovery/registration incomplete

## Related Issues
- #1472 - Agent switching via tab key not working
- #1445 - Missing validate-brief workflow

dlukt:
I don't know how to dev-story now in opencode

vlcinsky:
It might relate to level of OpenCode support in BMAD. I noticed, that when using command:
```shell
$ bmad install
```
it asks for AI clients to prepare BMAD for and that OpenCode is not in top section (which are probably completely implemented) but on secondary one "Other". This is probably for a reason.

I am looking forward to seeing full OpenCode support.

Agent switching was working with alpha versions, disappeared with Beta.

dlukt:
on discord uh someone said that the issue has been identified and will be fixed soon

max-olotl:
I recently encountered a similar "Missing Workflows" issue in Antigravity (pertaining to all workflows) and found a specific root cause and manual fix.

Environment:

OS: macOS
IDE: Antigravity
BMAD Version: Latest (v6.0.0-Beta.4)

The Problem: During the npx bmad-method install process, I chose Windsurf as the IDE, assuming Antigravity followed the same configuration patterns. However, after installation, no /bmad- slash commands appeared in the Antigravity chat interface.

Root Cause Discovery: The installer placed all workflow and agent definitions in: {project-root}/.windsurf/workflows/
However, Antigravity requires these definitions to be located in: {project-root}/.agent/workflows/

[...]

Issue #1453, opened 2026-01-29
Missing Antigravity IDE during setup (npx bmad-method install)


wistant:
Describe the bug
During the setup process using npx bmad-method install, the Antigravity IDE option does not appear in the IDE selection step. This prevents users who rely on Antigravity from completing the configuration as expected.

Steps to reproduce

Run npx bmad-method install

Proceed through the initial configuration steps

Reach the IDE selection prompt

Observe that Antigravity is not listed among the available IDE options

Expected behavior
Antigravity should be available as a selectable IDE option during the installation process, allowing users to configure BMad with their preferred IDE.

Environment (if relevant)

Model(s) used: N/A

Agentic IDE used: Antigravity

BMad version: 4.44.3

Project language: N/A

Screenshots or links
Not available at the moment.

PR
Not yet. I can contribute a fix if needed.

Additional context
This issue occurs during the initial setup flow and blocks Antigravity users from completing the installation properly. The problem was observed after installing version 4.44.3.

PR #2551, opened 2026-07-05
feat: add Antigravity CLI (AGY) as supported installer platform

### LINKED ISSUE #2440: Add support for Antigravity CLI (AGY)
Hello,

I'd like to request to add Antrigravity CLI (aka AGY) to the CLI tools available when installing BMad.

Currently it only supports Antigravity IDE and Gemini. When installing both of them i don't have the available skills in AGY.

Here it what it looks at to see available skills:

`.agents/agents/{agent_name}/agent.json`

Would it be possible to add support for this CLI tool?

Thanks in advance.

bdsoha:
Adds the Antigravity CLI (`agy`) as its own selectable tool. The existing `antigravity` entry targets the IDE (`.agent/skills`, `~/.gemini/antigravity/skills`); the CLI is a separate surface that reads workspace skills from the shared `.agents/skills` standard and its global skills from `~/.gemini/antigravity-cli/skills` (per the CLI's `/skills` command). Different dirs, so the two installs never collide.

Installer is config-driven, so this is one `platform-codes.yaml` entry plus a test — no installer logic changes.

Verified `npm test` passes and `--list-tools` now shows `antigravity-cli`.

Closes #2440.

Related:
- #1453 — Antigravity missing from the installer picker
- #1472 — Antigravity/OpenCode agent switching & missing workflows
- Discussion #944 — Google Antigravity + Gemini 3.0 w/ BMAD
- Discussion #1752 — installing BMAD across Antigravity + OpenCode

### DIFF
--- a/test/test-installation-components.js
+++ b/test/test-installation-components.js
@@ -245,6 +245,14 @@ async function runTests() {
     const antigravityCliInstaller = platformCodes6b.platforms['antigravity-cli']?.installer;
 
     assert(antigravityCliInstaller?.target_dir === '.agents/skills', 'Antigravity CLI target_dir uses shared skills path');
+    assert(
+      antigravityCliInstaller?.global_target_dir === '~/.gemini/antigravity-cli/skills',
+      'Antigravity CLI global_target_dir uses the CLI-specific skills path',
+    );
+    assert(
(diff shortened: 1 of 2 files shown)


bmadcode:
thanks @bdsoha

PR #2558, opened 2026-07-05
bmad-brainstorming: add Idea Selector toolbar (select/filter/export)

paulpreibisch:
## Why

Right now a brainstorm.html keepsake is great to read, but there's no way to act on it — reviewing a session with 50-100+ ideas means scrolling back and forth with nothing to mark your favorites with. This adds a small floating toolbar to every generated keepsake so that while someone's reviewing the brainstorm, they can tick the ideas they like, filter the page down to just their picks to sanity-check the shortlist, and export that shortlist as a markdown file to carry into whatever comes next (a PRD, a follow-up session, a team discussion).

## What's in this PR

- New asset: `src/core-skills/bmad-brainstorming/assets/idea-selector-widget.html` — a single self-contained `<script>` block, no external dependencies, self-initializing. Adds:
  - A checkbox on every idea ("Mark as selected")
  - A **Selected** filter pill with a live count badge — click to show only ticked ideas
  - A **Clear** button to reset
  - An **Export** button that downloads everything selected as `selected-ideas.md`
  - A **?** help modal explaining all of the above
  - An auto-hiding top/bottom toolbar (reveals on hover near the screen edge, with a pin toggle)
- Updates `references/finalize.md` so building the HTML keepsake always: (1) tags each idea-level element with a `data-idea` attribute as it's built, and (2) appends the widget verbatim before `</body>` — making this a permanent, always-on part of every keepsake rather than something dependent on a session's particular CSS class names.
- Also documents a gotcha found while testing: if an idea element already has its own click behavior (e.g. a flip-card), that handler needs to ignore clicks landing in `.bv-ctrl` (the widget's injected checkbox row) so ticking the box doesn't also trigger the card's own interaction.

## Test plan
- [x] Validated the widget's JS parses cleanly (`new Function(...)` over the extracted `<script>` block)
- [x] Applied it to a real brainstorm.html (5 technique sections, 83 ideas, including click-to-flip cards) and confirmed: checkboxes appear on every idea, Selected filter shows only ticked ideas with the correct count, Export downloads a correctly formatted `selected-ideas.md`, and clicking a checkbox on a flip card no longer also flips it
- [ ] Maintainer review of toolbar placement/UX on a fresh session


paulpreibisch:
Superseded by #2559 — replaced the fixed floating-toolbar widget with selection controls built natively into each artifact's own visual theme, plus a real export-ideas.py script instead of client-side markdown generation. Sorry for the churn — this version is a clear improvement and better fits the skill's no-template philosophy.

PR #2577, opened 2026-07-11
fix(dev-auto): make followup_review_recommended converge

### LINKED ISSUE #2576: bmad-dev-auto step-04: the followup_review_recommended rule is structurally non-convergent — every review pass recommends another
## Summary

The `## Finalize` rule in `src/bmm-skills/4-implementation/bmad-dev-auto/step-04-review.md` sets `followup_review_recommended` from a criterion that **every review pass structurally satisfies**, so a driver that honors the flag re-reviews finished, verify-green work until it hits whatever outer budget it enforces. The recommendation never converges on its own.

## Mechanism

step-04 always runs the adversarial review layers and **auto-applies every `patch` finding** (Classify → `patch` — "Auto-fix. These are the only findings that survive loopbacks."). So essentially every pass "made review-driven changes."

The Finalize rule then evaluates *exactly those changes* to decide whether another pass is warranted:

> Base the judgment on the final pass's triage log and fixes, including patched-finding volume … **Many low-severity patched findings can be significant by volume.**

Because pass N's own patches are the input to the "is a follow-up warranted?" judgment, pass N's patches justify pass N+1 — which then makes its own patches, which justify pass N+2. There is no fixed point: any pass that patches a handful of low-severity findings can rationalize "significant by volume," and the next pass starts the same cycle. The only thing that stops it is the driver's outer review-cycle cap, at which point the loop terminates by exhaustion rather than by convergence.

## Evidence

Observed on a real project driven by an orchestrator (`bmad-loop`) with a hard `max_review_cycles = 3`:

- **10 of 10 stories** across 8 runs (2026-06-26 → 2026-07-11) ran **exactly** the full 3-cycle review budget — never fewer.
- From **round 1**, every round finished `status: done`, verify **green**, with **zero** `high` / `intent_gap` / `bad_spec` findings.
- Each round's *only* stated reason to continue was its own `followup_review_recommended: true`, citing that round's own review-driven patches ("this pass made review-driven changes … an independent pass is still worthwhile").
- Later rounds' findings were increasingly marginal or **self-inflicted by earlier review edits** — i.e. the loop was largely reviewing its own prior-round churn.

Cost per extra cycle: ~15–36 min of wall-clock plus heavy token spend, and deferred-work/ATTENTION noise — all on work that was already finalized and verified after round 1.

## Current wording (main, verified today)

`src/bmm-skills/4-implementation/bmad-dev-auto/step-04-review.md`, `## Finalize`:

> - Follow-up review recommendation: `true` when the final review pass made review-driven changes significant enough to benefit from an independent follow-up review; otherwise `false`. Use judgment, not a fixed numeric threshold. Base the judgment on the final pass's triage log and fixes, including patched-finding volume, consequence/severity, breadth, behavior/API/security/data impact, and implementation complexity. Many low-severity patched findings can be significant by volume. Do not recommend follow-up for only a few localized low-consequence fixes.

Note: the bullet has grown since this was first diagnosed — it now adds "Use judgment, not a fixed numeric threshold" and "Do not recommend follow-up for only a few localized low-consequence fixes." Those help at the margin, but the two structurally non-convergent phrases survive verbatim: the trigger is still "made review-driven changes significant enough," and "Many low-severity patched findings can be significant by volume" still lets any pass with enough small patches justify another. Since step-04 auto-applies patches on every pass, "only a few localized" rarely holds after the first round, so the carve-out doesn't break the cycle.

## Proposed replacement

Invert the default to `false` and gate `true` behind a *named, unverified* risk from *this pass's own* changes — so a follow-up pass that finds nothing new terminates instead of self-perpetuating:

> - Follow-up review recommendation: default `false`. Set `true` ONLY when the final pass's own review-driven changes carry substantive risk that nothing in this session then verified: (a) it patched a `high` finding, or (b) its patches changed behavior across files or module boundaries and no verification exercised the changed behavior green in this session. The fact that this pass made review-driven patches — however many — is NOT by itself grounds for a follow-up: `low`/`medium` patches whose verification ran green are settled work, regardless of volume. If this run is itself a follow-up review pass (the spec arrived with status `done`) and this pass's triage-log entry records zero `high`, zero `intent_gap`, and zero `bad_spec` findings, set `false` and record residual polish in the deferred-work ledger instead of recommending another pass. A `true` recommendation must name, in the Auto Run Result, the specific unverified risk the follow-up must re-examine; if none can be named, it is `false`.

This maps onto the file's existing structures: the `low`/`medium`/`high` severity tiers assigned in Classify, the `## Review Triage Log` entries, and the `intent_gap`/`bad_spec` categories.

## Note

`bmad-loop` is adding an orchestrator-side damping knob (`[limits] max_followup_reviews`, default 1) as belt-and-suspenders — it caps how many rounds a driver grants purely on the skill's own `followup_review_recommended`. But that's a per-driver workaround; the skill rule itself should converge for *every* driver that honors the flag, which is why this is filed upstream.

pbean:
## What

Replace the `Follow-up review recommendation` rule in `bmad-dev-auto` step-04 Finalize so `followup_review_recommended` converges: default `false`; `true` ONLY for (a) a patched `high` finding, or (b) first-run patches that changed runtime behavior across module boundaries — with the unverified risk named under `Auto Run Result`. A follow-up pass without a new `high` patch sets `false` and routes residual polish to the deferred-work ledger. Sync `spec-template.md` frontmatter comment and `docs/reference/dev-auto.md`.

## Why

Fixes #2576. Step-04 auto-applies every `patch` finding, and the old rule judged "significant review-driven changes" from those very patches — pass N's patches justified pass N+1 forever. Observed: 10/10 stories burned the full 3-cycle driver budget while verify-green with zero high/intent_gap/bad_spec from round 1.

## How

Deviations from the issue's proposed text: first-run-ness is folded into clause (b), so the follow-up case needs no precedence language; the follow-up detector is `final_revision` presence (durable across a crashed follow-up resumed at `in-review`, and `NO_VCS`) rather than "arrived with status `done`"; the "zero intent_gap / zero bad_spec" conditions are dropped (structurally guaranteed at Finalize — those routes never reach it); and (b) states patches are never re-verified rather than implying they "ran green" — step-04 runs no verification. Post-patch re-verification is deliberately out of scope: candidate follow-up issue.

## Testing

`npm run lint:md`, `npm run validate:skills`, `npm run docs:validate-links` green; pre-commit ran full `npm test` and the docs build.

### DIFF
--- a/docs/reference/dev-auto.md
+++ b/docs/reference/dev-auto.md
@@ -114,7 +114,7 @@ On successful completion, the workflow writes or updates the spec with:
   - Review findings breakdown
   - Verification performed
   - Residual risks
-- `followup_review_recommended` flag. True if LLM decided another review pass seems worthwhile. It's a suggestion, not a must. Simplest way to give it a second review pass is to re-run the skill pointing it at the spec file.
+- `followup_review_recommended` flag. Default false; true only when the final review pass leaves a specific unverified risk — a patched high-severity finding, or first-run patches that changed behavior across module boundaries — named under Auto Run Result. Patch volume alone never sets it, and a follow-up pass that patches no new high-severity finding sets it back to false, so repeated re-runs converge. It's a suggestion, not a must. Simplest way to give it a second review pass is to re-run the skill pointing it at the spec file.
 - `baseline_revision` and `final_revision` — HEAD before implementation and after the final commit. Together they bracket the run's commits: `git log baseline_revision..final_revision` lists exactly what it produced, and equal values mean no commits were made. Both are `NO_VCS` when version control is unavailable.
 
 If version control is available, the workflow commits the change. It does not push.
--- a/src/bmm-skills/4-implementation/bmad-dev-auto/spec-template.md
+++ b/src/bmm-skills/4-implementation/bmad-dev-auto/spec-template.md
@@ -4,7 +4,7 @@ type: 'feature' # feature | bugfix | refactor | chore
 created: '{date}'
 status: 'draft' # draft | ready-for-dev | in-progress | in-review | done | blocked
 review_loop_iteration: 0 # incremented by step-04 before each review loopback
-followup_review_recommended: false # set by step-04 on status: done from the final review pass significance judgment
+followup_review_recommended: false # set by step-04 at done; true ONLY for a named unverified risk (high patch, or first-run cross-module behavioral patches) — never from patch volume
 context: [] # optional: `{project-root}/`-prefixed paths to project-wide standards/docs the implementation agent should load. Keep short — only what isn't already distilled into the spec body.
 warnings: [] # optional: machine-readable warnings for orchestration, e.g. oversized, multiple-goals
 ---
--- a/src/bmm-skills/4-implementation/bmad-dev-auto/step-04-review.md
+++ b/src/bmm-skills/4-implementation/bmad-dev-auto/step-04-review.md
@@ -80,7 +80,11 @@ Prepare `Auto Run Result` details:
 - Summary of implemented change
 - Files changed with one-line descriptions
 - Review findings breakdown: patches applied, items deferred, items rejected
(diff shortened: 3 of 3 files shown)


pbean:
Closing as superseded: #2580 landed the scored-threshold approach (and per its description, one of the two should land, not both). The severity-weighted count over patch-triaged findings resolves the non-convergence from #2576 at the same root — every concern this PR carried is covered. Thanks for picking it up!

PR #2580, opened 2026-07-12
fix(dev-auto): score follow-up review recommendation; route patches to implementer

### LINKED ISSUE #2570: bmad-dev-auto: step-04-review.md launches the reviewer subagents without restating the synchronous-invocation ban
Downstream context: bmad-code-org/bmad-loop#109 — under bmad-loop, a `bmad-dev-auto` session backgrounded its implementation sub-agent despite the SKILL-level ban (recent Claude Code versions bias toward background `Agent` calls; the reporter measured the prompt ban losing ~20% of the time), then ended its turn to await the completion notification and the CLI exit stranded the sub-agent. bmad-loop now enforces this mechanically for the Claude CLI (`CLAUDE_CODE_DISABLE_BACKGROUND_TASKS=1`, bmad-code-org/bmad-loop#122), but the skill text is the only defense for other harnesses.

## Gap

`SKILL.md`'s **Subagents** section carries the ban ("Invoke every subagent **synchronously** … Never run a subagent in the background / detached / async (e.g. `run_in_background: true`), and never end your turn to 'await a completion notification.'"), and `step-03-implement.md` restates it at the point of use for the implementation subagent.

`src/bmm-skills/4-implementation/bmad-dev-auto/step-04-review.md` does not. Its Review section reads:

> Launch two subagents without prior session context.

"Launch … in parallel" phrasing plus a harness that biases toward backgrounding makes this the next most likely place a model detaches work, especially deep into a long session when the SKILL preamble has aged out of attention. The same section exists in `bmad-quick-dev/step-04-review.md`.

## Suggestion

Mirror step-03's restatement at the reviewer-launch site in both skills, e.g.:

> Launch two subagents without prior session context. Invoke both **synchronously** — several blocking calls awaited together in this turn, never backgrounded/detached (`run_in_background`), and never end your turn to await their results (see SKILL.md → Subagents).

Other subagent-spawning sites that also lack the restatement, if you want to sweep them in the same pass: `step-02-plan.md` ("Use subagents for deep exploration") and `step-01-clarify-and-route.md` (epic-context compile subagent).

### LINKED ISSUE #2576: bmad-dev-auto step-04: the followup_review_recommended rule is structurally non-convergent — every review pass recommends another

alexeyv:
## What

Three changes to the implementation skills' review step:

1. **Score the follow-up review recommendation instead of judging it** (`bmad-dev-auto` step-04). `followup_review_recommended` is computed from the final pass's findings triaged `patch` — never defer or reject: `true` if any was `high` severity, or if `3 × medium + 1 × low` reaches 5; otherwise `false`. The scoring lives only in step-04; the spec-template comment and reference docs describe the flag without restating the heuristic, so it can be recalibrated in one place.
2. **Route review patches to the step-03 implementer and re-verify** (`bmad-dev-auto` and `bmad-quick-dev` step-04). If the implementation subagent can be re-engaged with its context intact, step-04 sends it the patch findings — it has the freshest context and trivial fixes cost it almost nothing; falls back to fixing in the main session on platforms without persistent subagents. Either way, the spec's verification re-runs after patches land: previously, patches applied after step-03's verification shipped unchecked.
3. **Restate the synchronous-invocation ban at spawn sites** (both skills). Fixes #2570: the reviewer fan-out in step-04 now carries the no-backgrounding restatement, and the other spawn sites (epic-context compile, planning exploration) are marked synchronous.

## Why

Fixes #2576. The old rule asked the session to judge whether its own review-driven changes were "significant enough" to warrant another pass. Since step-04 auto-applies every `patch` finding, every pass made review-driven changes, so the judgment recommended another pass nearly every time and the flag never converged — observed stories burned the driver's full review budget while verify-green from round 1.

A severity-weighted count over the triage log the pass just wrote is arithmetic, not judgment: the usual two-or-three-low-patch pass terminates (score < 5), while a patched `high` or a pile of small patches still earns a follow-up. Counting only `patch`-triaged findings keeps pre-existing code quality (defer) and reviewer noise (reject) out of the signal. The flag stays a plain boolean suggestion; budgeting and damping are the orchestrator's job.

## Alternative to #2577

#2577 fixes the same non-convergence by making `true` conditional on a named unverified risk and ruling out patch volume as grounds entirely. This PR keeps the volume signal as a computed threshold. One of the two should land, not both.


pbean (approved):
besides the coderabbit follow-up looks good to me

PR #2583, opened 2026-07-13
fix(installer): preserve existing core config (#2435)

### LINKED ISSUE #2435: [BUG] installer: project_name (and any non-CLI-flag core key) dropped on update when one or more core flags are passed
### Description

When running bmad-method install --action update against an existing installation and supplying any of --user-name, --communication-language, --document-output-language, or --output-folder, the installer drops every existing [core] key that was not supplied via CLI flag. 
In particular project_name disappears entirely from the regenerated _bmad/config.toml and every _bmad/<module>/config.yaml, because the installer has no --project-name flag and the merge-with-existing logic reads from the wrong property.

### Root cause (proposed)
In tools/installer/ui.js — collectModuleConfigs(), branch A (any core-flag supplied):
```
// Seed core config from CLI options if provided
if (options.userName || options.communicationLanguage || ...) {
  const coreConfig = {};
  if (options.userName)     coreConfig.user_name     = options.userName;
  if (options.outputFolder) coreConfig.output_folder = options.outputFolder;
  // ...

  // Load existing config to merge with provided options
  await configCollector.loadExistingConfig(directory);
  const existingConfig = configCollector.collectedConfig.core || {};   // ← reads collectedConfig
  configCollector.collectedConfig.core = { ...existingConfig, ...coreConfig };
  // ...
}
But loadExistingConfig() in tools/installer/modules/official-modules.js populates this._existingConfig (the underscore-private member), not this.collectedConfig:

async loadExistingConfig(projectDir) {
  this._existingConfig = {};
  // ... reads each _bmad/<module>/config.yaml into this._existingConfig[entry.name]
  await this._hoistCoreKeysFromLegacyModuleConfigs();
  return foundAny;
}
```

So configCollector.collectedConfig.core is {} at the moment of merge, and the spread { ...existingConfig, ...coreConfig } produces a core block containing only the CLI-supplied keys.
In the --yes-with-no-CLI-flags branch, this is masked because that branch seeds defaults explicitly (including project_name: path.basename(directory)).

### Why this matters
There is no --project-name install flag, so on --action update runs there is no way for a user to preserve project_name once they supply any other core flag. The documented contract ("your prior answers are remembered as defaults") is violated. Teams using committed _bmad/ configs experience surprise diffs on every update run.

### Workaround used in our project

We pin project_name in _bmad/custom/config.toml — BMAD's documented "never overwritten" file — after the install. This works regardless of the bug because custom/config.toml is layered on top of installer-generated configs at read time.

### Steps to reproduce

In a directory that already has a complete BMAD install with a populated [core] block (including a custom project_name such as "My Cool Project"):

npx bmad-method@6.8.0 install \
  --directory     /path/to/project \
  --modules       core,bmm,bmb,cis \
  --action        update \
  --tools         rovo-dev \
  --user-name     "My Real Name" \
  --output-folder "{project-root}/_bmad-output" \
  --yes

### Expected behavior

project_name (and user_skill_level, etc.) are preserved at their pre-existing values because the installer documents itself as "your prior answers are remembered as defaults".

### Actual behavior

project_name is completely absent from the regenerated _bmad/config.toml and from every _bmad/<module>/config.yaml. Only the explicitly-passed-via-CLI keys (user_name, output_folder) survive.

### Screenshots

_No response_

### Which module is this for?

Not sure / Other

### BMad Version

6.7.1 6.8.0

Other

### Operating System

macOS

### Relevant log output

```shell

```

### Confirm

sanmaxdev:
## What
Preserve existing core configuration when an update supplies core command-line overrides. Add regression coverage for the update path.

## Why
The existing configuration loader stores prior values separately, but the override merge read the new collection instead. As a result, values such as `project_name` were dropped.

Fixes #2435

## How
- Merge supplied overrides on top of the loaded core configuration
- Cover preserved values and explicit replacements through the installer UI path

## Testing
- `npm ci`
- `npm run test:install`
- `npm run quality`

### DIFF
--- a/test/test-installation-components.js
+++ b/test/test-installation-components.js
@@ -3615,6 +3615,48 @@ async function runTests() {
   console.log('');
 
   // ============================================================
+  // Test Suite 49: CLI core overrides preserve existing config (#2435)
+  // ============================================================
+  console.log(`${colors.yellow}Test Suite 49: CLI core overrides preserve existing config${colors.reset}\n`);
+
+  let root49;
+  try {
+    const { UI } = require('../tools/installer/ui');
+    root49 = await fs.mkdtemp(path.join(os.tmpdir(), 'bmad-core-override-'));
+    const bmadDir49 = path.join(root49, '_bmad');
+    await fs.ensureDir(path.join(bmadDir49, '_config'));
+    await fs.writeFile(path.join(bmadDir49, '_config', 'manifest.yaml'), 'installation:\n  version: 6.10.0\n', 'utf8');
+    await fs.writeFile(
+      path.join(bmadDir49, 'config.toml'),
+      '[core]\nproject_name = "Existing Project"\ndocument_output_language = "French"\noutput_folder = "existing-output"\n',
+      'utf8',
+    );
+    await fs.writeFile(path.join(bmadDir49, 'config.user.toml'), '[core]\nuser_name = "Existing User"\n', 'utf8');
+
+    const { moduleConfigs } = await new UI().collectModuleConfigs(root49, ['core'], {
+      userName: 'Updated User',
+      communicationLanguage: 'Spanish',
+      documentOutputLanguage: 'English',
+      outputFolder: 'updated-output',
+      yes: true,
+    });
+
+    assert(moduleConfigs.core.project_name === 'Existing Project', 'CLI core overrides preserve existing project_name');
+    assert(moduleConfigs.core.user_name === 'Updated User', 'CLI core overrides replace explicitly supplied user_name');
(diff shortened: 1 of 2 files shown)


bmadcode:
Just saw this, thank you @sanmaxdev will review and merge after I get a chance to test it.

bmadcode:
Thank you for this fix and the investigation behind it. Closing because the underlying bug is now fixed on main: the override merge reads `configCollector.existingConfig.core` and layers CLI overrides on top (landed with #2601 and hardened by #2671), which is why this branch conflicts — the block it patched was rewritten. Equivalent regression coverage is carried by the open #2633.

—

PR #2585, opened 2026-07-14
fix(review): harden triage against over-broad dismissal and coverage gaps

alexeyv:
## What

Hardens the shared triage prompt in `bmad-code-review` (step-03) and its near-identical siblings in `bmad-quick-dev` and `bmad-dev-auto` (step-04 Classify) with three edits:

1. **Missing verification is a finding.** A confirmed coverage gap on changed behavior is rated and routed like any other finding (usually `patch` — the fix is the missing test), instead of being parked in a cannot-verify list.
2. **Dismissal evidence bar.** Dismiss/reject is allowed only when what was checked covers the finding's ENTIRE claimed trigger space; the rationale must state what was checked, and partial coverage keeps the remainder alive. Wrongly dismissing a real finding is worse than keeping a `low` one.
3. **Proof of impact only above `low`.** A mechanically real but undemonstrated cost/behavior change rates `low`; "the impact was not shown" is never a dismissal reason.

## Why

A triage A/B (same cached hunter outputs, same scaffold, only the step-03 method text varied) showed control routing is per-item noise: mean cared-set recall 0.42 across runs, with one control run dismissing even a finding it had kept before. The hardened text routed the full cared set 6/6 deterministically across three runs, with zero false-positive resurrections and ~7% token overhead. Each edit maps to a documented weakness observed across benchmark runs: coverage-routing ambiguity, over-broad dismissal rationale, and an unstated proof-of-consequence bar.

## Notes for review

- `bmad-code-review/steps/step-03-triage.md` adopts the validated text verbatim.
- The quick-dev and dev-auto ports adapt the dismiss-bucket edit to their `reject` vocabulary. The evidence bar is scoped to findings on the current change; the existing "when unsure between defer and reject, prefer reject" tie-break is kept for pre-existing issues so the deferred-work list doesn't accumulate speculative entries.
- Quick-dev and dev-auto still lack step-03's "read the code before rating" instruction (added to code-review in #2523); porting that is left for a follow-up to keep this change to the validated edits.


alexeyv:
The current version of this draft overcompensates.

bmadcode:
Thank you — closing this draft: it targets paths that no longer exist (`4-implementation/`, `bmad-quick-dev`, `bmad-dev-auto` — the tree is now `ship/` with `bmad-build`), and your own last comment noted the draft overcompensates. The concern lives on in merged #2697 and your #2758.

—

alexeyv:
The triage hardening represented here has since been implemented in the current review pipeline, so this older draft is no longer needed. Closing as completed elsewhere.

Issue #1962, opened 2026-03-13
Non-interactive install (--yes) prompts for directory and creates literal {output_folder}


jschulte:
**Describe the bug**
When running a non-interactive install with `--yes` and partial CLI options (e.g. `--user-name` without `--output-folder`), two bugs occur:

1. The installer still prompts interactively for the installation directory instead of defaulting to the current working directory
2. A literal directory named `{output_folder}` is created instead of `_bmad-output`, because the `output_folder` core config value is never set

**Steps to reproduce**
```bash
npx bmad-method install --modules bmm --tools claude-code,codex --user-name "Jonah" --yes
```

1. Installer prompts for directory despite `--yes` flag
2. After installation completes, a folder literally called `{output_folder}` exists in the project

**Expected behavior**
1. `--yes` without `--directory` should default to `process.cwd()` (current directory)
2. Core config defaults (`output_folder`, `communication_language`, `document_output_language`) should be backfilled from `src/core/module.yaml` when `--yes` skips interactive prompts

**Environment**
- BMad version: 6.0.1
- Agentic IDE used: Claude Code, Codex
- Node.js: v22+

**PR**
Fix incoming — see linked PR.

bmadcode:
Closing — duplicate of #1917 and fixed by the same change. `tools/installer/ui.js:708` now provides an `_bmad-output` default. Reopen if `--yes` still creates a literal `{output_folder}` directory on v6.4.0.

jbmml:
The bug still reproduces on **v6.6.0** (your close comment invited reopening if it did on v6.4.0+ — so flagging here).

## Reproduction

Controlled tests on a fresh project dir, bmad-method@6.6.0 installed globally:

| # | Install flags (in addition to `--yes --modules bmm --tools opencode --user-name Joey`) | Result |
|---|---|---|
| A | `--set core.output_folder=_bmad-output` | Creates literal `{output_folder}/` |
| B | `--output-folder _bmad-output` | Creates `_bmad-output/` ✓ |
| C | (none — relying on documented default `_bmad-output`) | Creates literal `{output_folder}/` |

Only the dedicated `--output-folder` flag works. The default-value path and the `--set core.output_folder=` path both still fail to substitute.

## Root cause (tools/installer/ui.js:745–797)

The fix added in #1963 is structurally gated:

```js
if (options.userName || options.communicationLanguage ||
    options.documentOutputLanguage || options.outputFolder) {
[...]

PR #2586, opened 2026-07-14
fix(installer): complete partial headless defaults

alexeyv:
## What

Complete omitted core configuration defaults when `--yes` is combined with partial CLI options. This prevents BMM artifact paths from retaining `{output_folder}` and creating a literal directory with that name.

## Why

The partial CLI-options branch bypassed the complete headless-default branch. A command such as `bmad install --yes --user-name Alex` therefore omitted `core.output_folder`. This is a regression of #1962.

## How

- Build the full headless core default layer whenever prompts are disabled.
- Merge defaults, persisted configuration, and explicit CLI values in precedence order.
- Cover fresh partial options, custom output paths, username defaults, and existing partial configurations.

## Testing

- Reproduced the literal `{output_folder}` directory on `main`.
- Re-ran the same installer command and confirmed `_bmad-output/planning-artifacts` and `_bmad-output/implementation-artifacts` are created instead.
- Ran `npm ci && npm run quality` successfully on the committed HEAD.

### DIFF
--- a/test/test-installation-components.js
+++ b/test/test-installation-components.js
@@ -3615,6 +3615,139 @@ async function runTests() {
   console.log('');
 
   // ============================================================
+  // Test Suite 49: partial headless core CLI defaults
+  // ============================================================
+  console.log(`${colors.yellow}Test Suite 49: partial headless core CLI defaults${colors.reset}\n`);
+
+  {
+    const { UI } = require('../tools/installer/ui');
+    const fixtureRoot49 = await fs.mkdtemp(path.join(os.tmpdir(), 'bmad-partial-cli-defaults-'));
+
+    try {
+      const ui49 = new UI();
+      const { moduleConfigs: partialConfigs49 } = await ui49.collectModuleConfigs(fixtureRoot49, ['bmm'], {
+        yes: true,
+        userName: 'CLI User',
+        communicationLanguage: 'French',
+        documentOutputLanguage: 'Spanish',
+      });
+
+      assert(partialConfigs49.core.user_name === 'CLI User', 'partial --yes preserves the explicit user name');
+      assert(partialConfigs49.core.communication_language === 'French', 'partial --yes preserves the explicit communication language');
+      assert(partialConfigs49.core.document_output_language === 'Spanish', 'partial --yes preserves the explicit document output language');
+      assert(
+        partialConfigs49.core.project_name === path.basename(fixtureRoot49),
+        'partial --yes defaults the omitted project name from the target directory',
+      );
+      assert(partialConfigs49.core.output_folder === '_bmad-output', 'partial --yes defaults the omitted output folder');
+      assert(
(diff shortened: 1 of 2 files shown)


alexeyv:
## PR Review: #2586

**Title:** fix(installer): complete partial headless defaults  
**Author:** @alexeyv  
**Branch:** `fix/partial-cli-defaults-output-folder` → `main`  
**Review layers:** Adversarial + Edge Case Hunter

### Findings

#### 1. Empty persisted values override valid headless defaults `[likely]` `[Both]`

**Severity:** Moderate

At `tools/installer/ui.js:854` and `tools/installer/ui.js:871`, object spread treats an existing empty `output_folder` as authoritative. Placeholder resolution later treats that empty value as missing.

A supported `--set core.output_folder=` invocation can create this state. A focused reproduction produced:

```json
"planning_artifacts": "{project-root}/{output_folder}/planning-artifacts"
```

This recreates the literal-placeholder failure the PR is intended to prevent.

**Suggested fix:** Filter invalid empty persisted values before merging, or use per-field fallback semantics. Add a regression test with `output_folder = ""`.

#### 2. `--set core.output_folder` is applied after derived paths and directories `[likely]` `[Adversarial]`

**Severity:** Moderate

`tools/installer/ui.js:789-854` parses `--set` but does not apply core overrides during configuration collection. Directories are created from the default-derived configuration at `tools/installer/core/installer.js:274`, while the override is only applied at `tools/installer/core/installer.js:335`.

Consequently, `--set core.output_folder=generated` creates and records BMM artifact paths under `_bmad-output`, then patches only the core setting to `generated`. The resulting core configuration, BMM paths, and filesystem disagree.

**Suggested fix:** Apply dependency-bearing core overrides before collecting module defaults, or reject these keys through `--set` and require their dedicated CLI flags.

### Summary

**Critical:** 0 | **Moderate:** 2 | **Minor:** 0  
**Sources:** 1 adversarial | 0 edge-case-only | 1 both  
**Verification:** `npm run test:install` passed all 409 assertions; all five GitHub quality checks are green at SHA `bc16d24`.  
**Binary files:** None.

_This review treats the separate directory-prompt behavior from #1962 as outside this PR's defaults-focused scope._

_Review generated by Raven's Verdict. LLM-produced analysis; findings may be incorrect or lack context. Verify before acting._

alexeyv:
Closing as obsolete. The headless-defaults fix this PR carried already landed on `main` via c2530ea5 (#2601), which builds the full core default layer whenever prompts are disabled and reads from `existingConfig` rather than `collectedConfig`.

Verified against current `main`: a fresh `--yes --user-name Alex` install now writes `output_folder: _bmad-output` and resolves the BMM artifact paths correctly — no literal `{output_folder}` directory. The empty-persisted-value case from the review comment above also no longer reproduces.

The remaining `--set core.output_folder` ordering issue (finding 2) is unrelated to this PR's defaults change and is still live on `main`; it needs its own fix.

PR #2605, opened 2026-07-19
feat(elicitation): fall back to built-in methods when catalog is unavailable

### LINKED ISSUE #2516: feat: fall back to built-in methods when the elicitation registry is missing or malformed

devilliersdutoit:
## What

`Serving the Catalog` documents no behavior for a failed `pick_methods.py` call, so the skill can stall at the exact moment it is meant to help. This adds a fallback: use well-known methods from built-in knowledge and keep going.

Rebuilt against the post-#2603 structure. Replaces #2534, which targeted the old `Step 1 / FLOW` layout and is now unmergeable.

## Why

Maintainer feedback on the earlier #2517: *"it might be better for the agent to invent and still perform the AE instead of failing."* This follows that, and mirrors the persona path in `Running a Method`, which already invents viewpoints when no agent fits.

Fixes #2516

## How

- One paragraph in `Serving the Catalog`, covering the three real failure modes verified against `pick_methods.py`:
  - missing methods file exits `2`
  - header-only catalog returns no rows (`categories` exits `0` with empty output)
  - mis-shaped file yields blank `method_name`/`description` rows at exit `0`
- Menu contract stays intact — reshuffle and `[a]` draw from the same built-in set, so no downstream step is left referencing a catalog that isn't there.
- `{workflow.preferences}` still honored; `{workflow.additional_methods}` used if readable.

## Testing

Exercised `pick_methods.py` against missing, header-only, and malformed catalogs to confirm the exit codes and output shapes above. Docs-only change, no script edits. `npm run validate:refs` shows one broken reference, pre-existing on `main` in `full-scan-instructions.md` and unrelated to this change.


devilliersdutoit:
Closing this — withdrawing the change for now.

PR #2622, opened 2026-07-24
fix(memlog): serialize concurrent writes

### LINKED ISSUE #2621: `memlog.py`: concurrent appends can lose entries
## Summary

`src/scripts/memlog.py` makes each individual replacement atomic, but the
complete read → modify → replace cycle in `cmd_append`, `cmd_set`, and
`cmd_init` has no inter-process lock. Concurrent writers can read the same base
state, and a later replacement can discard an earlier update. On Windows,
writers can also fail with `PermissionError` while sharing the fixed `.tmp`
path.

This matters when parallel agents or headless jobs journal to the same
`.memlog.md`.

## Reproduction

Run 50 concurrent `append` processes against one memlog on current `main`
(`bb45db4`) or `v6.10.0`.

Five consecutive runs on Windows 11 / Python 3.13 produced:

- 28–33 failed processes;
- only 17–24 of 50 entries present.

The current `test_memlog.py` suite exercises sequential behavior but not
concurrent writers.

## Expected behavior

All successful append requests should be preserved without corrupting or
silently replacing another writer's update.

## Proposed fix

Protect the complete read/modify/replace cycle with a cross-platform lock
acquired through `os.open(lock, O_CREAT | O_EXCL)`:

- `<memlog>.lock` holder metadata for diagnosis;
- bounded 10-second wait;
- no automatic stale-lock deletion;
- release by the owner in `finally`;
- a subprocess regression test covering concurrent appends.

With the proposed patch, five consecutive 50-process runs completed with 50/50
entries, zero failed processes, and no residual locks.

onbermejo:
## What

Serialize `init`, `append`, and `set` across processes with a standard-library lock and add a 50-process regression test.

## Why

Atomic replacement prevents partial files, but it does not protect the complete read/modify/replace cycle. Concurrent writers can lose entries or fail on Windows.

Fixes #2621

## How

- Acquire `<memlog>.lock` atomically with bounded waiting and holder diagnostics.
- Hold the lock across each complete mutation and always release it in `finally`.
- Launch 50 simultaneous subprocess appends and assert every entry survives.

## Testing

- `python -m pytest src/scripts/tests/test_memlog.py -q` (31 passed)
- Concurrency regression repeated five times (50 subprocesses per run)
- Base `main` Quality & Validation is green; local `npm test` reaches an unrelated Windows-specific `test:skills` failure


onbermejo:
Superseded by #2804 — the complete crash-safe fix (persistent coordination guard, stale-lock reclamation with a bounded lease, 42 concurrency/recovery tests) rebased on current `main`. It also addresses the earlier review notes about stale-lock replacement and deleting a replacement lock. Closing in favor of #2804.

PR #2629, opened 2026-07-26
feat(quick-dev): nail down implementation handoff to keep dispatch thin

alexeyv:
## What

quick-dev's `step-03` handed the spec to a coding subagent with loose prose ("Hand `{spec_file}` to a subagent and let it implement"), so the main session improvised a fat dispatch prompt each run — re-narrating the spec and re-expanding investigation detail at instruction weight. The review-layer subagents were already nailed down in `customize.toml`; the coding subagent was not.

This gives the coding subagent the same treatment dev-auto got in #2561:

- **`bmad-quick-dev/customize.toml`** — adds an `implementation_handoff` whose prompt is essentially *"Read the spec fully and implement it"* + guardrails + report-back format, sitting alongside the review layers.
- **`bmad-quick-dev/step-03-implement.md`** — injects that handoff verbatim (standalone block, matching how `step-04` injects review layers) and forbids parent-authored goal restatements, file lists, investigation detail, and CLAUDE.md/house-style rules in the dispatch.
- **`step-02-plan.md`** (quick-dev + dev-auto) — routes investigation findings into the spec's Code Map at planning time, so the thin handoff points at a rich spec instead of the dispatch prompt backfilling it.
- **`spec-template.md`** (quick-dev + dev-auto) — enriches the Code Map guidance to frame it as the distilled investigation map (symbol/line anchors, reuse pointers, read-only evidence).

## Why

The dispatch prompt is the subagent's privileged initial context — everything in it lands at instruction weight, diluting the load-bearing constraints among reference material. Moving investigation into the spec's Code Map at planning time also means step-04 reviewers and future sessions (which get spec-alone, no conversation context) see it too.

The intervention is write-time (what goes into the spec) more than dispatch-time: if the spec is rich, the thin handoff follows for free — which is what dev-auto already demonstrates.

## Verification

`npm run quality` passes (0 errors), including `validate:skills` and `validate:refs`. Confirmed `{workflow.implementation_handoff}` resolves and substitutes cleanly into the rendered `step-03`.


alexeyv:
Thanks @coderabbitai — reviewed both, declining both:

**Context loading when `implementation_handoff` is overridden (step-03, Major):** This is by design. The invariant is "the handoff loads the spec's `context:` files itself; the parent never pre-loads" — that's the contract. An override that replaces the recipe and drops the context-load line is the customizer breaking the recipe, no different from overriding a `review_layers` instruction into something incomplete; the skill doesn't defend against that. A non-overridable parent-side wrapper would force `context:` into the main session's window, which (a) doesn't help an override that drives an external tool via bash, and (b) reintroduces exactly the parent-side context expansion this change exists to remove. dev-auto's merged handoff (#2561) has the same property intentionally.

**Docs coverage (nitpick):** `docs/how-to/customize-bmad.md` documents only the baseline surface and explicitly states skill-specific fields self-document in the skill's `customize.toml` ("Read the workflow's customize.toml to see what's customizable"). The direct sibling `review_layers` (#2550) shipped under that convention with no dedicated docs page; `implementation_handoff` carries a full explanatory comment covering `{spec_file}` substitution and override semantics, so it already meets the repo's convention.

PR #2631, opened 2026-07-26
fix(installer): add missing readJson to fs-native shim

### LINKED ISSUE #2607: manifest.yaml records version: main for git-URL custom modules, even when the source is pinned with @<tag>
**Version:** bmad-method 6.10.0 (npx install path)

**Summary:** For custom modules installed from a git URL, `_bmad/_config/manifest.yaml` always records `version: main`, even when the source was pinned (`--custom-source https://github.com/owner/repo@v1.2.3`). The pin itself works: the clone cache's `.bmad-channel.json` correctly records `channel: pinned`, the tag, and the resolved SHA, and the install-time display line shows the real module version. Only the persisted manifest field is wrong.

**Repro:**
1. `npx bmad-method@6.10.0 install --directory . --custom-source "https://github.com/<owner>/<custom-module-repo>@<tag>" --tools claude-code --yes`
2. `grep -A2 "name: <module>" _bmad/_config/manifest.yaml` -> `version: main`
3. `cat ~/.bmad/cache/custom-modules/<repo>/.bmad-channel.json` -> `channel: pinned`, `version: <tag>`, correct `sha` (the parse worked)

**Cause (field mismatch):** in `tools/installer/core/manifest.js`, the custom-module branch of `getModuleVersionInfo` computes:
```js
version: resolved?.cloneRef || (hasGitClone ? 'main' : versionInfo.version),
```
but the URL parser in `tools/installer/modules/custom-module-manager.js` stores the parsed ref in `version` (`const version = versionSuffix || urlRef || null;` on the `type: 'url'` resolution object) -- `cloneRef` is never set on that path, so the expression falls through to the hardcoded `'main'`.

**Expected:** the manifest records the pinned ref (and ideally the resolved SHA, which is already known) -- e.g. read `resolved?.version` here, or set `cloneRef` where the URL is parsed.

santhiprakash:
## What
Adds the missing async `readJson` method to `tools/installer/fs-native.js` (the fs-extra drop-in shim) and a regression test. `fs.readJson(...)` was called by the installer but the method was never exported, so every call silently threw inside a try/catch.

## Why
`manifest.yaml` records `version: main` for pinned git-URL custom modules even when installed with `--custom-source <url>@<tag>`, because `CustomModuleManager.resolvePlugin()` reads `.bmad-source.json` via `await fs.readJson(...)` to stamp `cloneRef`/`cloneSha`/`rawInput` onto the resolved module. Since `fs-native` omits `readJson`, that call throws `TypeError: fs.readJson is not a function`, the surrounding `catch {}` swallows it, `cloneMetadata` stays `null`, and `manifest.js` falls through to the hardcoded `'main'` fallback. The clone itself is correct (`.bmad-channel.json` records the pinned ref + sha) — only the manifest field is wrong.

Fixes #2607

## How
- Added `async function readJson(p)` to `tools/installer/fs-native.js` (mirrors the existing `readJsonSync`/`writeJson` helpers, uses `node:fs/promises`) and exported it. This restores the fs-extra API contract the consumers rely on — same class of fix as #2257, which previously added other missing fs-native exports (`statSync`, `accessSync`, `readdirSync`, `realpath`, `rmdir`).
- This also unbreaks the two sibling `fs.readJson` call sites in `cloneRepo` (`custom-module-manager.js` L404 cache-version invalidation, L525 channel-marker fallback) that were silently no-op'ing for the same reason.
- Added `test/test-fs-native-readjson.js` (10 assertions): guards the missing export and asserts `resolvePlugin` stamps `cloneRef`/`cloneSha`/`rawInput` from `.bmad-source.json`. Wired into `npm test` and `npm run quality` via a new `test:fs-native` script so it runs in CI.

I intentionally did **not** touch `findModuleSourceByCode`'s re-discovery branch (which stamps only `repoUrl`, not `cloneRef`). That's a separate latent path that doesn't manifest in the reported `--custom-source` install flow (which goes through `resolvePlugin` first via `_resolveCustomSourcesCli`), and changing it is out of scope for this fix.

## Testing
- Reproduced the reporter's `--custom-source <git-url>@<tag> --yes` flow end-to-end against a local fixture on current `main` (`bb45db4a`): manifest recorded `version: main` before the fix; after the fix it records `version: v1.0.0`, `channel: pinned`, the resolved `sha`, and `rawSource` (`...@v1.0.0`).
- New `test:fs-native` fails without the fix (`cloneRef: undefined` for 7 of 10 assertions) and passes with it.
- Full `npm run quality` is green (format, eslint, markdownlint, docs build, and all installer test suites: 391 installation-component + 83 channel + 68 URL-parsing + 21 renderer + 7 file-ref + 3 skill + 10 fs-native).

Root-cause note: my first static-analysis pass pointed at the `findModuleSourceByCode` re-discovery branch (Path A); a runtime repro disproved that and isolated the real cause to the missing `readJson` export. The repro script and trace are summarized in the issue thread.


santhiprakash:
Merged current `main` and resolved the `package.json` conflict by keeping upstream's new `test:site-url` / `test:build-auto-contract` targets **and** retaining `test:fs-native` in both `quality` and `test`.

Verification on head `9a608a14`:
- `npm run test:fs-native` — 11/11 pass
- full pre-commit `npm test` suite — pass (lint, markdownlint, format:check included)

PR is MERGEABLE again; ready for maintainer review.

santhiprakash:
Closing: the source files modified by this PR have been removed or reorganized on upstream/main. The patch no longer applies.

Issue #2624, opened 2026-07-24
Harden activation: skills execute `{project-root}/_bmad/scripts/*.py`


atoomic:
## Summary

Every skill's **Step 1 on activation** shells out to a Python script located inside the *project* being worked
on, before anything has established where that script came from:

```bash
uv run {project-root}/_bmad/scripts/resolve_customization.py --skill {skill-root} --key workflow
```

Across `main` this pattern appears **101 times in 33 SKILL.md files**, over three scripts:
`resolve_customization.py` (51), `memlog.py` (40), `resolve_config.py` (10).

The skill has no way to tell an installer-written resolver from one a repository simply committed. There is no
hash, no version pin, no provenance check — only a path. So the contents of `_bmad/scripts/` are effectively
"whatever is in the working tree," and activation runs it.

## Why the plugin channel makes this sharp

`.claude-plugin/marketplace.json` publishes 6 plugins at 6.10.0. Every one of its 38 `skills` entries points at
a skill directory (`./src/core-skills/...`, `./src/bmm-skills/...`); **none points at `src/scripts/`**:

```json
{
  "name": "bmad-brainstorming",
  "source": "./",
  "version": "6.10.0",
  "skills": ["./src/core-skills/bmad-brainstorming"]
}
```

Only the npm installer places those scripts, via `_installSharedScripts()` in
`tools/installer/core/installer.js`, which syncs `src/scripts/*` → `_bmad/scripts/`.

So `/plugin install bmad-brainstorming@bmad-method` yields a **globally installed** skill whose first action is
to execute a script from `{project-root}/_bmad/scripts/` that the plugin never shipped. The skill is
org-/user-wide; `_bmad/` is per-repository. A plugin user who never runs the npm installer gets whatever the
open repository provides.

I recognise the marketplace path currently has no docs (I found no `/plugin install` guidance in the README or
`docs/`), so this may be an in-progress channel — but it is versioned in lockstep with releases, so it seemed
worth raising before it is promoted.

## Concretely

Clone a repository that has committed:

```
_bmad/scripts/resolve_customization.py     # prints valid resolver JSON on stdout, plus anything else it likes
```

Activate any BMAD skill. Step 1 runs it. Valid JSON on stdout means nothing looks wrong, so the session
continues normally.

Two things keep this below "vulnerability" in my reading, and I want to be straight about them rather than
overstate the case:

- Claude Code prompts before running Bash, so there is a gate. But this command appears at *every* activation
[...]

PR #2632, opened 2026-07-26
fix(skills): stop activation from executing project resolvers

rashidiff:
## What
Updated BMAD skill activation instructions so they no longer execute `resolve_customization.py` or `resolve_config.py` from `{project-root}/_bmad/scripts` during activation, and instead resolve customization/config by directly merging the documented TOML layers.

## Why
This removes the activation-time execution path for project-supplied resolver scripts, so `_bmad/` stays data-only during activation and skill startup no longer trusts arbitrary code from the working tree. (#2624)

## How
- Replaced activation-time `resolve_customization.py` calls in affected `SKILL.md` files with explicit TOML merge instructions using base, team, and user customization layers.
- Replaced activation-time `resolve_config.py` calls in affected `SKILL.md` files with direct four-layer config merge instructions.
- Kept non-activation resolver references such as `workflow.on_complete` untouched so the fix stays scoped to the reported issue.

## Testing
I did not add or run tests per your instruction. I did run `npm ci && npm run quality`, but the repo’s `docs:build` step fails on Windows by design because that pipeline explicitly does not support Windows.


rashidiff:
Conflict solved

rashidiff:
Ready to merge

bmadcode:
Thank you for raising this — the concern in #2624 (activation executing unverified project-local scripts) is real and worth tracking. Closing this PR in its current form: the conflict-resolution merge in early August dropped the core-skills coverage the original diff had, every edited file still carries the "if the script fails, resolve the block yourself" fallback that contradicts the new instruction, and the branch is conflicting with main again. More fundamentally, replacing the deterministic resolver scripts with in-chat TOML merging is a design change we are not taking right now — the v7 rework in flight will reshape how skills resolve config, and #2624 stays open as the record of the concern.

—

PR #2639, opened 2026-07-29
feat(implementation): prefer minimal planning approaches

alexeyv:
## What
Adds one concise implementation-selection heuristic to the planning step in Quick Dev and Dev Auto.

## Why
Both workflows already discover reuse points, but neither explicitly biases the planner toward suitable reuse, minimal new code, or investigating a bug's underlying cause. The guidance belongs in planning rather than generated specs or implementation handoffs.

## How
- Challenge whether new code is needed before selecting an approach.
- Prefer suitable reuse and the smallest maintainable change that satisfies clarified intent.
- Investigate underlying bug causes rather than planning only around reported symptoms.

Create Story and all spec templates remain unchanged.

## Testing
`HUSKY=0 npm ci && npm run quality` passed on committed `HEAD`, including renderer tests, reference validation, and all 48 skill validations with zero findings.


alexeyv:
Closing this. A/B tested the planning heuristic against real commits (3 tasks, 18 fresh planner runs, arms differing only in this sentence): no measurable effect on bug-shaped tasks — 12/12 runs in both arms produced the identical root cause, fix, and reuse citation. The one task where the arms diverged showed the guided arm also producing the widest-blast-radius plan. Not worth carrying. The branch also had merge-conflict markers committed in bmad-dev-auto/step-02-plan.md.

PR #2664, opened 2026-08-01
feat: support spec-folder epics in bmad-retrospective

alexeyv:
## What

Adds spec-folder epics to `bmad-retrospective` as a peer of sprint-status epics. A folder holding `SPEC.md`, an ordered `stories.yaml`, and `stories/<id>-*.md` artifacts — the shape an unattended build loop leaves behind — can now be selected, completeness-gated, resumed, and finalized without inventing sprint state.

- New `stories_status.py`: a read-only, JSON-only detector/inspector validated against `stories-schema.md` (list order authoritative, ids unique and prefix-free, exactly one artifact per id, `status` and revision range from frontmatter).
- Mode resolution in `SKILL.md`: a named folder selects stories mode over sprint status; a named epic number stays sprint mode; detection runs only when neither was supplied and there is no sprint status. Multiple candidates are never resolved silently.
- Stories mode gathers per-story revision ranges rather than one epic-wide range, writes and resumes `RETROSPECTIVE.md` in the spec folder, and never reads, creates, or writes sprint status.
- Source artifacts are hashed at selection and re-checked before finalization, so the read-only guarantee is verified rather than assumed.

Sprint-mode selection, evidence, verdict, and finalization behavior is unchanged.

## Why

An unattended build loop represents an epic as a spec folder, not as an entry in `sprint-status.yaml`. That shape previously could not be retrospected at all without synthesizing sprint state for it.

## Notes for review

This supersedes #2660, which delivered the same feature but rewrote the surrounding prompt prose wholesale. That rewrite deleted two behavioral rules that were each stated in two places — the absent-evidence guard (never render a verdict from the absence of a completeness check) and the rule that a gate must read the retrospective's frontmatter, since `sprint-status.yaml` records the retro key `done` whether the epic passed or failed. Both are preserved here.

The prompt diff is 72 lines across four files, and every one of the 12 deleted lines in the change is a line replaced by an extended version of itself.

`stories_status.py` is carried over from #2660 with four fixes: detection no longer offers folders inspection rejects (it now requires the same `SPEC.md` + `stories.yaml` + `stories/` triple), revisions containing `..` are rejected before they compose into a malformed range, and `inspect` returns explicitly rather than relying on `_emit` raising to avoid falling through into detection. Tests go from 27 to 32.

### DIFF
--- a/docs/explanation/retrospective.md
+++ b/docs/explanation/retrospective.md
@@ -5,7 +5,7 @@ sidebar:
   order: 15
 ---
 
-Run `bmad-retrospective` when an epic is done. It reads what the epic actually produced (the specs, the full diff, the per-story commits, the sprint status) and works from that evidence rather than anyone's recollection of how the work went. What comes back is a written review, a set of owned action items, and a verdict on whether the epic met its bar.
+Run `bmad-retrospective` at the end of an epic. It reads what the epic actually produced (the specs, the full diff, the per-story commits, the sprint status) and works from that evidence rather than anyone's recollection of how the work went. What comes back is a written review, a set of owned action items, and a verdict on whether the epic met its bar.
 
 ## What it does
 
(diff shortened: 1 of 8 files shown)


alexeyv:
Superseded: reimplemented as a prompt-only change. The inspector script and its test suite turned out not to be load-bearing for a review skill, so this version's 650 lines are replaced by 23.

PR #2672, opened 2026-08-02
Add bmad-ticket: deterministic ticket tree skill for the v7 planning lane

bmadcode:
## What

Adds **bmad-ticket**, the v7 deterministic ticketing skill (`src/bmm-skills/plan/bmad-ticket/`), replacing the v6 create-epics-and-stories / sprint-status pipeline for the planning lane, per the locked v7-spec-ticketing plan (D-1..D-36).

### The skill

- **Three routes, cheap exit first**: *refine* (one ticket, minimal ceremony), *slice* (open scope → detailed epic set, stops at epics), *incept* (one epic → its stories, just-in-time). Guided / quick / autonomous modes; inline Args surface with a side-effect-free `--help` path.
- **Ticket tree grammar**: epic = folder + `ticket.md` envelope, leaves = `KEY-n-slug.md`, parent = containing folder, one stored fact (`status`), generated identity-only `index.md`. Derived state (frontier, board, rollups, graph) is always computed, never written.
- **Epic done is intentional, never calculated** (D-23 revised): computed epic state is only not-started / in-progress; `done`/`dropped` are stored by a human call (retrospective or user), and only a stored done releases dependents.
- **Review-before-present**: `finalize_reviewers` (adversarial-general, edge-case-hunter, verification-gap) run over every full draft before the user — or the autonomous gate — sees it. Skeleton-first drafting; coequal-expert opening.
- **Archive-on-done** (D-36): when done lands on an epic, the skill offers to move its stories to `.archive/<date>-<slug>/` as the dated record (`--purge` deletes when the record lives elsewhere, e.g. Jira). Envelope stays as the durable layer; archived ids are never reissued.

### Deterministic layer (dual-homed: skill-local + `src/scripts/`)

- `update_ticket.py` — the frontmatter write gate: transition-graph status gating (`--force` = explicit user override, gibberish always refused), hitl derived upward from risk, cycle refusal, newline-injection refusal, duplicate-id refusal, atomic writes, archive hint on epic done.
- `ticket_tree.py` — `next-id`, `index`, `validate`, `list`, `frontier`, `board`, `coverage --proposed`, `graph --mermaid` (lanes + critical path), `render` (single epics-and-stories view), `archive`. Malformed files, duplicate ids, orphan leaves, and scalar list fields are detected, never silently swallowed.
- Byte-identity sync test keeps both homes honest; 60+ tests green in each.

### Integration + docs

- `bmad-spec` hands off all decomposition to bmad-ticket with the route named (stories.yaml removed).
- `bmad-create-epics-and-stories` → v6-shims: pitch-then-comply forward (JIT rationale; v6 single-file shape honored via slice → chained incept → `render`).
- `bmad-sprint-planning` legacy boundary: existing sprint-status.yaml flows stay supported; net-new work never feeds it.
- `module-help.csv`: bmad-ticket (TK) takes the row; prd/ux/architecture next-step pointers updated.
- New explanation doc: `docs/explanation/tickets-and-specs.md`.
- Full eval suite: 8 quality cases (autonomous + HITL), 8/8 trigger + 8/8 non-trigger queries.

### Known seam (deliberate, follow-up slice)

`bmad-build-auto` still reads `stories.yaml` and HALTs when it is absent. The build lane's move to the ticket tree is a separate v7 slice; until it lands, the v7 planning lane and the v6 build lane don't chain. Tracked in the v7 plan.

https://claude.ai/code/session_012KL7CNy8457v7uQ7Xkmor4


bmadcode:
Re CodeRabbit's suggestion that `coverage` should return `ok: false` when `uncovered` is non-empty: declining deliberately. Across every verb in `ticket_tree.py`, `ok` means "the command ran and the output is trustworthy" — an uncovered requirement is a *finding*, not a failure, and callers (the skill's routes and the eval rubrics) read the `uncovered` array for the verdict. Overloading `ok` would make a successful coverage report indistinguishable from a broken run.

All other actionable findings from this round are addressed in 12c411aa: file-mode preservation + fsync in the atomic writers, frontmatter-scoped `_rewrite_depends_on`, zero-indent block-sequence parsing, unknown-type validation, `hitl_threshold` type-checking with a top-level JSON-contract catch, risky-scalar quoting, encoded index links, and the sync test asserting instead of skipping — each with a regression test.

PR #2673, opened 2026-08-02
fix(review): close two remaining verification-gap weak-check gaps

eugeniawang:
## Summary

Field-tested the verification-gap lens across 5 runs on real production diffs (submitted as feedback on Discord, 2026-07-02, back when it lived at `src/core-skills/bmad-review-verification-gap/SKILL.md` on `feat/verification-gap-reviewer`). Most of that feedback has since been addressed independently through the skill's evolution into a `bmad-review` lens (search-by-symbol/import-reference gating on absence claims, showing the search, treating config/data files as behavioral in Step 2). Two concrete gaps from that testing are still open in the current `lens-verification-gap.md`:

- **Fallback-masked assertions don't count as coverage.** A pattern we hit in a real authz diff: `expect(record().owner_id ?? OWNER).toBe(OWNER)` passes whether `owner_id` was stamped correctly or never set at all — the `??` fallback happens to equal the expected value, so the assertion can't fail on the regression it looks like it's guarding. Added to the "these do not count" list in Step 4.
- **Evidence-mechanism phrasing guardrail.** In one run the lens correctly found a real gap but explained *why* the test missed it in terms of file/execution order — which is never actually true under per-test isolation (mocks/monkeypatches are per-test, not ordering-dependent) and undermines an otherwise-correct finding. Step 5 now requires the explanation to be grounded only in what the test itself sets up and asserts.

Also added one line recommending the strongest available self-check when tooling allows it: apply the Demonstration as a real mutation and run the affected tests — a genuine gap stays green, a false one turns red.

## Test plan
- [ ] Docs-only change to `lens-verification-gap.md`; no code path affected
- [ ] Re-run the lens on a diff containing a `??`/`||`-fallback-masked assertion and confirm it's now flagged as not counting

### DIFF
--- a/src/core-skills/bmad-review/references/lens-verification-gap.md
+++ b/src/core-skills/bmad-review/references/lens-verification-gap.md
@@ -54,7 +54,7 @@ Find and read the relevant test. Ask whether the Demonstration would make an ass
 - For a regression-style Demonstration: if no test runs the path, the test is skipped/flaky/not run normally, or the test runs the code without checking the changed result, report a `Regression gap` or `Broken-verification gap`.
 - For a qualifying Missing-adoption case: if none of the site tests you found assert it adopts the new behavior, report a `Missing-adoption gap`.
 
-A test counts only if it runs normally and an assertion observes the changed output, branch, or contract. These do not count: no execution; success/no-throw/snapshot-only checks; mock/log-call checks; human-only checks; tests that mock away the integration; e2e tests that pass through without checking the changed output; stale assertions or fixtures.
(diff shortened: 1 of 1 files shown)


alexeyv:
Thank you very much for testing this lens on real changes and taking the time to describe what went wrong. This is exactly the kind of feedback that helps improve these prompts, and #2683 exists entirely because of your work.

I came to a smaller patch because prompt language does not stay inside the prompt. It leaks into reviews and interactive conversations. Jargon makes the result harder to understand, metaphor makes simple instructions less direct, and overprompting makes the model carry rare cases through every run. For a shared prompt, every permanent sentence needs to earn that cost.

The fallback case seems useful enough to keep as one short example. It is a plausible recurring mistake, and an example may help the model notice it without creating a new category of rules.

Mutation testing is a strong check, but its cost varies enormously between projects. This PR does not yet show that it changes enough findings to justify adding it to the shared lens. Projects that benefit from it can add it through their own workflow or customization.

The final observation is also useful, but I think the positive instruction is enough: explain why the test misses the bug using what the test sets up and checks. The broader prohibition on test-order explanations is unnecessary and can be wrong when tests really do share state.

I tried to preserve the value of your findings while adding the least possible prompt text. Thank you again—I would genuinely value your feedback on whether #2683 retains what you observed.

PR #2674, opened 2026-08-03
feat: bmad-project-context skill — verified kernel + bundle context system

bmadcode:
## What

Replaces `bmad-document-project` and `bmad-generate-project-context` with one skill, `bmad-project-context`: an engine with three intents (**ingest** / **query** / **audit**) that builds and maintains a project's context system — a tiny always-loaded **kernel** plus a **bundle** of small verified knowledge entries, instead of generated documentation.

Governing thesis, backed by measurement: generated docs make agents worse; a curated minimum of verified, non-derivable truths makes them better.

## Highlights

- **Kernel + bundle contracts** — instruction-budgeted kernel (pruning test: would removing this line change agent behavior?), small frontmattered entries with trust metadata (`verified` asserts a human in the loop; auto mode marks everything `generated`), `index.md` as sole entry point, compass files for monorepo areas.
- **`context.py` core runtime script** — measurement, mutation, and resolution only: `validate` (measured budget checks), `index`, `sweep` (staleness/missing paths), `resolve` (cross-project ladder: self → workspace → cache → sparse git fetch), `compass`, `sync` (consent-gated managed blocks in AGENTS.md with byte-preserved surroundings and link re-anchoring), `bootstrap` (standalone self-install), `config` (delegates to the installed BMad resolver so script and session can never disagree). Discovery is deliberately NOT scripted — the model uses its own tools and subagent fan-out.
- **Placement contract** — `context_placement: bmad | agent-files | both`; works with a full BMad install or standalone in any repo.
- **Interactive default, auto/headless mode** — chunked interrogation rounds interactive; headless accepts intent/scope/placement/bundle-root, logs every inference as an assumption, ends with a JSON status.
- **Old skills husked** — both reduced to 10-line deprecation shims forwarding here; docs updated with deprecation notes kept for old-name searches; analyst menu `DP` → `PC`.

## Testing

- 46 pytest tests for `context.py` (frontmatter, index, sweep, resolver against local bare repos, sync link-rewriting, config layering, bootstrap, skill-copy byte parity).
- 14 judged behavioral evals (greenfield, brownfield, refresh, headless, monorepo, workspace resolve, standalone, audit) — all passing; eval-found defects (link re-anchoring, trust laundering, filesystem crawling) fixed and locked with rules.
- Validated on a real brownfield repo end-to-end.
- `npm test` suite green (runs in the pre-commit hook).

## Notes

- Translated docs (cs/fr/zh-cn/vi-vn) still reference the old skills — needs the translation pipeline pass.
- Trigger evals are authored (`evals/triggers.json`) but not yet executed; eval-runner trigger mode is ticketed separately.


bmadcode:
Review triage (commit 97523c73):

**Fixed**
- Eval fixture files silently dropped by the repo .gitignore are now force-added: `fixture-brownfield/pnpm-lock.yaml`, `fixture-brownfield/_bmad/context.yaml`, `fixture-refresh/context/.memlog.md`. Good catch — without them the brownfield and refresh evals lose their planted state on a fresh clone.
- `docs/reference/agents.md` Analyst row updated: `DP`/Document Project → `PC`/Project Context (the note was updated but the table row was missed).
- `context.py`: `cmd_index` no longer raises IndexError on an empty `index.md` (an empty file is now overwritable, not foreign); inline `#` comments in frontmatter values are stripped only when preceded by whitespace, so values like `C#` survive; `cache_lookup` returns None on a corrupt/truncated pointer file instead of raising; the pointer is written atomically via os.replace.
- `triggers.json`: added a positive trigger for the query intent.

**Skipped, with reasons**
- *Implement the documented Stripe webhook contract in the fixtures* — the mismatch between the doc and `webhooks.ts` is deliberately planted drift; the eval exists to test that the skill detects it. "Fixing" the fixture would destroy the test.
- *Validate `[[project:entry]]` links in `cmd_validate`* — cross-project links resolve through `context.py resolve` (the resolution ladder), not the local validator, by design: the same link must work when the sibling project isn't checked out.
- *sparse_fetch timeout / rev-parse validation / workspace-walk bounding* — real, but they belong to the resolver hardening pass (forges, auth, cache invalidation, offline matrix) which is deliberately deferred from this PR; the resolver ships as a light prototype.
- *tomllib skip on Python 3.10* — acknowledged; the TOML chain degrades gracefully to the YAML fallback at runtime, and CI runs ≥3.11.

PR #2414, opened 2026-05-23
fix: write customization JSON as UTF-8

cyphercodes:
## What
Forces `resolve_customization.py` to emit its JSON stdout as UTF-8 and adds a regression test for emoji icons under a cp1252 stdout.

## Why
Windows cp1252 stdout cannot encode emoji icons from `customize.toml`, causing `UnicodeEncodeError` before the agent JSON reaches callers. Fixes #2406.

## How
- Reconfigure stdout to UTF-8 before writing JSON while keeping `ensure_ascii=False`.
- Add a stdlib unittest that runs the resolver with `PYTHONIOENCODING=cp1252` and verifies the emoji is emitted as UTF-8.

## Testing
- `python3 -m unittest src/scripts/tests/test_resolve_customization.py`
- `npm ci && npm run quality`
- `git diff --check HEAD~1..HEAD`

Issue #2406, opened 2026-05-21
[BUG] resolve_customization.py crashes on Windows cp1252 — UnicodeEncodeError on emoji icons


gustavozandarim:
### Description

When activating any BMad agent that has an emoji icon in its `customize.toml` (e.g. `🧭`, `🔍`, `📐`, `✍️`), `resolve_customization.py` crashes immediately on Windows machines with a `UnicodeEncodeError`.

This happens because `sys.stdout` defaults to cp1252 on Windows, which cannot encode emoji characters. The script uses `json.dumps(..., ensure_ascii=False)` — which correctly preserves emojis in the string — but then `sys.stdout.write()` fails when trying to output them.

**Environment:**
- OS: Windows (cp1252 default console encoding)
- Python: 3.14.3
- BMad version: 6.7.1
- Tool: Claude Code

### Steps to reproduce

1. Install BMad on a Windows machine with default cp1252 console encoding
2. Install any module that uses emoji icons in agent `customize.toml` files (e.g. the OAM module with agents using `🧭`, `🔍`, `📐`, `✍️`)
3. Open Claude Code in the project folder
4. Invoke any of those agents via `/agent-name` or by name
5. The agent attempts to run `resolve_customization.py` — crash occurs immediately

### Expected behavior

The resolver runs cleanly on Windows, outputs the agent JSON with emoji icons preserved, and exits with code 0 — with no environment variable workaround required.

### Actual behavior

The script crashes with the following traceback:

```
Traceback (most recent call last):
  File "_bmad\scripts\resolve_customization.py", line 230, in 
    main()
  File "_bmad\scripts\resolve_customization.py", line 226, in main
    sys.stdout.write(json.dumps(output, indent=2, ensure_ascii=False) + "\n")
  File "encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\U0001f9ed'
in position 26: character maps to 
```

The agent then improvises a retry using `PYTHONIOENCODING=utf-8` or `PYTHONUTF8=1`, which works but adds latency and depends on agent improvisation on every single activation.

**Proposed fix** — add at the beginning of `main()` in `resolve_customization.py`, before any stdout write:

```python
if sys.platform == "win32":
    sys.stdout.reconfigure(encoding="utf-8")
```

[...]

PR #2684, opened 2026-08-04
fix(party-mode): decode subprocess JSON as UTF-8 on Windows

kodesh87:
## What
Decode resolver subprocess stdout as UTF-8 in `resolve_party.py` so Party Mode can read emoji agent icons on Windows cp1252 locales.

## Why
Fixes follow-up to #2406. After #2414, `resolve_customization.py` writes UTF-8 JSON (including emoji icons) to stdout when piped. On Windows with cp1252 locale, `resolve_party.py` `_run_json()` still called `subprocess.run(..., text=True)` without `encoding=`, causing `UnicodeDecodeError` in the reader thread when Party Mode activates.

## How
- Pass `encoding="utf-8"` to `subprocess.run` in `_run_json()`
- Guard empty stdout with `stdout = out.stdout or ""`
- Catch `UnicodeDecodeError` alongside other subprocess failures
- Add regression test pinning UTF-8 decoding

## Testing
- `python src/core-skills/bmad-party-mode/scripts/tests/test_resolve_party.py` — 17 tests pass
- Repro context: Windows + cp1252 + agents with emoji icons in `customize.toml`

### DIFF
--- a/src/core-skills/bmad-party-mode/scripts/resolve_party.py
+++ b/src/core-skills/bmad-party-mode/scripts/resolve_party.py
@@ -46,13 +46,16 @@ except ImportError:  # pragma: no cover - guarded for <3.11
 def _run_json(cmd):
     """Run a resolver script and parse its JSON stdout. None on any failure."""
     try:
-        out = subprocess.run(cmd, capture_output=True, text=True, timeout=60)
-    except (OSError, subprocess.SubprocessError):
+        out = subprocess.run(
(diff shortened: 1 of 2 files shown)


alexeyv:
Thanks @kodesh87 for the submission and for identifying this Windows UTF-8 decoding issue. Your contribution inspired #2687, which applies the fix as the minimal explicit UTF-8 decoding change while preserving the existing exception behavior. We have merged that PR, so I’m closing this one in its favor. Thanks again!

Issue #2319, opened 2026-04-26
[BUG] Bmad is not working on OpenCode CLI


m-elhabib-dev:
### Description

https://github.com/bmad-code-org/BMAD-METHOD/issues/818#issuecomment-4321923615

### Steps to reproduce

Install bmad with normal installation guide

### Expected behavior

to works in opencode cli like other IDEs

### Actual behavior

I just installed bmad and selected opencode and antigravity, the bmad commands works on antigravity but in opencode cli are not appear, i tried the GUI opencode and it works, so please fix it for opencode cli users.

### Which module is this for?

BMad Method (BMM) - Core Framework

### BMad Version

6.5.0

### Which AI IDE are you using?

Other

### Operating System

Linux

### Relevant log output

```shell

```

### Confirm

- [x] I've searched for existing issues
- [x] I'm using the latest version

m-elhabib-dev:
I just tried to install it for opencode only and its now visible in CLI and GUI so I think the GUI recognized antigravity bmad.

bmadcode:
I just installed opencode only and I am seeing them.

<img width="531" height="455" alt="Image" src="https://github.com/user-attachments/assets/9a93c3d6-5d84-4745-9a3d-3620ac71ee14" />

m-elhabib-dev:
yes in opencode GUI it works like your screenshot, but opencode CLI donot recognize it.

bmadcode:
how do skills normally work in opencode cli @M-Alhbyb - the docs show that the same installation location works - maybe launching them is different?

bmadcode:
ok i installed - and bmad skills work in the CLi as already reported in discord also. Need to type /skills in opencode cli. then you can type the name of the skill in the pop up

<img width="1908" height="526" alt="Image" src="https://github.com/user-attachments/assets/96c276db-2d4d-4a69-8e4c-49871d4ad03a" />

moabe-br-2019:
To invoke BMAD in Open Code, you must type /skills, and then you will see the BMAD commands.

salmanfaris06:
I am still encountering issues with agent loading. The error `Unknown agent type: bmad-quick-dev` persists because the agent is not being recognized as a valid entity; it appears to be loaded only as skills. Please verify the agent type integration within the BMAD framework.

<img width="1001" height="196" alt="Image" src="https://github.com/user-attachments/assets/e73a8576-c05b-4523-8853-d44d327da43b" />

Issue #1956, opened 2026-03-13
[BUG] opencode does not load any skills involving a `workflow.md` in the skill folder


stesix:
### Description

I use opencode v1.2.20 and I've upgraded to bmad v6.1.0.

Some skills are referencing a `workflow.md` file without providing the full path to the file. Whenever I try to use one of these skills, e.g. `bmad-help`, opencode doesn't know which `workflow.md` should be used.

The following skills are referencing a workflow.md present in the same folder:
- bmad-brainstorming
- bmad-editorial-review-prose
- bmad-editorial-review-structure
- bmad-help
- bmad-index-docs
- bmad-party-mode
- bmad-quick-dev-new-preview
- bmad-review-adversarial-general
- bmad-review-edge-case-hunter
- bmad-shard-doc

Fix: 
```bash
cd .opencode
for skill_folder in * ; do 
    sed -i -E -e 's|\(workflow.md\)|\({project-root}/.opencode/skills/'"${skill_folder}"'/workflow.md\)|g' ${skill_folder}/SKILL.md
done
```

### Steps to reproduce

Use opencode in a project with bmad 6.1.0
Try the `bmad-help` skill

### Expected behavior

The `workflow.md` is read and processed as requested by the skill.

### Actual behavior

```text
> Follow the instructions in [workflow.md](workflow.md).

→ Read ${my_personal_repo_path}/workflow.md 
Error: File not found: ${my_personal_repo_path}/workflow.md

✱ Glob "**/workflow.md" (68 matches)

I found multiple workflow.md files. Which one should I follow?  

My default is .opencode/skills/bmad-quick-dev-new-preview/workflow.md since it matches the “quick dev” path.
Reply with the exact path you want, or confirm the default.
```

### Which module is this for?

BMad Method (BMM) - Core Framework

### BMad Version

6.1.0

### Which AI IDE are you using?

Other

### Operating System

macOS

### Relevant log output

```shell

```

### Confirm

- [x] I've searched for existing issues
- [x] I'm using the latest version

jairovm:
This is what I ran to fix the paths.
```bash
cd .opencode/skills

for skill_folder in *; do
  perl -pi -e "s#\(workflow\.md\)#({project-root}/.opencode/skills/$skill_folder/workflow.md)#g" "$skill_folder/SKILL.md"
done
```

bcarusodev:
I'm facing the same issue

rip222:
[...]

PR #2692, opened 2026-08-06
fix(installer): generate working OpenCode command pointers

blade035:
## Summary

Fixes opencode integration: the command pointers generated by the installer reference skills using `@skills/<id>` syntax, which **opencode does not support**. opencode resolves `@` references as file paths relative to the project root (`packages/opencode/src/config/markdown.ts` `FILE_REGEX`, `packages/opencode/src/session/prompt.ts` `resolvePromptParts`) — there is no `@skills/` namespace. As a result, no skill content is ever injected into the prompt; the LLM must rediscover the skill by agentic file exploration (non-deterministic, and a no-op for non-agentic editors).

This PR changes the installer's default command-pointer body template to emit a **real relative path** (`@{target_dir}/{canonicalId}/SKILL.md`), which opencode resolves correctly.

Related issues:
- #1956 ([BUG] opencode does not load any skills involving a `workflow.md` in the skill folder) — likely the same root cause: skill content never reaches the prompt.
- #2319 ([BUG] Bmad is not working on OpenCode CLI) — closed; this PR addresses the underlying pointer/syntax problem.

## What changed

- `tools/installer/ide/_config-driven.js`
  - `DEFAULT_COMMANDS_BODY_TEMPLATE`: `@skills/{canonicalId}` → `@{target_dir}/{canonicalId}/SKILL.md`
  - `{target_dir}` was already supported by `expandBodyTemplate` (lines 96–97); only the default template and comments were wrong.
  - Corrected misleading comments that claimed `@skills/<id>` "matches OpenCode's native syntax".
- `tools/installer/ide/platform-codes.yaml`
  - opencode entry: added `commands_body_template: "@{target_dir}/{canonicalId}/SKILL.md"` + explanatory comment.
- `test/test-installation-components.js` (Suite 8)
  - Assertion updated to expect `@.agents/skills/bmad-master/SKILL.md`.

## Verification

All green:
- `node test/test-installation-components.js` — all suites pass (incl. updated Suite 8)
- `npm run lint`, `npx prettier --check`, `npm run format:check`
- full `npm test` (test:refs / test:install / test:urls / test:channels / test:renderer / test:retrospective / test:sprint-planning / test:skills / lint / lint:md / format:check)

End-to-end repro with the generated pointer (opencode CLI 1.18.14):

1. Fresh install generates pointer body `@.agents/skills/bmad-master/SKILL.md` (was `@skills/bmad-master`).
2. `opencode run --command bmad-master` → skill marker `PR_MARKER_LOADED_9f2c` seen in the session: content is injected.
3. Proof the reference is resolved as a file: a pointer targeting `pointer-target.md` produced its marker `FROM_POINTER_bbb` in the session.
4. Negative control: the old `@skills/bmad-master` body injected **nothing** — the model had to agentically `Read _bmad/_config/skill-manifest.csv` and the SKILL.md itself to "discover" the skill.
5. Without any pointer, the skill-registered command failed in the CLI (`UnknownError: Unexpected server error`), while a plain `hello.md` command worked — the failure is specific to skill-backed commands.

## Migration note (existing installs)

`commands_body_template` only takes effect on a fresh or re-run install; pointer files that already exist are not rewritten (existing content is respected, and the generator's reserved-guard skips files it doesn't own). Existing opencode installs with old `@skills/<id>` pointers must delete the stale files in `commands_target_dir` (`remove-command-pointers` / manual deletion) and re-run the installer.

### DIFF
--- a/test/test-installation-components.js
+++ b/test/test-installation-components.js
@@ -360,6 +360,7 @@ async function runTests() {
       commandContent.includes('@.agents/skills/bmad-master/SKILL.md'),
       'Command pointer body references the installed skill via @<target_dir>/<canonicalId>/SKILL.md',
     );
+    assert(!commandContent.includes('@skills/bmad-master'), 'Command pointer does not use the legacy @skills/<canonicalId> namespace');
     assert(commandContent.includes('description:'), 'Command pointer carries a description in YAML frontmatter');
 
     // Idempotency: re-running install must not duplicate or rewrite pointers.
--- a/test/test-installation-components.js
+++ b/test/test-installation-components.js
@@ -356,7 +356,10 @@ async function runTests() {
     assert(await fs.pathExists(commandFile), 'OpenCode install writes per-skill command pointer file');
 
     const commandContent = await fs.readFile(commandFile, 'utf8');
-    assert(commandContent.includes('@skills/bmad-master'), 'Command pointer body references the skill via @skills/<canonicalId>');
+    assert(
+      commandContent.includes('@.agents/skills/bmad-master/SKILL.md'),
+      'Command pointer body references the installed skill via @<target_dir>/<canonicalId>/SKILL.md',
+    );
     assert(commandContent.includes('description:'), 'Command pointer carries a description in YAML frontmatter');
 
     // Idempotency: re-running install must not duplicate or rewrite pointers.
--- a/tools/installer/ide/_config-driven.js
+++ b/tools/installer/ide/_config-driven.js
@@ -53,9 +53,14 @@ function isSafeCanonicalId(value) {
 }
 
 // Default body template for command pointer files. Used when a platform's
-// installer config doesn't override `commands_body_template`. Matches
-// OpenCode's native `@skills/<id>` skill-reference syntax.
-const DEFAULT_COMMANDS_BODY_TEMPLATE = '@skills/{canonicalId}';
+// installer config doesn't override `commands_body_template`.
+//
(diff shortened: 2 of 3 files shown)


sanmaxdev (commented):
Verified the command pointer resolves directly to the installed `SKILL.md` and the regression guard excludes the legacy namespace. `npm ci && npm run quality` passed on `ddf9f4926c6388dbebf121bee1dac34d8ffeefbc`.

bmadcode:
Thanks @blade035 — and thanks especially for tracing this into opencode's own source rather than just reporting that it didn't work. Establishing that `@skills/` isn't a namespace opencode resolves, so no skill content was ever reaching the prompt, is exactly the kind of report that makes a fix obvious.

**BMAD is changing how it is distributed.** The project is moving to a flat `skills/` tree installed through `npx skills` and plugin marketplaces, and `tools/installer/` is removed entirely as part of that change. `tools/installer/ide/_config-driven.js` and `ide/platform-codes.yaml` are both removed, so the generated command pointers this PR fixes no longer exist in that form.

Opencode support does still need to work. Once the new distribution is up, if opencode pointers are still broken please open a fresh issue against it — your finding about `@` resolution will still be the key detail.

---

blade035:
Thanks @bmadcode — makes sense given where the distribution is heading.

I'll watch #2768 and re-test opencode once the flat `skills/` tree lands, reusing the marker-injection repro from this PR's description. If command pointers are still broken there, I'll open a fresh issue with the `@`-resolution root cause.

PR #2693, opened 2026-08-07
fix: pin UTF-8 stdout in resolve_config so Windows cp1252 doesn't crash

### LINKED ISSUE #2682: resolve_config.py crashes on Windows (UnicodeEncodeError) — party mode silently resolves zero agents
## Summary

`src/scripts/resolve_config.py` writes JSON with `ensure_ascii=False` without pinning stdout's
encoding. On Windows, stdout defaults to the locale codepage (cp1252), so the write raises
`UnicodeEncodeError` on the emoji `icon` fields in the shipped agent configs. The script exits
non-zero having produced **no output**.

Its sibling in the same directory, `src/scripts/resolve_customization.py`, already guards the
identical write. This looks like a straightforward oversight rather than a design choice.

## Impact — silent, not loud

The crash is invisible to the user. `resolve_party.load_agents()` treats an unusable result as
"config could not be resolved" and returns an empty dict, so **party mode presents an empty room
instead of an error**:

```json
{ "party_mode": "session", "installed_agents_resolved": false, "members": [] }
```

That is on a stock install with all six agent skills present. Nothing in the UI indicates that
anything failed.

## Reproduction

Windows, stock install:

```
npx bmad-method@6.10.0 install --yes --directory . --modules bmm --tools claude-code --output-folder _bmad-output
python _bmad/scripts/resolve_config.py --project-root . --key agents
```

```
Traceback (most recent call last):
  File "...\_bmad\scripts\resolve_config.py", line 174, in main
    sys.stdout.write(json.dumps(output, indent=2, ensure_ascii=False) + "\n")
  File "C:\Python314\Lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\U0001f4ca' in position 184: character maps to <undefined>
```

`\U0001f4ca` is 📊, the `icon` on the analyst agent. The shipped agents carry emoji icons
(📊, 📚, 📋, …), so on Windows this fires every time — it is not an edge case.

Then observe the downstream effect:

```
python .claude/skills/bmad-party-mode/scripts/resolve_party.py --project-root . --skill bmad-party-mode
```

## Root cause

(Line numbers above are from the *installed* copy at `_bmad/scripts/`; in the repo the same
statement is `src/scripts/resolve_config.py:69`.)

```python
sys.stdout.write(json.dumps(output, indent=2, ensure_ascii=False) + "\n")
```

`ensure_ascii=False` emits raw non-ASCII, but stdout is still bound to the platform default
encoding. On a UTF-8 platform (macOS/Linux) this is harmless, which is likely why it has gone
unnoticed.

## Why the test suite does not catch it

`src/scripts/tests/test_resolve_config.py` passes on Windows. Every `encoding="utf-8"` in it is on
*fixture writes* (creating temp TOML files); nothing exercises the script's **stdout** path. The
defect only manifests when stdout is a cp1252 stream, which the tests never construct.

## Suggested fix

Adopt the guard already used by the sibling script — `src/scripts/resolve_customization.py:48-51`:

```diff
--- a/src/scripts/resolve_config.py
+++ b/src/scripts/resolve_config.py
@@
+    reconfigure = getattr(sys.stdout, "reconfigure", None)
+    if reconfigure is not None:
+        reconfigure(encoding="utf-8")
     sys.stdout.write(json.dumps(output, indent=2, ensure_ascii=False) + "\n")
```

Verified locally: the child then emits valid JSON including the emoji, and party mode resolves
`installed_agents_resolved: true` with all six agents.

## Two related defects, same root cause (non-ASCII crossing a process boundary)

Both readers of the above scripts decode child output without pinning an encoding:

- `src/core-skills/bmad-party-mode/scripts/resolve_party.py:49`
- `src/core-skills/bmad-forge-idea/scripts/resolve_personas.py:51`

```diff
-        out = subprocess.run(cmd, capture_output=True, text=True, timeout=60)
+        out = subprocess.run(cmd, capture_output=True, text=True, encoding="utf-8", timeout=60)
     except (OSError, subprocess.SubprocessError):
         return None
-    if out.returncode != 0 or not out.stdout.strip():
+    if out.returncode != 0 or not (out.stdout or "").strip():
```

Without `encoding=`, a UTF-8 child decoded as cp1252 can kill the reader thread and return
`stdout=None`, which the following line dereferences. Note `_run_json`'s docstring promises *"None
on any failure"*, but its `except (OSError, subprocess.SubprocessError)` does not catch
`UnicodeDecodeError` — so the fail-safe does not cover this failure mode. The `(out.stdout or "")`
guard makes the promise true regardless.

## Environment

- Windows 11, Python 3.14.6, stdout codepage cp1252
- `bmad-method@6.10.0` (npm `latest`)
- **Also verified present in `6.10.1-next.47`** — all three sites unchanged, and
  `resolve_customization.py` still carries the guard that `resolve_config.py` lacks

armelhbobdad:
## What

Pins `resolve_config.py`'s stdout to UTF-8 before the JSON dump, and adds the stdout regression test the suite was missing.

## Why

The full-config dump writes `ensure_ascii=False` JSON to a stdout still bound to the platform default. On Windows that is cp1252, which cannot encode the emoji icons in the shipped agent configs, so the script raises `UnicodeEncodeError` and exits having produced no output.

Fixes #2682

That issue was auto-closed by #2687, but #2687 changed only the **consumer** side (`encoding="utf-8"` on party mode's `subprocess.run`). That governs how the parent decodes the child's bytes; it does not change the child's stdout encoding, so the producer still raises before writing anything and `_run_json` still sees a non-zero exit. Verified still reproducible on `main`.

## How

- Adopt `write_json_stdout()` verbatim from the sibling `resolve_customization.py`, which already guards the identical write (fixed in #2414).
- Add `test_writes_emoji_json_when_stdout_encoding_is_cp1252`, mirroring the existing test of the same name in `test_resolve_customization.py`. The issue noted the suite never exercised the stdout path — it does now.

## Testing

`npm run test:renderer` → 11 tests, OK. The new test fails against the unpatched script (confirmed by reverting the fix and re-running), so it is a real tripwire.

## Note on #2578

@aranellaeth's #2578 proposed this same `resolve_config.py` guard back on 2026-07-12 and deserves the credit. Its consumer hunks have since been merged piecemeal (#2687 party mode, #2688 forge-idea), leaving the producer fix — the part that actually closes #2682 — outstanding. This PR is deliberately scoped to just that piece plus its test, so it can land independently; close it in favour of #2578 if you would rather merge that whole.

One review note for whichever lands: `reconfigure(encoding="utf-8")` without `errors=` also resets the handler to `strict`. Harmless for stdout (this PR), but #2578 also reconfigures **stderr** in `brain.py`, which silently downgrades POSIX's default `backslashreplace` — a surrogateescaped path in an error message would turn a clean diagnostic into a traceback. Passing `errors=stream.errors` preserves it.

### DIFF
--- a/src/scripts/resolve_config.py
+++ b/src/scripts/resolve_config.py
@@ -34,6 +34,14 @@ def extract_key(data, dotted_key: str):
     return current
 
 
+def write_json_stdout(output) -> None:
+    """Pin stdout to UTF-8 — a Windows cp1252 default cannot encode emoji icons."""
+    reconfigure = getattr(sys.stdout, "reconfigure", None)
+    if reconfigure is not None:
+        reconfigure(encoding="utf-8")
+    sys.stdout.write(json.dumps(output, indent=2, ensure_ascii=False) + "\n")
+
+
 def main() -> int:
     parser = argparse.ArgumentParser(
         description="Resolve BMad central config using four-layer TOML merge."
@@ -66,7 +74,7 @@ def main() -> int:
             value = extract_key(merged, key)
             if value is not _MISSING:
                 output[key] = value
-    sys.stdout.write(json.dumps(output, indent=2, ensure_ascii=False) + "\n")
+    write_json_stdout(output)
     return 0
 
 
--- a/src/scripts/tests/test_resolve_config.py
+++ b/src/scripts/tests/test_resolve_config.py
@@ -1,4 +1,5 @@
 import json
+import os
 import shutil
 import subprocess
 import sys
@@ -73,6 +74,33 @@ class ResolveConfigCliTests(unittest.TestCase):
             result = self._run(root)
             self.assertNotEqual(result.returncode, 0)
             self.assertIn("failed to parse", result.stderr)
+
+    def test_writes_emoji_json_when_stdout_encoding_is_cp1252(self):
+        with tempfile.TemporaryDirectory() as temp_dir:
+            root = Path(temp_dir)
+            (root / "_bmad").mkdir(parents=True)
+            (root / "_bmad" / "config.toml").write_text(
+                '[agents]\nname = "Analyst"\nicon = "📊"\n',
+                encoding="utf-8",
+            )
+
+            env = os.environ.copy()
+            env["PYTHONIOENCODING"] = "cp1252"
+            result = subprocess.run(
+                [sys.executable, str(SCRIPT), "--project-root", str(root)],
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                env=env,
+                check=False,
+            )
+
+            stderr = result.stderr.decode("utf-8", errors="replace")
+            self.assertEqual(result.returncode, 0, msg=stderr)
+
+            output = result.stdout.decode("utf-8")
+            self.assertIn("📊", output)
+            resolved = json.loads(output)
+            self.assertEqual(resolved["agents"]["icon"], "📊")
 
     @staticmethod
     def _run(root: Path, *args: str) -> subprocess.CompletedProcess[str]:


armelhbobdad:
Good catch from Greptile — valid, and worse than described. Fixed in 62f2250.

I reproduced it before changing anything. The failure is not a propagating `UnicodeDecodeError`: it is raised on subprocess's **reader thread**, so `subprocess.run` returns normally with `stdout=None` and `returncode=0`. That slips straight past `_installed_resolver_config`'s `if proc.returncode != 0` guard and dies in `json.loads(None)` as a `TypeError`, which the surrounding `except json.JSONDecodeError` does not catch.

Three-way repro on Windows 11 / Python 3.11, config carrying `project_name = "Café 📍"` (U+1F4CD's UTF-8 contains `0x8D`, undefined in cp1252):

| producer | consumer | result |
|---|---|---|
| `main` | `main` | `None` — child crashed, clean fallback |
| this PR (before 62f2250) | `main` | **`UNCAUGHT: TypeError`** |
| this PR (now) | patched | `{'project_name': 'Café 📍'}` |

So the middle row was a genuine regression this PR introduced: it turned a silent-but-graceful fallback into a crash. Worth stating plainly.

The bot said "two `context.py` consumers" and that is literally true — `src/scripts/context.py` and `src/bmm-skills/plan/bmad-project-context/scripts/context.py` are byte-identical copies (same sha256), each with its own call site. Both are patched and still byte-identical after the change. They are also the *only* remaining locale-decoding consumers of `resolve_config.py`; notably #2578 does not cover them either.

The fix is the same one-liner already merged for the sibling consumers — `encoding="utf-8"` on the capture, matching #2687 and #2688 (both +3/-1, no test). I used `encoding="utf-8"` alone rather than adding `errors="replace"`: the producer now guarantees UTF-8, so the decode is exact, and `errors="replace"` would only convert genuine corruption into silent mojibake.

No test added for the consumer hop, matching the precedent of those two merged PRs — `test_context.py` is pytest-style, has no npm script wiring it into CI, and covers none of the resolver path. Happy to add one if you would rather have it.

`npm run test:renderer` → 11 tests, OK.

armelhbobdad:
Added the consumer test in 3d769a0.

**Where it lives.** Not in `test_context.py`, which would have been the obvious home. That file is wired into no npm script — CI's Python coverage is `test:renderer`, `test:retrospective` and `test:sprint-planning` — and it has rotted since: on a clean `main` checkout it is **26 failed / 20 passed**. A test added there would never run, and wiring it in would mean importing 26 pre-existing failures into CI, which is out of scope here. Flagging it as a separate cleanup worth doing. The new test therefore sits in `test_resolve_config.py`, which `test:renderer` already runs.

**Why it is not a cp1252 test.** The existing precedent (`PYTHONIOENCODING=cp1252`) works for the producer because that variable governs the script's own stdout. It is the wrong lever for the consumer: the parent's capture decodes with `locale.getpreferredencoding(False)`, which `PYTHONIOENCODING` does not touch. A locale-based test would bite on Windows and pass vacuously on Linux CI.

Instead the driver runs under `-X warn_default_encoding -W error::EncodingWarning`. `subprocess` calls `io.text_encoding(None)` when `encoding=` is omitted, so the omission becomes a hard error on **every** platform, Linux CI included. Verified `context.py` is encoding-clean first (19 explicit `encoding=`, its only bare `open` is `"rb"`), so there are no false positives.

It exercises the real `_installed_resolver_config` — imported from `context.py`, not a copy — against a temp project with an installed resolver and `project_name = "Café 📍"`.

Verified as a tripwire in both directions:

| state | result |
|---|---|
| this PR | **OK** |
| consumer `encoding=` removed | **FAILED** |
| producer guard reverted | **FAILED** (errors=1) |

`npm run test:renderer` → 12 tests, OK. Both `context.py` copies remain byte-identical.

aranellaeth:
Author of #2578 here. Thanks for the flag on my PR, and for scoping this one to the producer — that is the right call, and I'd rather this land than mine.

I've rebased #2578 onto current `main` and narrowed it to `brain.py`, the one file neither #2687, #2688 nor this PR touches and the only one still unfixed. The two PRs no longer overlap, so nothing needs closing in favour of anything.

Two things I reproduced on Windows 11 while doing that, both of which back up your review notes.

**Your `errors=` note is correct, and it applied to my patch.** `reconfigure(encoding=...)` with no `errors=` silently resets the handler to `strict`:

```python
>>> s = io.TextIOWrapper(io.BytesIO(), encoding="ascii", errors="backslashreplace")
>>> s.reconfigure(encoding="utf-8");                    (s.encoding, s.errors)
('utf-8', 'strict')
>>> t.reconfigure(encoding="utf-8", errors=t.errors);   (t.encoding, t.errors)
('utf-8', 'backslashreplace')
```

My `brain.py` hunk reconfigured stderr, so it would have downgraded POSIX's default exactly as you described. Fixed in the rescoped PR by passing `errors=` through.

**Your reader-thread finding reproduces — and it also lands on the two consumers that already merged.** #2687 and #2688 took the `encoding="utf-8"` half of the fix but not the `errors="replace"` / `out.stdout or ""` half. Against a child emitting invalid UTF-8, on Python 3.13:

```
encoding="utf-8", strict          ->  returncode=0   stdout=None
    out.stdout.strip()  AttributeError: 'NoneType' object has no attribute 'strip'
    (uncaught — the handler is `except json.JSONDecodeError`)

encoding="utf-8", errors="replace" ->  returncode=0   stdout='{"agents": ...}'
```

Same failure you found in `_installed_resolver_config`, one exception type off because `_run_json` calls `.strip()` before `json.loads`. So `resolve_party.py` and `resolve_personas.py` are currently half-fixed on `main`. Worth a small follow-up; I'll open it if nobody beats me to it.

And thank you for the credit note in the description — genuinely appreciated.

bmadcode:
Rebased onto main to clear the conflict. #2698 removed `src/scripts/context.py` and the bmm-skills copy, so the two commits touching them had nothing left to patch and were dropped:

- `fix: decode resolve_config output as UTF-8 in context resolvers` — both target files are gone from main
- `test: cover the context.py consumer of resolve_config output` — its driver loads the deleted file

The `resolve_config.py` fix and its cp1252 stdout test are untouched. Verified locally: `npm run test:renderer` passes (11 python + 24 node), and the new test fails against the unpatched script. Thanks for the fix — merging once CI is green.

bmadcode (approved):
Verified locally after the rebase: test:renderer passes (11 python + 24 node), the new cp1252 stdout test fails against the unpatched script, and both subprocess consumers already decode UTF-8 (#2687/#2688).

PR #2695, opened 2026-08-07
fix(installer): stop copying __pycache__ into IDE skill trees

### LINKED ISSUE #2694: Installer ships upstream `__pycache__/*.pyc` into the IDE skill tree, contradicting its own "no Python caches in user projects" policy
> **Correction (2026-08-07):** the original text of this issue claimed `files-manifest.csv` contained rows for files that were never installed. **That claim was wrong** — I resolved the manifest's module-relative paths against `_bmad/` when installed skills actually live in the IDE skill tree (`.claude/skills/`). The manifest is accurate; its hashes match the installed files exactly. The real defect, verified below, is narrower: the skill-copy filter ships upstream Python bytecode into user projects. Body rewritten accordingly; apologies for the noise.

---

## Summary

The IDE skill installer copies `__pycache__/*.pyc` out of the published npm package and into every user's skill tree. The installer already declares this must not happen — `_installSharedScripts` filters exactly these artifacts with the comment *"Ship only the runtime scripts — dev-only tests and Python caches must not land in user projects."* That policy is enforced on one install path and not the other.

On a `6.10.1-next.51` install with `claude-code` selected:

```
$ find .claude/skills -name "*.pyc" -path "*bmad*"
.claude/skills/bmad-retrospective/scripts/__pycache__/sprint_status.cpython-311.pyc
.claude/skills/bmad-retrospective/scripts/tests/__pycache__/test_git_evidence.cpython-311-pytest-9.1.1.pyc
.claude/skills/bmad-retrospective/scripts/tests/__pycache__/test_sprint_status.cpython-311-pytest-9.1.1.pyc
.claude/skills/bmad-sprint-planning/scripts/__pycache__/sprint_plan.cpython-311.pyc
.claude/skills/bmad-sprint-planning/scripts/tests/__pycache__/test_sprint_plan.cpython-311-pytest-9.1.1.pyc
```

Meanwhile `_bmad/scripts/` — installed through the path that *does* filter — is clean:

```
$ ls -a _bmad/scripts/
config_utils.py  context.py  memlog.py  render_skill.py  resolve_config.py  resolve_customization.py
```

## These are the maintainer's build artifacts, not the user's

The bytecode is byte-identical from package to disk:

```
package  src/bmm-skills/plan/bmad-sprint-planning/scripts/__pycache__/sprint_plan.cpython-311.pyc
         5a0f93358727b816bd3dff7ce048be09f712a14402e0ec4b5d9433540cd6022d
manifest 5a0f93358727b816bd3dff7ce048be09f712a14402e0ec4b5d9433540cd6022d
on disk  5a0f93358727b816bd3dff7ce048be09f712a14402e0ec4b5d9433540cd6022d
```

They are `cpython-311` while the local interpreter is Python 3.14.3 — so they cannot have been produced locally. The published tarball carries 9 `.pyc` files in total (`package.json` has no `files` field and no `.npmignore` covers them); 5 sit under `src/bmm-skills/**` and reach users, 4 sit under `src/scripts/**` and are correctly filtered out.

## Cause

`ConfigDrivenIdeSetup`'s skill copy builds a filter that covers OS and editor artifacts only (`tools/installer/ide/_config-driven.js:452-461`):

```js
const skipPatterns = new Set(['.DS_Store', 'Thumbs.db', 'desktop.ini']);
const skipSuffixes = ['~', '.swp', '.swo', '.bak'];
const filter = (src) => {
  const name = path.basename(src);
  if (src === sourceDir) return true;
  if (skipPatterns.has(name)) return false;
  if (name.startsWith('.') && name !== '.gitkeep') return false;
  if (skipSuffixes.some((s) => name.endsWith(s))) return false;
  return true;
};
await fs.copy(sourceDir, skillDir, { filter });
```

`.pytest_cache` happens to be excluded by the leading-dot rule. `__pycache__` does not start with a dot, so it and its `.pyc` contents pass straight through.

Compare `Installer._installSharedScripts` (`tools/installer/core/installer.js:676-681`), which encodes the intended policy:

```js
// Ship only the runtime scripts — dev-only tests and Python caches must not land in user projects.
const isInstallable = (srcPath) => {
  const base = path.basename(srcPath);
  return base !== 'tests' && base !== '__pycache__' && base !== '.pytest_cache' && !base.endsWith('.pyc');
};
```

`FileOps.shouldIgnore` (`tools/installer/file-ops.js:129`) lists `__pycache__` and `*.pyc` too, but the skill copy doesn't route through it.

## Why it matters

- Stale bytecode compiled against a *different* source revision sits next to the `.py` files users actually run. CPython validates the cache against source size/mtime, so this is unlikely to cause a wrong execution — but it is dead weight that survives every update.
- `.pyc` files are not reproducible across builds, so they add churn to the published package for no consumer.
- Users who commit `.claude/skills/` (common, since skills are team configuration) get binary artifacts in code review.
- It contradicts a policy the codebase states explicitly in a comment.

## Reproduce

```bash
npx bmad-method@6.10.1-next.51 install     # select bmm + claude-code
find .claude/skills -name "*.pyc"          # 5 files
ls -a _bmad/scripts/                       # clean, for contrast
```

## Suggested fix

Add `__pycache__` and `*.pyc` to the skill-copy filter so both install paths enforce the same rule. PR incoming.

Independently worth considering: exclude `__pycache__` from the published package via a `files` field or `.npmignore`, so the artifacts never ship at all.

## Environment

- bmad-method `6.10.1-next.51`, IDE `claude-code`
- macOS 15 (Darwin 24.6.0), Node via npx, local Python 3.14.3
- Modules: core, bmm, bmad-loop

tjoignant:
## What

IDE skill copying now excludes `__pycache__` directories and `.pyc`/`.pyo` files, preventing Python bytecode from the npm package from reaching users' skill trees.

## Why

Fixes #2694. The published package currently contains maintainer-generated Python bytecode, and the IDE skill installer copies it into projects.

## How

- Extended the existing recursive skill-copy filter with Python cache names and suffixes.
- Added observable-behavior assertions to the existing Claude Code installation test for cache directories and root-level bytecode files.
- Kept Python source and tracked skill content behavior unchanged.

## Testing

`HUSKY=0 npm ci && npm run quality`

### DIFF
--- a/test/test-installation-components.js
+++ b/test/test-installation-components.js
@@ -420,6 +420,11 @@ async function runTests() {
 
     const tempProjectDir9 = await fs.mkdtemp(path.join(os.tmpdir(), 'bmad-claude-code-test-'));
     const installedBmadDir9 = await createTestBmadFixture();
+    const sourceSkillDir9 = path.join(installedBmadDir9, 'core', 'bmad-master');
+    await fs.ensureDir(path.join(sourceSkillDir9, '__pycache__'));
+    await fs.writeFile(path.join(sourceSkillDir9, '__pycache__', 'cached.pyc'), 'bytecode');
+    await fs.writeFile(path.join(sourceSkillDir9, 'cached.pyc'), 'bytecode');
+    await fs.writeFile(path.join(sourceSkillDir9, 'cached.pyo'), 'bytecode');
 
     const ideManager9 = new IdeManager();
     await ideManager9.ensureInitialized();
@@ -432,6 +437,16 @@ async function runTests() {
 
     const skillFile9 = path.join(tempProjectDir9, '.claude', 'skills', 'bmad-master', 'SKILL.md');
     assert(await fs.pathExists(skillFile9), 'Claude Code install writes SKILL.md directory output');
+    const installedSkillDir9 = path.dirname(skillFile9);
+    assert(
+      !(await fs.pathExists(path.join(installedSkillDir9, '__pycache__'))),
+      'Claude Code skill install excludes Python cache directories',
+    );
+    assert(
+      !(await fs.pathExists(path.join(installedSkillDir9, 'cached.pyc'))) &&
+        !(await fs.pathExists(path.join(installedSkillDir9, 'cached.pyo'))),
+      'Claude Code skill install excludes Python bytecode',
+    );
 
     // Verify name frontmatter matches directory name
     const skillContent9 = await fs.readFile(skillFile9, 'utf8');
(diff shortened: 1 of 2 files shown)


alexeyv (approved):
Simplified the cache coverage into the existing installer test; the full quality gate passes.

alexeyv:
@tjoignant thanks for the PR!

PR #2696, opened 2026-08-07
feat(build): add direct and early one-shot routes

alexeyv:
## What

Adds two low-ceremony pre-render routes to `bmad-build`:

- **Direct** for user-pinned edits to a bounded span.
- **Early one-shot** for clear single-goal implementation requests whose targeted source probe establishes zero blast radius.

Everything else retains the full workflow.

## Why

Small changes were paying full activation, artifact-discovery, and clarification costs before reaching useful work. In the controlled one-shot experiment, early routing reached the first green verification about 38 seconds sooner; the closest end-to-end comparison saved 14 seconds. The direct route removes still more ceremony for edits that require no investigation.

## How

- Triage before rendering among direct, early one-shot probe, and full workflow.
- Inline named routes from the same immutable renderer snapshot.
- Keep direct edits in a commit-owning interactive loop with merge-blocker review for code.
- Make one-shot a standalone named route that derives its spec path and skips sprint synchronization when no story is present.
- Preserve full-workflow fallback and carry probe findings forward.

## Testing

- `npm ci && npm run quality`
- Shared renderer: 29/29
- Installer: 458/458
- Strict reference and skill validation: clean

Project-root discovery remains deliberately out of scope; callers still provide `--project-root`.

### DIFF
--- a/src/bmm-skills/ship/bmad-build/SKILL.md
+++ b/src/bmm-skills/ship/bmad-build/SKILL.md
@@ -3,11 +3,23 @@ name: bmad-build
 description: 'Implements any user intent, requirement, story, bug fix or change request by producing clean working code artifacts that follow the project''s existing architecture, patterns and conventions. Use when the user wants to build, fix, tweak, refactor, add or modify any code, component or feature.'
 ---
 
-Run the following command exactly once without changing the current working directory. Replace `{project-root}` with the absolute path to the project root and `{skill-root}` with the absolute path to this skill's directory:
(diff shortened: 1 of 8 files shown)


bmadcode:
Thank you — closing this draft: it routes at step-01 before investigation, which the problem statement of your later #2762 identifies as the flaw being removed. Conflicting and untouched since early August.

—

alexeyv:
Superseded by #2762, which preserves the low-ceremony path but moves the routing decision until after investigation and design, when forks, irreversibles, and footprint are known rather than predicted. Closing this PR in favor of that approach.

PR #2707, opened 2026-08-10
fix(installer): stop --yes updates from silently deleting custom modules

eugeniawang:
## What

`bmad install --action update --yes` no longer deletes an installed module recorded in the manifest as `source: custom` whose source is still resolvable, and module removals are logged instead of happening silently.

## Why

What we hit. Two separate BMAD installs, in two unrelated directories, each with its own `_bmad/` — not nested, no shared install:

- **Project A** (`<dev>/_bmad/`) — `core`, `bmm`, `bmb`, `cis`, `tea`, `bmad-loop`, `wds`. All built-in or from the registry.
- **Project B** (`<video-project>/_bmad/`) — `core`, `bmm`, and `manticore`, the last installed from `github.com/bmad-code-org/bmad-manticore` and recorded `source: custom`.

One `bmad install --directory <video-project> --action update --yes` against Project B deleted `<video-project>/_bmad/manticore/` outright, along with all 14 of its `mc-*` skills under that project's `.claude/skills/` and `.agents/skills/` — 310 files. It removed the module's manifest entry and exited 0 with nothing in the output about any of it. `core` and `bmm` in that same directory updated normally to `6.10.1-next.55`. Only the custom module was lost, and the interactive path preserves it correctly.

Note the condition: a custom module whose source can no longer be found *is* preserved today, by the safety net below. It is the correctly-installed one, with its source still cached, that gets deleted.

`getDefaultModules()` (`ui.js:1376-1407`), used only by the `--yes` path, builds its keep-list from the built-in catalog and the external registry, so an installed custom module is never a candidate. `_retainUnavailableInstalledModules()` (`ui.js:151-192`) is the safety net, but it preserves a module only when **no** source can be found:

```js
const customSource = await customMgr.findModuleSourceByCode(moduleId, { bmadDir });
if (!customSource) {
  preserveModules.push(moduleId);
}
```

A correctly-installed custom module has a findable source in `~/.bmad/cache/custom-modules/`, so it is read as a deliberate deselection and `_removeDeselectedModules()` (`installer.js:150-166`) deletes it. The only log statement in that function is inside its `catch`, so a successful removal prints nothing.

The interactive path already has the guard, at `ui.js:960-971`:

```js
// Identify installed modules that aren't official (previously installed
// community modules or custom-source modules). Preserve them on update;
// they can be managed via --custom-source, uninstall, or a dedicated installer.
const installedNonOfficial = [...installedModuleIds].filter((id) => !officialRegistryCodes.has(id));
```

And `ui.js:345-350` shows the `--yes` path was assumed to behave the same way:

```js
} else if (options.customSource && !options.yes) {
  // Custom source without --modules or --yes: start with empty list
  // (only custom source modules + core will be installed).
  // When --yes is also set, fall through to the --yes branch so all
  // installed modules are included alongside the custom source modules.
```

That comment assumes the `--yes` branch includes all installed modules. `getDefaultModules` does not include custom ones.

## How

- `getDefaultModules()` — carry forward installed modules present in neither catalog, matching `selectAllModules()` and the assumption at `ui.js:345-350`. Ids are normalized through `externalManager.resolveCanonicalCode()` first, the same way `_retainUnavailableInstalledModules()` does, so a renamed or aliased id is not carried forward unresolved.
- `_removeDeselectedModules()` — log on the success path, so a module cannot disappear without a line of output.
- No change to `--modules`, which bypasses `getDefaultModules()` entirely (`ui.js:338`).

## Testing

After the fix, `manticore` and its 14 skills survive the same command and `core`/`bmm` still update to `6.10.1-next.55`. We also confirmed on the unpatched code that passing `--custom-source https://github.com/bmad-code-org/bmad-manticore` preserves the module, which pinned the cause to module *selection* rather than to removal.

Two limits on what we verified, both worth stating:

- We run this exact command daily against Project A, whose modules are all built-in or from the registry. 62 runs since early June, 55 of them clean (the other 7 failed on network or on an unrelated `--tools` requirement), and no module has ever been lost there. The failure appears only once a `source: custom` module is present.
- We ran one project with one custom module. That the same applies to any resolvable custom module follows from `getDefaultModules()` never consulting custom sources at all, but we have not run a multi-custom-module case.


alexeyv (changes requested):
Requesting one focused change before merge.

In `getDefaultModules()`, the new loop resolves an installed ID to its canonical code only for the membership checks, then appends the original ID:

```js
const canonicalId = await externalManager.resolveCanonicalCode(moduleId);
// ...
defaultModules.push(moduleId);
```

For the existing `bauto` → `bmad-loop` alias, this still returns `bauto`. That contradicts the comment and PR description, and it bypasses code that indexes registry and channel state by the canonical code. A pinned or `next` installation can consequently fall back to the registry default during a full `--yes` update.

Please carry the canonical ID forward and deduplicate selections by canonical ID. Please also add focused deterministic regression tests covering:

- A resolvable custom module remains selected during `--action update --yes`.
- An installed alias such as `bauto` produces the canonical `bmad-loop` selection.
- The canonicalized selection retains the installed module's channel decision.

The custom-module preservation change and removal logging otherwise look good.

eugeniawang:
Thanks — you were right on both counts. Pushed as 064c6097.

**Canonical ID carried forward.** The loop now pushes `canonicalId` and dedupes on it:

```js
for (const moduleId of installedModuleIds) {
  const canonicalId = await externalManager.resolveCanonicalCode(moduleId);
  if (canonicalId === 'core') continue;
  if (seen.has(canonicalId)) continue;
  seen.add(canonicalId);
  defaultModules.push(canonicalId);
}
```

One thing I had to change to make the dedupe actually canonical: `seen` was only populated by the built-in loop, so an installed registry module reached the carry-forward loop unrecorded. The old `defaultModules.includes(...)` checks papered over that. Both catalog loops now add to `seen`, and the carry-forward loop tests it alone — so `bauto` and `bmad-loop` installed side by side (which `installer.js:1515` explicitly contemplates) yield exactly one `bmad-loop` entry.

**Tests** — `test/test-installer-default-modules.js`, registered as `test:default-modules` in the `test` and `quality` chains. Seven assertions across your three cases plus the dedupe pair. Both catalogs are stubbed at the prototype level and restored in a `finally`, so the tests are offline and independent of any BMAD install or `~/.bmad` cache. The registry fixture declares `bmad-loop` with `aliases: ['bauto']`, so `resolveCanonicalCode()` runs for real rather than being stubbed.

Each test was confirmed to go red when the behaviour it guards is removed:

| Removed | Assertion that failed |
|---|---|
| Carry-forward loop | `custom module survives --action update --yes` — expected `"manticore"`, got `["bmm"]` |
| Canonicalization (`push(moduleId)`) | `installed alias selects its canonical module code` — expected `"bmad-loop"`, got `["bauto"]` |
| Canonicalization (`push(moduleId)`) | `canonicalized selection retains the installed --next channel decision` — expected `"next"`, got `undefined` |
| Dedupe by canonical ID | `alias and canonical module deduplicate to one canonical selection` — expected `1`, got `3` |

That third one is your channel point made concrete: pre-fix the selection is `['bauto']`, so `buildPlan()` — which keys on `mod.code` — has no `bmad-loop` entry at all and the `--next` decision is simply absent.

Full `npm test` is clean locally.

bmadcode:
it looks like there are a few conflicts @eugeniawang - this is a good fix to get in if you can resolve the conflicts we can merge this for next release

bmadcode:
@eugeniawang — thank you for this. It's one of the best-investigated reports we've had: you traced it from a 310-file data loss all the way to the exact inversion in `_retainUnavailableInstalledModules()`, where a **broken** custom module is preserved but a **correctly-installed** one gets deleted because its source is still cached. You then addressed review the next day and wrote regression tests that you verified go red when the behaviour is removed. That's a higher standard than most PRs meet.

I'm closing it, and I want to be straight about why, because it isn't a reflection on the work.

**The installer is being replaced.** BMAD is moving to a flat `skills/` tree installed through `npx skills` and plugin marketplaces, and `tools/installer/` — including `ui.js` and `core/installer.js`, both files this PR patches — is removed entirely in that change. `getDefaultModules()` ceases to exist, so the specific defect goes away with it. Asking you to rebase onto a tree that's about to be deleted would waste your time.

**But the failure mode is worth carrying forward**, and this is where you could genuinely help if you're willing. The underlying shape of the bug is: *an update path that computes a keep-list from a catalog will silently delete anything not in that catalog.* The replacement has a different shape — it discovers installed modules from disk rather than deriving a keep-list — but it does still remove things during update. If you wanted to point your same reproduction at the new setup path once it lands, that would be the most valuable review it could get. No obligation at all.

Sorry for the wait after the earlier "resolve the conflicts and we can merge this" — the distribution decision landed after that, and you deserved to hear it sooner.

If you have questions, or want a heads-up when the new install path is up so you can take a look, please reach out in [Discord](https://discord.gg/gk8jAdXWmj) — genuinely happy to keep talking about this one.

---

PR #2713, opened 2026-08-12
fix(build): narrow implicit skill routing

### LINKED ISSUE #2708: bmad-build's description is too wide-ranging
**Describe the bug**
The description of the bmad-build skill is very wide-ranging, which causes coding agents to select it even when not appropriate.

**Steps to reproduce**
See screenshot below for an example where Codex triggered bmad-build in an inappropriate context. I just asked it to commit BMAD-installed files, and it instead started a review sub-agent because it tried to follow bmad-build's workflow. This happened with my second message after updating BMAD to 6.11.0 on the repo.

Image

For what it's worth, after I asked Codex to explain why it thought that invoking bmad-build here was appropriate, its response was:

> • My faulty reasoning was:
> 
>   1. I saw “commit” as a repository mutation.
>   2. I overgeneralized the bmad-build description—especially “any … change request”—to mean any operation changing repository state.
>   3. I failed to distinguish:
>       - changing repository content: build/fix/refactor → bmad-build
>       - recording existing content: stage/commit/tag/push → direct Git workflow
> 
>   4. After invoking it, I followed its rendered workflow mechanically instead of recognizing that reviews, specs, and tests were evidence the skill did not fit the
>      request.

**Expected behavior**
bmad-build's description should make it clear that it should only be invoked when doing BMAD development. Other non-BMAD tasks should not result in it being invoked, at least not without consulting the user.

**Environment (if relevant)**
- Model(s) used: gpt-5.6-sol
- Agentic IDE used: Codex
- BMad version: 6.11.0
- Project language:

**Additional context**
I'm currently working around this by disabling implicit invocation of bmad-build by codex, but this shouldn't be necessary.

sanmaxdev:
## What

Narrows the `bmad-build` description so implicit routing requires an explicit BMAD request or the skill name.

## Why

Generic Git operations that only record existing content should not enter the build workflow. Fixes #2708.

## How

- Frame the skill as BMAD's build workflow.
- Replace the broad any-code trigger with explicit BMAD intent.

## Testing

- `npm ci && npm run quality`
- `node tools/validate-skills.js --json src/bmm-skills/ship/bmad-build`

### DIFF
--- a/src/bmm-skills/ship/bmad-build/SKILL.md
+++ b/src/bmm-skills/ship/bmad-build/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: bmad-build
-description: 'Implement any requirement, story, bug fix, or change request as working code that follows the project''s existing architecture, patterns, and conventions. Use when the user wants to build, fix, tweak, refactor, add, or modify any code, component, or feature'
+description: 'Implement a software change through BMAD''s build workflow, from intent clarification through review and verification. Use when the user explicitly invokes bmad-build or asks BMAD to build, fix, refactor, add, or modify code'
 ---
 
 Run the following command exactly once without changing the current working directory. Replace `{project-root}` with the absolute path to the project root and `{skill-root}` with the absolute path to this skill's directory:


alexeyv:
A couple of thoughts:

1. The problem is real, particularly on Codex.

2. This may be swinging a bit too far in the opposite direction; I do want the thing to auto-trigger in response to, say, a story tracker URL.

3. We also need a similar move for bmad-review.

alexeyv:
Thank you for this, @sanmaxdev — the diagnosis was exactly right, and the mutate-vs-record distinction your PR drew out of #2708 shaped the eventual fix. The problem is real and worst on Codex.

The one thing we wanted to keep that this version gives up is implicit triggering on handed-over work items — a bare story or issue link should still enter the build workflow without BMAD being named. #2714 supersedes this PR with a description gated on delegation rather than explicit invocation, validated against a 26-case routing eval (including your commit scenario) on two agent platforms.

Closing in favor of #2714 — much appreciated, and the eval cases now guard the boundary you found.

PR #2736, opened 2026-08-15
feat: add npx skills distribution for core+bmm

alexeyv:
## Why

`npx skills` is the one-command install that already reaches every coding agent that knows what a skill is. Method skills cannot go through that door today: they live nested under `src/core-skills` and `src/bmm-skills/{agents,plan,ship}/`, and they all depend on an installer-built `{project-root}/_bmad`. A raw `npx skills add bmad-code-org/BMAD-METHOD` would miss the skills or land them broken.

This epic adds a second, thinner front door. `npx bmad-method install` stays the full product.

## Epic

Someone with any `npx skills`-supported agent can:

1. `npx skills add <local-flatten-dir> --skill '*'`
2. run `bmad-help setup`
3. run `bmad-prd` or `bmad-build` and have `resolve_customization.py` / `resolve_config.py` succeed against `{project-root}/_bmad/scripts` whether that path is a symlink or a copy

Skipping step 2 fails at the script path with file-not-found. After a later `npx skills update`, setup fills new config keys, replaces the baked catalog, and either repairs the symlink or recopies scripts. `custom/` and `*.user.toml` stay put. Other skills do not bootstrap, upgrade, or open a sibling help skill.

## Stories

- [x] **1 — Packager flatten and fat help payload.** New Python (`tools/package_npx_skills.py`, not `tools/installer`) emits `skills/<canonical-id>/` from core+bmm Method sources. Only dest `bmad-help` carries the shared Python, module.yaml defaults (including the bmm agent roster), and a baked core+bmm help catalog.
- [x] **2 — First `bmad-help setup` and softer config load.** `load_central_config` treats missing `config.toml` as not bootstrapped. Setup materializes `_bmad` from the help payload: scripts as a symlink if allowed else a copy, empty `custom/`, catalog, output folder. Other skills keep failing at `uv run _bmad/scripts` if this has not run. No gate sentence.
- [x] **3 — Idempotent setup repair and preserve user layers.** A second setup and the post-`npx skills update` path fill new config keys, keep existing answers, replace the baked catalog, and never touch `custom/` or `*.user.toml`. Repair a wrong/broken scripts symlink, or recopy when a copy is not byte-identical. Last writer wins on a classic installer `_bmad`; no migrator.

## Out of this epic

Replacing `npx bmad-method install`. Growing `tools/installer`. External modules (CIS, GDS, TEA, WDS). Agent extras, `plugin.json`, `.claude-plugin`. Publishing a remote tree. A first-run interview, VERSION stamp, or gate sentences in other skills.

Draft because stories 2 and 3 are still open.


alexeyv:
Superseded by #2768 — the branch outgrew this description (packager removed, plugins and remote publishing now in scope), so the PR was rewritten rather than edited.

PR #2750, opened 2026-08-16
chore: default persistent_facts to an empty array

bmadcode:
## What

Sets `persistent_facts = []` in every shipped `customize.toml`.

## Why

These files shipped pre-seeded with `"file:{project-root}/**/project-context.md"`, which made loading project-context an opt-out default baked into every skill rather than a customization the user chooses. `persistent_facts` is a user-customization surface; it should start empty.

## Notes

- Comments, spacing, and every other key are untouched — only the array value changes.
- Files that already had `persistent_facts = []` were left alone.


bmadcode:
Follow-up commit pushed addressing the review.

**Stale default-loading comments — fixed.** Seven `customize.toml` comment blocks, plus `bmad-architecture/SKILL.md`, `bmad-create-story/SKILL.md`, `docs/how-to/expand-bmad-for-your-org.md`, `docs/reference/build-auto.md`, and the fr/vi-vn/zh-cn translations of the two doc lines.

Worth noting these were wrong twice over: the array now ships empty, **and** `bmad-project-context` no longer produces a `project-context.md` at all — it writes a verified block into `AGENTS.md` and explicitly treats `project-context.md` as a legacy artifact from the retired skills. So the comments named a default that no longer loads, produced by a skill that no longer emits it.

The replacement text states the actual model: repo-wide context belongs in `AGENTS.md`, which every skill already sees; `persistent_facts` carries context only one skill needs, loaded on demand rather than as constant memory. Each site shows the `file:` entry a user can add to opt back in.

**Dev-story guardrails — declining, they are not lost.** All three still live in `src/bmm-skills/agents/bmad-agent-dev/customize.toml:48-51` under `principles`, untouched by this PR. The copy emptied here was a duplicate inside `bmad-dev-story`, a deprecated v6 shim superseded by `bmad-build`. They were also miscategorized in that slot — those are agent principles, not persistent facts, which is exactly why the surviving copy sits under `principles`.

Separately, `bmad-dev-story/SKILL.md:59` and `bmad-code-review/SKILL.md:50` load `project_context` as their own hardcoded workflow variable, independent of `persistent_facts`. Unaffected by this change, left alone.

Not in scope: `docs/{fr,cs,vi-vn,zh-cn}/explanation/project-context.md` and `docs/{fr,cs,vi-vn}/how-to/project-context.md` still describe the old auto-load model, but their English sources were already rewritten around AGENTS.md — that translation drift predates this PR and belongs in its own change.

PR #2758, opened 2026-08-18
feat(skills): unify keep-or-dismiss and severity into one five-way triage verdict

alexeyv:
## What

Replaces the two-judgment review triage (assign severity, then keep-or-dismiss) with a single five-way verdict per finding, in the same wording across bmad-build-auto, bmad-build (incl. one-shot), and bmad-code-review — each keeping its own routes and bookkeeping.

For every finding the triager verifies the claim at the cited site — reading past the diff into callers and upstream guards until it can answer yes or no — and renders exactly one verdict:

- **`high` / `medium` / `low`** — the bad outcome is real, graded by how much it hurts end users or developers. Developer-only harm must name where it will bite; a vague "this is messy" is not gradeable.
- **`false`** — the bad outcome does not happen at the cited location. Requires a written refutation of the finding's own claim; a true fact about nearby code that leaves the claim standing does not count.
- **`maybe-false`** — could not tell. Records what would settle it. Goes through root-cause grouping like any other finding; an entry defers only when every member is `maybe-false`, and a `maybe-false` grouped with a verified finding rides its route.

Then, in order: reject `false` on its refutation; reject `low` when users are unlikely to meet it in everyday use *and* the fix adds guards, branches, parameters, or other complexity; reject anything whose fix edits the spec under build; group survivors by shared root cause; route entries (intent_gap / bad_spec / patch / defer, unchanged). Every finding gets one row in a per-pass triage ledger, reconciled against what the layers reported.

Also, from #2729 (closed as superseded): findings are judged on whether the problem is real, never on whether the proposed fix is plausible; a loud failure on a state the triager could not demonstrate is handling, not a defect; `patch` requires the smallest fix to be trivial, add no public surface, and guard no undemonstrated state — otherwise it routes to intent_gap or bad_spec; patch handoffs ask for the smallest fix. A finding excluded only by the spec's scope section, the plan, or the diff's shape (not by the intent) routes to intent_gap or bad_spec, never patch or defer.

## Why

Keep-or-dismiss and severity are the same judgment made twice from the same verified facts, and keeping them separate is what lets them disagree. Replaying a five-story epic through the old pipeline (219 raw findings) and auditing both triages found: 10.5% of findings silently dropped with no decision at all; 13% of recorded dismissals not withstanding inspection (a claimed verification that never happened, a straw-man restatement, an ignored probe); unverified claims binned as noise; and dev-facing findings with zero end-user consequence erased because the user-only scale had no home for them.

Under the verdict scheme: a dropped finding is a visible ledger count mismatch; a dismissal is a `false` verdict carrying a checkable refutation; an unchecked claim cannot be `false` and `maybe-false` costs an honest "what would settle it"; and developer harm has a grade, gated by the named-site requirement.

## Validation

**Review bench, `ui@970a801a0`** (3 ground-truth defects; same commit, protocol, and pinned `claude-opus-5` as every prior arm) — `review-bench/_runs/run_ui_2026-08-28_pr2758-r1/`:

| | h3-r1 (this PR's ancestor) | this PR @ 773acf47 |
|---|---|---|
| Judge precision / recall | 1.00 / 1.00 | 1.00 / 1.00 |
| Valid / false positives / rejected | 7 / 0 / 6 | 11 / 0 / 4 |
| Judge-found misses | 1 | 0 |
| Ground truth A / B / C | high / medium / — | high / medium / — |
| Reviewer tokens / wall | 599k / 946 s | 503k / 790 s |

Fully compliant run: zero human turns, all 335 assistant messages pinned, three layers launched in one message. Every `false` rejection carried a refutation and all four survived the judge. Root-cause grouping folded the falsified "parity" comment claim into Bug B's entry rather than shipping a separate nit. The one judge-found miss of the ancestor run (the "No fill" sentinel mismatch) was found and kept. Zero `maybe-false` verdicts out of 22 raw findings. Residuals are pre-existing and unrelated to triage: Bug B one notch light (`high` spent on the process abort), Bug C (a test that does not compile) invisible to every layer and to the judge.

`npm run quality` — 0 errors. Prompt-only change; no deterministic code paths affected.

## Follow-up (not blocking)

- `deferred` in the spec frontmatter now carries all-`maybe-false` entries alongside verified pre-existing work. An orchestrator that sweeps `deferred` as a work queue would schedule unverified claims; a `verdict:` marker on the item or a separate list would fix that. Volume was zero in the bench run.
- `followup_review_recommended` counts only patched entries by verdict; `maybe-false` findings never enter the patch count, so the recommendation can come out lower than under the old scheme for the same underlying findings.

### DIFF
--- a/src/bmm-skills/ship/bmad-build-auto/spec-template.md
+++ b/src/bmm-skills/ship/bmad-build-auto/spec-template.md
@@ -73,9 +73,9 @@ deferred: [] # append-only machine-readable deferred review findings; each item 
 ## Review Triage Log
 
 <!-- Append-only. Populated by step-04 on EVERY review pass, including loopbacks and blocked exits.
-     Each entry records triage decision counts for intent_gap, bad_spec, patch, and defer, with
-     per-category severity breakdowns using low/medium/high, plus each dismissed finding with its
-     reason and the findings addressed in that pass. Empty until the first review pass. -->
+     Each entry records verdict counts (high/medium/low/false/maybe-false) and one row per
+     reviewer finding: verdict, route, and evidence — the refutation for false, what would settle
+     it for maybe-false, the action taken for patches. Empty until the first review pass. -->
 
 ## Design Notes
 
--- a/src/bmm-skills/ship/bmad-build-auto/step-04-review.md
+++ b/src/bmm-skills/ship/bmad-build-auto/step-04-review.md
@@ -28,50 +28,53 @@ Announce skipped layers first, then launch every active layer before handling an
 
 ### Classify
 
-1. Once every layer has reported — and not before — render a verdict on each finding on its own, ahead of any deduplication or grouping. For each finding:
-   - **Verify its own claimed consequence** at the location it names. Read past the diff hunk — into the callers, the guards upstream, whatever else the site depends on — far enough to tell whether that consequence actually occurs. Another finding's outcome, however adjacent, never settles this one.
-   - **Assign severity** from the verified consequence for the artifact's main consumer (software user, document reader, etc). Disregard any severity assigned by a reviewing subagent. Review subagents operate under by-design information asymmetry and do not have enough context to set final severity for this workflow.
-     - `low`: none or cosmetic
-     - `medium`: tolerable
-     - `high`: intolerable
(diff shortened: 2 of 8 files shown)


alexeyv:
Applied Greptile's P1: code-review's `patch` route now carries the same fix-shape test as build/build-auto (unambiguous, no new public surface, no guards on undemonstrated state; otherwise `decision_needed`).

Declining the rest:
- **bad_spec unreachable** — the spec-edit reject is about findings whose fix is to edit the spec text; `bad_spec` is a code defect rooted in the spec. Both were already on main.
- **Triage log appended twice** — step 4 defines the entry, the route branches say when to write it. Same wording as main.
- **One-shot HALT before log write** — HALT pauses for the human; Finalize Spec writes the log afterward. Same flow as main.
- **Defer placeholder** — the format lists example reasons; the third is inferred.
- **batch-related** — "batch related ones" means *batch the related ones*; hyphenating changes the meaning.

PR #2795, opened 2026-08-29
fix(scripts): pin stderr to UTF-8 in the config scripts

aranellaeth:
## Problem

The three scripts under `skills/bmad/scripts` write diagnostics that quote user-controlled text:

- `resolve_config.py:65` and `resolve_customization.py:142` — `f"error: {error}\n"`, where the message carries the offending config **path** and tomllib's parse error.
- `memlog.py:153` — quotes `path`; `memlog.py:159` — quotes the `--field key=value` pair straight from argv.

Only **stdout** is pinned (`write_json_stdout`). stderr is left at the platform default, so on a Windows console (cp1252) the filename the user needs in order to fix the problem comes back as escapes.

Measured, same run with and without the pin:

```
with:     error: failed to parse ...\proje-şık\_bmad\custom\bmad-fake.toml: ...
without:  error: failed to parse ...\proje-şık\_bmad\custom\bmad-fake.toml: ...
```

## Change

Generalizes the `pin_utf8` helper already merged for `brain.py` in #2578 to the three scripts, and calls it on stderr at `main()` entry. `write_json_stdout` now goes through the same helper instead of its own inline `reconfigure`.

`errors=` is passed through deliberately: `reconfigure(encoding=...)` alone resets the handler to `"strict"`, which would silently downgrade stderr's POSIX `backslashreplace` default and turn a diagnostic about an undecodable path into a traceback.

## Test

Adds one subprocess test that drives `resolve_customization.py` with `PYTHONIOENCODING=cp1252` against a project root whose path contains Turkish characters, and asserts the path appears readable in stderr with no traceback.

Verified it discriminates: with the `pin_utf8(sys.stderr)` line stripped the test fails on exactly that assertion; restored, it passes.

`skills/bmad/scripts/tests` — **82 passed** (81 before, +1 new).

Diff is +105 / -7 across three scripts and one test file.

Line numbers refer to `dev` at 23f134e2.


aranellaeth:
Pushed `ff78f08` to address the docstring coverage check, and it turned out to be covering a real slip rather than a style gap.

The first commit folded `write_json_stdout`'s body into the new `pin_utf8` helper and dropped its existing docstring in `resolve_config.py` along with it:

```
-def write_json_stdout(output) -> None:
-    """Pin stdout to UTF-8 — a Windows cp1252 default cannot encode emoji icons."""
```

That was unintended. It is restored, and its sibling in `resolve_customization.py` now carries the same. The follow-up also documents `main()` in the three scripts and the two test entities this PR touches.

Docstring coverage over the touched set goes from 4/13 to 11/13 (84.6%). The follow-up is +8/-0, pure additions. `src/scripts/tests` still passes 42.

The two test methods that remain undocumented predate this PR and I did not touch their bodies, so I left them alone rather than widening the diff.

aranellaeth:
Rebased onto `dev` (23f134e2) and force-pushed.

- The conflicts in `resolve_config.py` and `resolve_customization.py` came from the ruff reformat in #2830. They were formatting only; the `pin_utf8(sys.stderr)` call sits on top of dev's reformatted `main()`.
- 2d931ae7 brings the new test in line with the ruff config: ruff format, and `capture_output=True` in place of the two `PIPE` arguments (UP022).
- `pre-commit run` on the changed files passes (ruff check, ruff format, validate, pytest). The stderr test still fails with the pin removed.
- Paths, line numbers and test counts in the description are updated to the `skills/` layout.

alexeyv:
Thank you for tracking down the stderr side of this. Closing as superseded by #2920, which reached `dev` with #2917.

#2920 pins both `stdout` and `stderr` to UTF-8 on Windows in the `__main__` block of every script that runs on its own, 26 in all. That includes `memlog.py`, `resolve_config.py`, and `resolve_customization.py`. #2917 added a native Windows job to CI, so these scripts and their tests now run on Windows on every pull request.

PR #2800, opened 2026-08-31
fix(review): count incomplete early exits as unhandled paths

alexeyv:
## Problem

The edge-case-hunter prompt tells the reviewer to "discard handled ones silently" — a pure presence check. A branch whose handling exists but is incomplete passes that check and its analysis ends right there, so a changed condition or early return that exits with required work undone (cleanup skipped, lock held, state half-updated) is never reported.

## Change

In all four copies of the prompt — `bmad-build`, `bmad-build-auto`, `bmad-code-review` review-prompts and the `bmad-review` edge-case lens:

- Fold reachability and left-undone work into the definition of "handled": *"For a changed condition or early return, handling includes what can reach it and what remains undone when it returns — a path that exits with required work undone is unhandled."*
- Narrow the discard rule (preamble and collect bullet) from "handled" to "fully handled", so the discard cannot preempt the new check.
- Extend the existing scope carve-out so the check can be executed as written: *"…unless the provided content explicitly references external functions, or a changed early exit requires its callers and the code it skips."*

The check runs as part of determining "handled" rather than as a sibling instruction, because a sibling instruction would be neutered by the discard happening first. The third bullet closes a conflict the first two introduced: "what can reach it" means the callers, which the old carve-out never covered (callers reference the changed code, not the reverse), and the code after an exit is neither in the hunk nor an external function. Extending the same sentence keeps hunk scope governing everything else rather than adding a rule.

Scope stays mechanical (reachability, completeness of exit) — general "handling is wrong" correctness review remains out of this reviewer's lane. The v6 shim only forwards to the lens, so it needs no change.

`npm run quality` passes (0 errors).

### DIFF
--- a/src/bmm-skills/ship/bmad-build-auto/review-prompts/edge-case-hunter.md
+++ b/src/bmm-skills/ship/bmad-build-auto/review-prompts/edge-case-hunter.md
@@ -3,7 +3,7 @@
 **Goal:** You are a pure path tracer. Never comment on whether code is good or bad; only list missing handling.
 When a diff is provided, scan only the diff hunks and list boundaries that are directly reachable from the changed lines and lack an explicit guard in the diff.
 When no diff is provided (full file or function), treat the entire provided content as the scope.
-Ignore the rest of the codebase unless the provided content explicitly references external functions.
+Ignore the rest of the codebase unless the provided content explicitly references external functions, or a changed early exit requires its callers and the code it skips.
 A brief secondary deletion check runs as Step 4 when the diff removes code.
 A claims check runs as Step 5.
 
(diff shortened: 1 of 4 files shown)


alexeyv:
Closing unmerged. Isolated-layer A/B on the review-bench commit (9 runs: 3 control, 3 this change, 3 with the follow-up scope carve-out) showed no effect on discovery — every arm found ground-truth Bug B 1/3 — and no run's trace ever stopped at a guard, which is the failure mode this rule targeted. Superseded by #2801, whose handle-lifetime rule addresses the same bug class with measured mechanism-level evidence. Run artifacts: review-bench/_runs/run_ui_2026-08-30_ech-earlyexit-ab/.

PR #2804, opened 2026-08-31
fix(memlog): serialize concurrent writes with crash-safe lock recovery

### LINKED ISSUE #2621: `memlog.py`: concurrent appends can lose entries

onbermejo:
Fixes #2621. Concurrent `memlog` appends could interleave their read-modify-write and silently drop entries.

This serializes the full read/modify/replace cycle with a sidecar `.lock`:

- ownership tokens, so a writer only ever removes the lock it acquired;
- a persistent `.lock.guard` advisory lock that serializes lock creation/reclaim/removal (kernel-released on crash);
- stale-lock reclamation after a bounded lease, with inode + content revalidation to avoid ABA races;
- handling for an empty/not-yet-published guard, operator deletion mid-open, hard-link pre-seeding, and transient Windows handle retention;
- bounded waits with actionable timeout messages.

Adds 42 concurrency/recovery regression tests and a cross-platform (Linux/Windows) CI job. Rebased on current `main`.

Supersedes #2622, which carried only the initial serialization commit; this also addresses the earlier review notes about stale-lock replacement and deleting a replacement lock.


onbermejo:
Thanks both — addressed in 19ab0eab.

**Greptile P1 — empty guard blocks recovery (valid, fixed).** A `.lock.guard` left zero-length by a creator that died between the `O_EXCL` create and publishing its byte no longer wedges writes permanently. A guard that stays empty past the wait deadline is reclaimed and recreated instead of every later write timing out forever. Reclamation is safe: a zero-length guard never has an advisory-lock holder, so removing it cannot split holders, and a live creator now re-checks that the pathname still resolves to the inode it created before returning, so it cannot proceed on a guard reclaimed underneath it. The handle is closed before the unlink because Windows will not unlink an open file. New regression: `test_empty_guard_from_a_dead_creator_is_reclaimed`.

**Greptile P2 — Windows locking untested in CI (valid, fixed).** Added a `memlog-windows` job (windows-latest) running `npm run test:memlog`, so the `msvcrt.locking` acquire/release branch is now exercised in CI.

**CodeRabbit — writer outliving the 5-minute lease.** This is the same bounded-lease trade-off the lock uses deliberately: a healthy memlog write is a sub-millisecond local-file op and must never hold the lock for minutes, so a >5-min hold means a hung/suspended process — and the lease is exactly what lets a genuine crash recover. Kept as a documented trade-off rather than adding non-portable process-liveness checks.

**CodeRabbit — "guard-file replacement race can lose entries."** An initialized (non-empty) guard is never unlinked or replaced; it is persistent by design precisely to avoid splitting waiters across inodes, so there is no replacement race for a live guard. The only guard ever removed is a zero-length remnant with no lock holder (above). Lock reclamation is serialized under the guard and revalidates inode + token before unlinking (`test_orphan_reclaim_is_serialized_with_successor_acquisition`).

All 43 memlog tests pass on Linux and Windows locally.

onbermejo:
Good catch — fixed in 9e6dcf49. `_reclaim_zero_length_guard` now returns whether the guard is actually gone; the caller retries only after a successful reclaim and otherwise raises the bounded `TimeoutError` instead of resetting the deadline and looping. So an empty guard that cannot be unlinked (a handle held open on Windows, a permission error) now fails with an actionable timeout rather than wedging every write. Added `test_unremovable_empty_guard_times_out_instead_of_looping`; 44 memlog tests pass.

onbermejo:
Valid, and it made me stop patching and look at the root cause — fixed in 0129a06c by removing the mechanism instead of guarding it.

**Why the race existed at all.** The guard had a two-step birth (create the pathname, then publish an initialization byte). That is the only reason an "empty guard" state exists, which is what forced a reclaim path, whose `unlink` is exactly what could split a waiter and a publisher across two inodes. Each fix so far was patching a symptom of that byte.

**The byte was never needed.** I verified on both platforms that a zero-length file is fully lockable and still mutually exclusive:

- `fcntl.flock(fd, LOCK_EX|LOCK_NB)` on an empty file: acquires.
- `msvcrt.locking(fd, LK_NBLCK, 1)` at offset 0 on an empty file: acquires, and a second handle to the same empty file is refused with EACCES.

So guard acquisition is now a single `O_CREAT` open plus the advisory lock, the guard stays contentless, and it is never unlinked. That closes all three findings by construction rather than by check:

- **empty guard from a dead creator** — it is immediately usable, so there is nothing to recover;
- **unremovable empty guard** — there is no unlink left to fail or loop on;
- **guard reclamation splits lock inodes** — nothing is ever reclaimed, so every writer locks the same stable inode.

Net effect is ~130 fewer lines. Regressions now assert the invariants directly instead of the old dance: an empty guard is usable and keeps its inode across writes, a crashed writer leaves a reusable guard, and an operator-deleted guard is simply recreated. I also re-checked mutual exclusion explicitly (12 concurrent threads, peak concurrency 1) alongside the existing 50-process append test.

43 memlog tests pass on Linux and Windows.

onbermejo:
Fixed in 0302c4d9, and the report prompted a full audit of the lock lifecycle that turned up four more defects — three of them pre-existing in the acquisition path, not in the guard.

**On this finding.** The race is real and is closed: the held inode is re-checked against the pathname after the guard lock is taken, and the lock is retaken on the current guard if they differ.

The stated consequence does not hold, though. The critical section is protected by an atomic `O_CREAT|O_EXCL` create on the lock record, not by the coordination guard. I disabled the guard entirely — strictly worse than any split — and ran 15 concurrent writers: peak concurrency stayed at 1. So a split guard cannot admit two writers into the read-modify-replace cycle or lose an entry. What it genuinely degrades is reclaim serialization, which is why it is worth closing regardless.

**What the audit found beyond it.**

1. *Guard deleted between the open and the identity check* — raised an uncaught `FileNotFoundError` and failed the application write. My own regression from the previous rework; that window is transient and is now retried within the bounded wait.
2. *Lock acquisition could livelock* — a successful reclaim extended the deadline unconditionally, so a repeatedly reclaimable record deferred the timeout forever and `exclusive_lock` never returned. Reproduced: it hung indefinitely instead of raising. Reclaim-driven extensions are now capped by `MAX_LOCK_RECLAIMS`. This is the same unbounded-reset pattern flagged earlier in the guard, in the pre-existing path.
3. *Lock record and descriptor leaked on an unexpected failure* — any error after the `O_EXCL` create (e.g. while reading the descriptor's identity) left the descriptor open and the record on disk. On Windows the open handle also made the record undeletable, so every later write was wedged until the process exited. Both are now released before the error propagates.
4. *Release masked the caller's exception* — a coordination-guard timeout while dropping the lock replaced the body's error with a `TimeoutError`, so the caller lost its real failure. Release is now best effort when the body already failed; the lease reclaims anything left behind.

**Verification.** 50 memlog tests pass on Linux and Windows (seven added here). Each of the findings above was reproduced first and then re-checked against the final code. Also verified explicitly: no descriptor leaks across any error path (80 cycles including failures), every wait bounded, mutual exclusion holds at 16 concurrent writers (peak 1, zero errors), early `return`s inside the lock release cleanly, and `write_atomic` is byte-identical to `main`.

alexeyv:
This is a massively over-engineered solution to a one in a million years problem. Thanks, but no.

PR #2805, opened 2026-08-31
Review triage economics: drop the finding quota, redraw patch vs defer, one ledger contract

pbean:
## Why

Every review layer in `bmad-build`, `bmad-build-auto` and `bmad-code-review` ran a Blind Hunter subagent whose prompt carried a floor: find at least ten issues, and never stop with an empty list. A reviewer that must produce ten items on a clean forty-line diff produces ten items. Triage then paid to verify each one, the spec's triage log recorded each one, and the ones triage could not disprove but did not want landed in the deferred-work ledger — where the next sweep read them back as work.

The floor came from #2675, which credited the *question* — look for what is missing, not only what is wrong — and got implemented as a count. This branch keeps the question and drops the count, then redraws the line between `patch` and `defer` so that what the change caused gets fixed in the run instead of filed.

## What changed

Everything here is instruction text. The two step-04 review files stay in lockstep: shared sentences are byte-identical between `bmad-build` and `bmad-build-auto`, and only the human-present branches differ.

**1 — The hunter stops counting** (`f8ee78e4`, `b8012952`, `24bc95c4`, `81229443`)
All five copies of the Blind Hunter prompt lose the ten-finding floor and the never-stop-empty clause, and gain an evidence bar: file, line, the input or state that reaches it, the bad outcome. The what-is-missing second pass survives intact. Four copies are the review layers in `bmad-build`, `bmad-build-auto` and `bmad-code-review`; the fifth is `references/lens-adversarial.md` in `bmad-review`, which states the same bar in its own field shape — `location`, `trigger_condition`, `potential_consequence` rather than file-and-line — and drops "or improvement" from `guard_snippet`, the exact category the bar excludes. The `bmad-review` SKILL.md lens summary no longer advertises the old ten-finding contract.

**2 — Triage verifies cheaper and routes wider** (`f8ee78e4`, `b8012952`, `606f502d`)
- Carry forward: on a loopback, resumed review, or follow-up pass, a finding matching a logged row keeps that row's verdict and route when the code still reads as the row describes — written back as `carried`, never re-verified, never patched or deferred twice.
- Verify from the evidence a finding cites; trace from scratch only when it cites nothing. Findings naming the same site and outcome get one verification and separate rows.
- A finding with no demonstrated outcome is `false`. Triage does not build the claim on the reviewer's behalf.
- `patch` becomes a locality test, not a size test — the blast radius is named explicitly (files in the diff, their tests, code a test in the diff exercises, direct callers of changed symbols, sibling sites of a pattern the diff fixed).
- `defer` becomes a five-part positive test. All-`maybe-false` groups are rejected with what would settle them, not deferred. Out-of-scope routes on merits, never to defer.
- Verification gaps are patches: a test that would still pass with the change reverted, or a touched site that skips the helper the change introduced. Step-03 carries the same bar as its definition of done.

**3 — One ledger contract** (`24bc95c4`)
New `references/deferred-work-entry.md`, byte-identical in `bmad-build` and `bmad-code-review` and pinned by `test-template-sync`, becomes the only way to write `deferred-work.md`. Read the whole ledger first; an open match on location and substance gets a `seen-again:` line instead of a new entry; a closed match gets a new entry naming the closed id. Entries are `### DW-<n>` blocks; ids mint from the highest existing and nothing is renumbered. Five write sites point at it.

**4 — The loop stops paying twice** (`2bafb3fb`, `3933e297`, `14215d54`)
- Re-engagement is defined: continue the step-03 implementer, addressed by the name or id its launch returned. A fresh launch is not re-engagement. The patcher runs only the tests covering the files it edits and returns; full verification belongs to step-04.
- One `## Review Triage Log` section per spec, ever — append to the one that exists; if two exist, merge into the earlier heading first, preserving every row.
- Step-01 sets `followup_pass` on every re-dispatch of a `done` spec; a follow-up recommends another pass only when it patched a `high`, and must name the unverified risk. Patch volume is never grounds.
- Step-04 rewrites the diff file after patching so step-05 reads it instead of rebuilding it.

Docs and tests move with the skills: `build-a-change` describes the ledger contract, dedupe and carry-forward; `review-a-change` describes the five-way verdict; `test-template-sync` generalises to a cross-skill vendored-copy check; the renderer test asserts the reference is published and referenced from every write site.

### A/B on the hunter prompt — 20 context-free runs

Old and new prompts against two diffs: a real one, and the same diff with exactly one line changed (a proof-of-work ternary inverted — a defect that turns nine tests red). Five runs per cell, all `claude-opus-5`, model confirmed per transcript. Old prompt read from `git show main:`, new from the branch working tree.

| Condition | Runs | Mean | Plant found |
|---|---|---|---|
| Old · clean diff | 21, 24, 20, 19, 16 | 20.0 | — |
| New · clean diff | 2, 0, 2, 4, 1 | 1.8 | — |
| Old · planted defect | 20, 19, 21, 22, 16 | 19.6 | 5 / 5 |
| New · planted defect | 1, 1, 1, 1, 1 | 1.0 | 5 / 5 |

The planted defect was found in all ten runs that saw it. Under the old prompt it arrived first in a list of 16–22 items; under the new prompt it arrived alone, and four of five runs proved it by running the suite (nine tests red as written, 333 green with the operands swapped). The new prompt's clean-diff findings were real defects, not noise — one was independently rediscovered and landed later on the downstream repo's main. One run returned an empty list and said it had rejected stale-prose near-findings. Contamination screen clean: no reviewer read the planting script's body or cross-compared the diffs.

### Finding quality — from the sweep specs' own triage logs

19 review passes under the old skill, 12 under the new, across ten stories.

| Measure | Old | New |
|---|---|---|
| Findings graded per pass | ≈ 21 | ≈ 19 |
| Verified real per pass (high+med+low) | 7.9 (150) | 14.2 (171) |
| Real share of graded findings | 38% | 74% |
| Rejected per pass | 13.2 | 5.0 |
| `high` verdicts per pass | 0.63 (12) | 0.25 (3) |
| Entries routed `defer` | 51 | 0 |

Volume barely moved; composition flipped. The new skill surfaces and fixes more verified-real defects per pass while triage spends far less on refutation. Deferrals per story fell from ~7 to 0.5, and the follow-up pass converges — recommended at most once per story, every one resolving to `false`.

### Session cost

Measured from the loop journal's `session-start` / `session-end` pairs across 42 sessions:

- Review session, median: **24m41s → 26m56s** (n=15 then 7) — sessions did **not** get shorter.
- Review session, median weighted tokens: **545k → 382k** (−30%).
- Second review pass per dispatch (`review-2`): **6 sessions → 0** — median 23m01s each, 2.3 h and 3.6 M weighted tokens the new era never spent.
- Patch phase inside a review session: the fresh-subagent case that prompted `3933e297` cost 20 of that run's 38 review minutes; after the change, three sessions ran 3m55s, 5m12s and 9m21s, and 3 of 3 continued the live implementer rather than launching a patch agent.

The saving is a session that stops being scheduled and a third of the review session's tokens — not a shorter session.

## Limits

- The A/B is five runs per cell on one diff pair, one repo, one model, and tests recall against exactly one seeded defect. The old prompt's 16–24-item lists were never graded item by item, so a real defect the new prompt is silent about would not show up.
- Field numbers come from one sweep run whose stories differ in size and subject; the eras are not matched work. Seven new-era review sessions can say no large duration saving appeared, not resolve a small change either way.
- The `high` rate fell 12 → 3. The counts are too small to separate from the stories, and these commits also changed what `high` means (a vague "this is messy" with no named harm can no longer be graded at all), so part of that is definitional. When highs were present the new skill caught them — the rearm pass patched three verified `high` findings.
- The quality tally counts what triage concluded, not what was true. A `false` verdict is a refutation by the same step these commits rewrote; nothing here audits those refutations independently.
- The two eras log differently: the old format grades only surviving findings and writes rejects as prose, the new grades everything. Counting a dismissed bullet as the rejected finding it is makes the share comparable, but it is not a verdict-for-verdict match.

## Testing

Full `npm test` green, including the generalised `test-template-sync` (both vendored copies byte-identical to their sources) and the renderer assertion that `references/deferred-work-entry.md` is published into the snapshot and referenced from every ledger write site.


alexeyv:
Closing this one — not because it didn't land, but because it landed in pieces. Working through this PR carefully turned into a small harvest: four PRs came out of it, all merged, all carrying your co-author line.

- #2809 — the maybe-false triage gate: a maybe-false whose claim, if true, would be `medium` or `high` is deferred with its severity marked unverified, instead of quietly rejected. Grew straight out of this PR's triage-economics argument.
- #2811 — `followup_pass` gating in bmad-build-auto: a follow-up pass recommends another only when it patched a `high` and can name the specific unverified risk it leaves behind.
- #2812 — carry-forward: on a loopback or follow-up pass, triage checks each finding against the existing Review Triage Log and carries matching verdicts instead of re-verifying and re-deferring the same things.
- #2813 — the re-patch route: re-engagement defined (continue the same step-03 subagent; a fresh launch is not re-engagement), a fixed letter to the implementer with tests scoped to the files it edits, and the diff refreshed after patching.

The one piece not taken is the finding-quota change itself. We benchmarked the two branches head-to-head on a large diff and the results convinced us stopping rules deserve their own investigation rather than a verdict smuggled in here — that conversation continues elsewhere.

Thank you, @pbean — genuinely. This PR did what the best PRs do: it made someone read the code hard enough to find four improvements, and it supplied most of them itself. Much appreciated.

PR #2854, opened 2026-09-10
docs: refresh the README banner, restore its diagram, and fix the favicon

do-operator:
Three things the READMEs and the site were carrying wrong.

### Banner

The four READMEs share one banner image. It is now the current one from bmadcode.com — the BMad tile and wordmark over the line art — at 2x for retina. It replaces a 1408×224 crop and is smaller on disk than what it replaces (127 KB vs 374 KB).

### The README's delivery-loop diagram

Broken since #2850 moved the diagrams to `docs-site/src/diagrams`. It cannot simply be repointed at the new path: an authored diagram carries geometry and classes only, so it needs the site's stylesheet to have any colour at all, and a README loads an SVG as an `` where no stylesheet can reach it.

So `docs/images/` now holds *exports* of the source, generated by `npm run export-readme-diagrams` with the dark ramp substituted in as literal colours. Dark rather than light because a README renders on whichever background the reader's GitHub theme picks, and only a drawing carrying its own ground is legible on both. Literals rather than custom properties because an export has to survive renderers thinner than a browser — resvg drops `var()` and paints the fallback black.

The Korean README's copy comes out of the same geometry through the existing labels file, so the two can no longer drift apart.

### Favicon

It was a teal `B` that matched neither the header tile nor blog.bmadcode.com. It is now the BMad mark, drawn from the same path data `SiteTitle.astro` already carries, on the same navy the blog uses. The `.ico` is generated from the SVG and stays listed for browsers that ignore `image/svg+xml`.

### Checks

`lint`, `format:check`, `build` and `test` all pass locally.


docs/images/bmad-delivery-loop.svg
alexeyv, line 1: image
font sizes in this diagram are wrong - smaller than the text surrounding them, some barely legible. Including even the diagram title.

alexeyv (changes requested):

docs/images/bmad-delivery-loop.svg
do-operator: You caught a real state, but it was already three hours stale when you looked — your screenshot is 08e7230b, and f7c03ee6 had scaled the type for a docs column by then. Bumped it further anyway, because you are right that it was still too small.

At your 930px, against GitHub's 16px body:

| | original | you saw | now |
| --- | --- | --- | --- |
| START ANYWHERE | 11.6px | 9.2px | 14.5px |
| Vague notion | 13.3px | 10.5px | 14.9px |
| Clarify / Plan | 17.4px | 13.9px | 17.4px |
| RIGHT-SIZED band | 14.1px | 11.2px | 14.9px |

Worth noting the middle column is the anomaly, not the baseline: the diagram this replaced ran the same scale as the right-hand column, so the objection would have applied to it too. It is now at or above the original everywhere.

The other half of what made it feel thin: the original marked the three entry points in gold, and when the diagrams gave up their own colours so one drawing could serve light and dark, that meaning went with it. Those three drops now carry the theme accent, which is what the vocabulary already uses for an edge the drawing has to single out — so it works on both ramps. The README export additionally gets a proper ground instead of a flat fill.

do-operator: Correcting myself: the last line above is out of date. The gradient ground I mentioned is gone — Ruben didn't like it, and it read as an effect rather than a surface. The export is back to a flat fill. The larger type and the accent on the entry drops stay.

One more change since, prompted by the same diagram feeling off. Measuring rather than eyeballing turned up something worse than uneven spacing:

| | before | after |
| --- | --- | --- |
| clearance below START ANYWHERE | **−3** | 40 |
| clearance above RIGHT-SIZED | 35 | 40 |

The title's baseline sat three units *below* the big bulb's top ray. They only looked separate because they are far apart horizontally, but the eye reads the vertical band, which is likely part of why the label felt wrong to you beyond its size. Both bands are now held 40 units clear of the nearest mark, on a canvas grown 420 → 460 so the drawing keeps its size.

Worth a fresh look if you have a moment — the rendered file in this PR is current.

PR #2889, opened 2026-09-17
docs(architecture): fix subagent-unavailable fallback wording in reviewer-gate.md

gmoncor:
What: Replace "write the file first, then flush it from context" with wording that describes continuing from the file's compact summary instead of presupposing selective memory control.

Why: The instruction asks the model to "flush" something from its own context, which is not a real mechanical operation; the subagent path in the same file already models the correct discipline (compact summary only, never holding full review text).

How:
- Reword the sequential-fallback sentence to describe writing the file first, then continuing using only its compact summary and file path in subsequent steps, without re-reading or re-quoting the full review text.

Testing: Read-through diff; prompt-text-only change, no automated test applies per this repo's testing policy for static source text.


pbean:
this is not a bmad-method issue but a codex environment issue please see my response from bmad-loop in regards to this: https://github.com/bmad-code-org/bmad-loop/issues/781#issuecomment-5666012512

Issue #2883, opened 2026-09-17
[6.12.0] Generated _config catalogues: skill-manifest.csv paths all resolve to nothing, and bmad-help.csv mixes two phase naming schemes


johnkattenhorn:
Two defects in the generated `_bmad/_config/` catalogues on a fresh 6.12.0 install. Both are install-time output rather than anything a user edited, and both matter to anything that reads the install programmatically (my case is a scheduler that reads the phase graph to decide what to run next).

**Repro** — one command, into an empty git repo:

```sh
npx bmad-method@6.12.0 install --yes --directory ~/Code/<project> \
  --modules bmm --tools claude-code
```

`_bmad/_config/manifest.yaml` then reads `installation.version: 6.12.0`. Environment: Linux, node via npx, `uv 0.12.10` (the installer's uv check passed).

---

## 1. `skill-manifest.csv` names 29 files and none of them exist

Every row points at `_bmad/core/<skill>/SKILL.md`. The installer wrote the skills to `.claude/skills/<skill>/SKILL.md` — the directory chosen by `--tools claude-code` — and created no `_bmad/core/<skill>/` at all.

```sh
python3 -c "
import csv, pathlib
root = pathlib.Path.home() / 'Code/<project>'
rows = list(csv.DictReader(open(root / '_bmad/_config/skill-manifest.csv')))
print(len(rows), 'rows,', sum(1 for r in rows if not (root / r['path']).exists()), 'paths missing')"
```

```
29 rows, 29 paths missing
```

```sh
$ ls ~/Code/<project>/_bmad/core
v6-shims
$ ls ~/Code/<project>/.claude/skills/bmad-brainstorming/SKILL.md
.../.claude/skills/bmad-brainstorming/SKILL.md
```

**Expected:** the `path` column resolves, or is documented as relative to something other than the project root.

**Actual:** it resolves to nothing, so the manifest cannot be used to locate a skill. A consumer that trusts it silently hands out paths to files that are not there.

I do not know whether the intent is that `_bmad/core/<skill>/` should have been created and was not, or that the column should name the tool directory. Either way the manifest and the tree disagree.

## 2. The `phase` column mixes two naming schemes

`_bmad/_config/bmad-help.csv`, 25 skill rows (excluding the `_meta` row):

| phase | rows |
|---|---|
| `anytime` | 12 |
| `plan` | 6 |
| `ship` | 5 |
| `2-planning` | 2 |

[...]

PR #2894, opened 2026-09-17
fix(skills): correct broken bmad-help skill references to bmad

gmoncor:
What: Replace the non-existent bmad-help skill-invocation instruction with the real skill name bmad across the 12 files that contain it as an actual invocation instruction.

Why: No skill or manifest named bmad-help exists anywhere in the repo; the real skill is skills/bmad/SKILL.md (frontmatter name: bmad). Agents following this instruction verbatim would attempt to invoke a skill that does not exist.

How:
- In each of the 12 listed files, change the invocation-language sentence from bmad-help to bmad.
- Leave the mentions in docs/*.md and CHANGELOG.md untouched, unrelated user-facing prose/history about product branding, out of scope here.

Testing: Repo-wide search for bmad-help restricted to the 12 touched files confirms zero remaining occurrences as invocation instructions; read-through diff for correctness.

Related to #2883, not fixing it: that issue is about the generated bmad-help.csv config catalogue having inconsistent paths/phase schemas, a different artifact and root cause than this dead skill-invocation reference.


gmoncor:
Closing as superseded: rebasing onto current dev shows the exact same `bmad-help` → `bmad` fix already landed upstream (independently, with different wording) in every one of the 11 conflicting files, so this PR's diff is fully redundant against dev today. Separately, my own scope claim didn't fully hold — a few live `docs/*.md` files (e.g. `docs/index.md`, `docs/start/get-answers-about-bmad.md`) still have the same broken `bmad-help` invocation and were out of scope here; happy to open a follow-up for those specifically if useful.

PR #2900, opened 2026-09-17
fix(prd): document a fallback for memlog.py write failures

gmoncor:
What: Add one fallback sentence to the File-roles line and to the memlog.py call sites stating that on failure, the entry should be appended inline tagged `[UNLOGGED: <type> <text>]`, to be reconciled into `.memlog.md` during the existing Memlog audit Finalize step.

Why: Every memlog.py call site in this skill has no documented fallback, unlike the resolve_customization.py call in the same file which already documents one; an unhandled failure silently drops a decision-trail entry.

How:
- Add the fallback clause to the File-roles line and to the Update/headless memlog.py call sites.
- Specify where the inline note is persisted so it survives session close.

Testing: Read-through diff confirming the fallback location is durable and that the existing Memlog audit Finalize step is referenced as the reconciliation point.

Related to #2621 (concurrent appends can lose entries) — not the same defect: #2621 is about inter-process lock loss on concurrent writes, this is about the missing documented fallback in the consumer skill when memlog.py fails or is unavailable. Both touch memlog.py reliability as the decision-trail record.


gmoncor:
Greptile's finding (Create fallback cannot reconcile when memlog.py init itself failed, not just append) is real but needs actual new branching logic in the Triage step, not a wording fix — higher risk of a follow-up round than the doc-precision issues already fixed here. Tracked in #2937 instead of iterating further in this PR. Leaving in draft until that's resolved.

PR #2903, opened 2026-09-17
fix(architecture): record whether the Reviewer Gate ran in spine frontmatter

gmoncor:
What: Add a reviewed frontmatter field to spine-template.md (default false, set true once the Reviewer Gate has run) and update reviewer-gate.md to set it when the gate runs and leave/set it false when explicitly skipped.

Why: A throwaway prototype may run the Reviewer Gate quietly or skip it entirely, but the finished spine document carries no visible marker of which happened.

How:
- Add reviewed: false next to the existing status field in spine-template.md.
- Update reviewer-gate.md to set the field to true when the gate runs, at any tier.

Testing: Read-through diff; additive frontmatter field, no existing consumer of spine frontmatter breaks since the field is new and optional.


bmadcode:
@gmoncor Thank you for the many PR's - they are all draft so they have not been read or reviewed - do you intent to remove draft tag from them or do they still need work?

gmoncor:
My intent was to enable them for review. Will give 'em a check this afternoon (token reset) and ensure I put for review the most interesting ones.

gmoncor:
Thanks for the review, @coderabbitai / greptile. Pushed a fix for both P1 findings:

- **Validate Has Conflicting Instructions**: scoped the `reviewed` frontmatter write to Finalize only. Validate cannot mutate the spine (per this file's own "under Validate you report and don't change the spine"), so it now reports the gate's outcome in its own review output instead of writing to frontmatter.
- **Skipped Review Marker May Vanish**: the skip path now requires writing `reviewed: false` explicitly (adding the field if the spine predates it), instead of the ambiguous "leave it (or set it) false".

gmoncor:
Fixed the first two greptile findings (Validate/Finalize contradiction, missing-field-on-skip) in the last push. The third — whether Update mode also needs to write `reviewed` — is a design question about how ## Update's 'close as in Finalize' delegation should compose with the reviewer-gate.md instruction, not a one-line fix. Tracked in #2938 rather than iterating further here. Leaving in draft until that's resolved.

Issue #2879, opened 2026-09-16
[BUG] bmad setup fails on Windows when _bmad/ already exists (WinError 183 in replace_dir)


blueshing:
### Description

On Windows, `bmad setup` fails with `FileExistsError: [WinError 183]` whenever `{project-root}/_bmad` already exists. The first setup run on a clean project succeeds (that path does a plain `src.rename(dest)`), but every subsequent run aborts.

The cause is in `skills/bmad/scripts/setup.py` → `replace_dir()` ([main, line 1178](https://github.com/bmad-code-org/BMAD-METHOD/blob/main/skills/bmad/scripts/setup.py#L1178)):

```python
backup = Path(tempfile.mkdtemp(prefix="_bmad.old-", dir=dest.parent))
try:
    dest.rename(backup)
```

`tempfile.mkdtemp()` **creates** the backup directory, and then `dest.rename(backup)` is asked to rename onto it. On POSIX, `rename()` onto an existing empty directory succeeds, so this works on macOS/Linux. Windows `MoveFileEx` without `MOVEFILE_REPLACE_EXISTING` refuses any existing target — and it never replaces an existing *directory* — so the rename always raises.

`replace_dir()` is used by both setup and doctor (`materialize_bmad` / the doctor staging path), so `bmad doctor` is affected the same way on Windows.

### Steps to reproduce

1. On Windows, in a project directory, run `bmad setup` (via the `bmad-toolbox` plugin skill) — this succeeds and creates `_bmad/`.
2. Run `bmad setup` again in the same project.
3. The script exits non-zero with a traceback.

Equivalent direct invocation:

```
uv run --no-cache "{skill-root}/scripts/setup.py" --project-root "{project-root}" --skill "{skill-root}"
```

### Expected behavior

The second run re-materializes `_bmad/`, preserving `config.toml`, `custom/`, and `*.user.toml`, and exits 0 — as it does on macOS/Linux.

### Actual behavior

The run aborts. `_bmad/` is left intact and the staging directory is cleaned up, so nothing is corrupted, but setup can never be re-run (and `bmad doctor` cannot repair) on Windows.

### Relevant log output

```shell
Traceback (most recent call last):
  File "...\skills\bmad\scripts\setup.py", line 1628, in <module>
    raise SystemExit(main())
  File "...\skills\bmad\scripts\setup.py", line 174, in main
    setup(
  File "...\skills\bmad\scripts\setup.py", line 224, in setup
[...]

PR #2904, opened 2026-09-18
fix(setup): allow repeated Windows setup

lorenzozanee:
## What
Allow `bmad setup` to replace an existing `_bmad` directory on Windows.

## Why
Windows rejects the replacement when the temporary backup directory already exists, causing repeated setup to fail with WinError 183. Fixes #2879

## How
- Remove the empty temporary backup directory before renaming the existing runtime.
- Add a regression test that simulates Windows directory-rename behavior.

## Testing
The focused regression, setup test suite, repository quality checks, and `git diff --check` pass. Native Windows execution was not available locally.


alexeyv:
Thank you for this, and for the clear write-up of the cause. Closing as superseded by #2918, which reached `dev` with #2917.

#2918 fixes the same defect (#2879) a different way: the backup is a name that does not exist yet, so no rename lands on an existing directory and there is nothing to remove first. #2917 also added a native Windows job to CI, so the failure was reproduced on Windows before the fix and passes there after it, which was the evidence this PR could not have.

Issue #2886, opened 2026-09-17
Code Map cannot say which files a story writes and which it must not touch


johnkattenhorn:
The Code Map is asked to carry two different kinds of file and has one way to write them down. `step-02-plan.md:23`:

> Put the investigation into `## Code Map`: paths, symbols or lines, **what to reuse, and what not to change**.

(Emphasis mine.)

`spec-template.md` gives one shape for both:

```
- `FILE` -- ROLE_OR_RELEVANCE
```

So "this story writes this file" and "this file exists, leave it alone" come out looking identical, and the difference survives only in the English of the role text.

## What it looks like in practice

A Code Map `bmad-build` wrote on 2026-09-17, planning a story that adds two new modules:

```
- `src/mdlinks/records.py` -- NEW. The whole deliverable. Six frozen
  dataclasses, `MdlinksError`, `sort_key`.
- `pyproject.toml:15` -- `packages = ["src/scratch"]`. **Blocker, read below.**
  Hatch exposes only `scratch`; a new `src/mdlinks` is not packaged.
- `src/scratch/__init__.py` -- existing package, unrelated. **Do not modify or
  remove.**
- `tests/test_smoke.py` -- imports `scratch`. **Do not modify.** It exists so
  `pytest` collects something; breaking it fails the gate for an unrelated reason.
```

Four bullets, one identical shape, and four different relationships: a file to create, a file to edit, a file to leave alone, and a file that must keep working. Two of them carry **Do not modify** in bold prose, which is the only signal there is.

## Why it matters inside BMAD

`step-03-implement.md:33` makes the spec the subagent's sole source of truth and tells the dispatcher not to restate ownership boundaries, because "that material already lives in it (investigation findings in its Code Map…)". The implementing subagent therefore has to read `**Do not modify.**` out of prose to know which files it may write. That works because the subagent is a language model. It is still the only place in the spec where a hard boundary is expressed in a form nothing can check.

It also means a spec cannot be linted for it. A review step can ask whether the acceptance criteria are Given/When/Then, and cannot ask whether the story touched a file its own Code Map told it not to.

## Where I hit it

I am building a scheduler that runs BMAD stories as parallel lanes, and it takes each story's file list from its Code Map to decide which stories can run at once without writing over each other. Reading all four bullets as files this story will touch makes it claim two it is forbidden to modify, so other stories that legitimately need them wait for nothing.

[...]

PR #2908, opened 2026-09-18
docs: distinguish Code Map read and write files

lorenzozanee:
## What

Add explicit `(write)` and `(read)` markers to Code Map entries in both build skill variants.

## Why

Code Map entries currently use one shape for files a story may change and files it must preserve, leaving ownership semantics in prose.

Fixes #2886

## How
- Add marker guidance and examples to both Code Map templates.
- Tell both planning steps to mark new entries while preserving unmarked legacy entries.

## Testing

Ran the focused Code Map contract check, whitespace validation, and the full repository quality gate.


alexeyv:
It's far from obvious that it's a good idea, actually. See my comment closing the issue.

PR #2909, opened 2026-09-18
fix(setup): allow repeated Windows setup

lorenzozanee:
## What
Allow `bmad setup` to replace an existing `_bmad` directory on Windows.

## Why
Windows rejects the replacement when the temporary backup directory already exists, causing repeated setup to fail with WinError 183.

Fixes #2879

## How
- Remove the empty temporary backup directory before renaming the existing runtime.
- Add regression coverage for the Windows directory-rename behavior.

## Testing
The focused regression, setup test suite, repository quality checks, and `git diff --check` pass. Native Windows execution was not available locally.


alexeyv:
fixed another way in #2917 , thanks for bringing it up

